PR #53, opened 2026-04-24
bugfix/filechange-json-replacement

### LINKED ISSUE #52: refactor: replace fragile fileChanges JSON with tool-based file operations
**Problem**
The implementation required the AI to produce file changes as a structured JSON array (fileChanges) with SEARCH/REPLACE diff blocks:

```json
"fileChanges": [
  {
    "path": "src/main/java/Foo.java",
    "operation": "UPDATE",
    "diff": "<<<<<<< SEARCH\nold code\n=======\nnew code\n>>>>>>> REPLACE"
  }
]
```
This approach is extremely fragile in practice:

* Only premium AI models (e.g. Claude Opus, GPT-4o) reliably produce syntactically valid SEARCH/REPLACE diffs. Cheaper/smaller models consistently fail.
* Diffs breaks silently the moment the actual file content differed even slightly from what the model expected — different variable names, extra blank lines, or minor formatting differences were enough to cause DiffApplyException.

**Excpected**
Replace all file-modification logic with four explicit file tools executed in a cloned workspace, identical to how validation tools (mvn, gradle, etc.) are already handled:

Tool
`write-file` 
`patch-file`
`mkdir`
`delete-file`

The agent should be work as follows:
* Clones the repository once into a temp workspace
* Uses cat to read exact file content before patching
* Runs file tools + validation tools in the same runTools array
* Commits and pushes the workspace after validation passes
* Cleans up the workspace

tmseidel:


max-california (commented):
## 🤖 AI Code Review

### Diff chunk 1/3
## Code Review: bugfix/filechange-json-replacement

### Overall Assessment
This is a substantial architectural refactor — replacing a fragile SEARCH/REPLACE diff system with explicit file tools (`write-file`, `patch-file`, etc.) executed in a cloned workspace. The motivation is sound and the approach is cleaner. Below are the issues I found.

---

### Bugs / Logic Errors

**1. `executeWithRetry` is dead code**
`executeWithRetry` is defined but never called in `handleIssueAssigned`. The main loop uses `runToolImplementationLoop` instead. In `handleIssueComment`, `executeWithRetry` *is* called — but `runToolImplementationLoop` is not. This inconsistency means the two code paths have different retry semantics. One of them should be removed or unified.

**2. `getLastPlanFromSession` is referenced but never defined**
```java
ImplementationPlan lastPlan = getLastPlanFromSession(session);
```
This method appears 4 times but is never implemented in the diff. This will cause a compilation failure.

**3. `attempt--` decrement interacts poorly with `maxRetries`**
In `runToolImplementationLoop`, context request rounds don't count as attempts (via `attempt--`). However, `maxRetries` also controls the outer loop — so a slow AI doing many context rounds could exhaust retries without ever attempting implementation. Consider a separate counter for context rounds vs. implementation attempts (the `fileRequestRounds` counter exists but the `attempt--` bypass undermines it).

**4. `buildToolsInfo()` uses `availableTools` but also calls `getAvailableFileTools()`**
```java
List<String> availableTools = toolExecutionService.getAvailableTools();
return "\n\n**Available file tools**...: " + String.join(", ", toolExecutionService.getAvailableFileTools())
```
The `availableTools` local variable is assigned but only used at the end. `getAvailableFileTools()` is a new method — if not implemented on `ToolExecutionService`, this won't compile. Worth verifying all new interface methods are implemented.

**5. `repositoryClient.getDefaultBranch(owner, repo)` called in `runToolImplementationLoop`**
```java
repositoryClient.getDefaultBranch(owner, repo),
```
This is called without using `baseBranch` which was already resolved earlier in `handleIssueAssigned`. This makes an unnecessary API call and could return a different branch than intended.

---

### Design / Architecture Issues

**6. `executeWithRetry` creates its own local `history` but also uses `sessionService`**
The `history` list in `executeWithRetry` starts empty and is built locally, but `sessionService.addMessage` is also called. These two message histories will diverge — the `aiClient.chat` call uses the local `history`, not the session's full history. The session history passed via `updHistory.subList(0, updHistory.size() - 1)` could contain stale or duplicate entries.

**7. Workspace cleanup not guaranteed in `handleIssueComment` on early return**
```java
if (plan == null || !plan.hasToolRequest()) {
    sessionService.setStatus(session, AgentSession.AgentSessionStatus.PR_CREATED);
    return;  // workspaceDir not cleaned up here
}
```
The `finally` block at the end will clean up, but this early `return` before entering the `try`... wait, `workspaceDir` is set inside the `try` block, so the `finally` will run. This is actually fine — but it's worth double-checking all early returns are inside the `try` block. ✅

**8. Removed `deleteBranch` on failure without replacement**
Previously on failure the created branch was deleted. Now the workspace is cleaned up but the remote branch (if partially pushed) is not. This is less of an issue if `commitAndPush` is atomic, but worth noting.

---

### Minor Issues

**9. Inconsistent indentation in string templates**
```java
"""
        🤖 **AI Agent**: I was unable to produce...
        You can mention me in a comment..."""
```
vs.
```java
"""
🤖 **AI Agent**: Failed to implement..."""
```
Some text blocks have leading spaces (will appear in the comment), others don't. The old code had this too, but the new code makes it worse.

**10. `normalizeWhitespace` method still referenced indirectly**
The deleted `DiffApplyService` contained this — verify no other class references it.

---

### Summary of Critical Issues

| Severity | Issue |
|----------|-------|
| 🔴 Compile error | `getLastPlanFromSession` not defined |
| 🔴 Compile error | Possibly missing `getAvailableFileTools()`, `isFileTool()`, `isValidationTool()` on `ToolExecutionService` (not shown in diff) |
| 🟠 Logic bug | `executeWithRetry` is dead code in `handleIssueAssigned` |
| 🟠 Logic bug | Unnecessary `getDefaultBranch` API call in `runToolImplementationLoop` |
| 🟡 Design | Dual history tracking in `executeWithRetry` |

### Diff chunk 2/3
## Code Review: bugfix/filechange-json-replacement

### Overall Assessment
This is a significant architectural refactor replacing a file-change model (where the AI returned `fileChanges` JSON) with a tool-based approach (`runTools` containing `write-file`, `patch-file`, etc.). The direction is sound, but there are several issues worth addressing.

---

### Bugs / Logic Errors

**1. `AgentSessionService.java` – missing newline at end of file**
The file ends without a newline (`+}` without trailing `\n`). Minor but inconsistent with Java conventions and may cause noisy diffs.

**2. `truncateToFirstJsonObject` – doesn't handle JSON arrays**
The method only tracks `{`/`}` braces. If the AI ever returns a top-level JSON array `[...]` the method will return the input unchanged, then `repairTruncatedJson` will receive unexpected input.

**3. `AiResponseParser` – ID counter `idx` not reset between `requestTools` and `runTools`**
The `idx` variable starts at 1 for both loops independently — that's fine — but if both `runTools` is empty and `runTool` is used, the fallback assigns `"tool-1"` unconditionally. This is fine, but it's inconsistent with the indexed approach used elsewhere.

**4. `executePatchFileTool` – replaces ALL occurrences**
```java
String newContent = originalContent.replace(searchText, replacementText);
```
`String.replace` replaces **every** occurrence of `searchText`. If the search text appears multiple times, this will silently apply the replacement everywhere, potentially corrupting the file. A search-count check or `replaceFirst` with a clear error when >1 match found would be safer.

**5. `WorkspaceService.commitAndPush` – no newline before closing brace**
Minor: there's a double blank line before `cleanupWorkspace` javadoc (cosmetic but inconsistent).

**6. `getLastPlanFromSession` – iterates all messages but `toAiMessages` may be expensive**
Each call deserializes the full message set. If called in a loop this could be problematic, though in the visible diff it appears called once.

**7. Missing null-safety in `buildMultiToolFeedback`**
```java
ToolResult result = i < results.size() ? results.get(i) : null;
```
This is handled (the `result == null` branch exists), but the method does not guard against `results` itself being `null`. The Javadoc says "Corresponding results" but there's no `@NonNull` or null check.

---

### Security Issues

**8. `executeWriteFileTool` / `executePatchFileTool` / `executeMkdirTool` – path traversal**
The code calls `resolveWorkspacePath` (not shown in this diff chunk) which presumably validates the path. If that method doesn't canonicalize and check the resolved path is still under `workspaceDir`, there is a path traversal vulnerability (e.g., AI passes `../../etc/passwd` as path argument). **This should be verified.**

**9. `postToolResultComment` – tool ID injected into Markdown without sanitization**
```java
comment.append("🔧 **Tool Execution**").append(id).append(": `")...
```
If the AI-generated `id` contains backticks or other Markdown, it could break the comment format. IDs from the example look like UUIDs, but the parser also generates `"tool-1"` style IDs from AI responses which could theoretically contain arbitrary strings.

---

### Performance Issues

**10. `sanitizeInvalidJsonEscapes` – compiles Pattern on every call**
```java
java.util.regex.Pattern p = java.util.regex.Pattern.compile("(\\\\\\\\)|\\\\([^\"\\\\bfnrtu/])");
```
This should be a `static final` constant.

---

### Code Style / Best Practices

**11. `IssueNotificationService` constructor change is a breaking API change**
Adding `ToolExecutionService` as a constructor parameter means all instantiation sites must be updated. The diff appears to update the test constructor, but this creates a coupling between the notification service (presentation layer) and the tool execution service (execution layer). Consider passing just the set of silent tool names, or a predicate.

**12. `buildMissingToolFeedback` contains hardcoded example UUIDs**
The UUIDs `a6f3c1d2-7e84-4b0a-9f12-e5d8c3a10001` appear in both the prompt builder and the prompt example. This is fine for examples but could confuse the AI into thinking these specific IDs are expected.

**13. `AgentSession` constructor no longer sets status**
```java
-        this.status = AgentSessionStatus.IN_PROGRESS;
```
This was removed. New sessions will have `null` status unless set elsewhere. Any code that checks `session.getStatus()` without null-guarding will fail.

**14. `appendToolRequestLine` – `reduce` instead of `Collectors.joining`**
```java
.reduce((a, b) -> a + " " + b).orElse("")
```
This is O(n²) string concatenation. Use `Collectors.joining(" ")` instead.

---

### Good Changes

- Removing `FileChangeApplier` and `DiffApplyService` integration simplifies the flow significantly.
- `truncateToFirstJsonObject` correctly handles escaped characters in strings — well implemented.
- The `resolveWorkspacePath` guard pattern (if properly implemented) is the right approach for workspace security.
- `Math.clamp` usage is cleaner than the previous `min/max` chain.
- Arg truncation in `appendToolRequestLine` prevents large file contents leaking into issue comments — good defensive practice.

### Diff chunk 3/3
_Review for this chunk failed: 429 Too Many Requests: "{"type":"error","error":{"type":"rate_limit_error","message":"This request would exceed your organization's rate limit of 30,000 input tokens per minute (org: a0c8125e-7275-4697-9a9d-b21902566e1b, model: claude-sonnet-4-6). For details, refer to: https://docs.claude.com/en/api/rate-limits. You can see the response headers for current usage. Please reduce the prompt length or the maximum tokens requested, or try again later. You may also contact sales at https://claude.com/contact-sales to discuss your options for a rate limit increase."},"request_id":"req_011CaN72H3yEkuVEufyvDvyc"}"_

**Note:** 1 of 3 diff chunk(s) could not be reviewed due to API errors.

---
*Automated review by AI Gitea Bot*

max-california (commented):
## 🤖 AI Code Review

## Code Review: bugfix/filechange-json-replacement (Updated)

### Overall Assessment
The updated diff is a significant improvement. The documentation is clearer, the `DiffApplyService` removal is now complete and consistent, and the architectural direction (tool-based file changes instead of diff blocks) is sound. Most of the critical issues from the previous review appear to have been addressed. Here are the remaining concerns:

---

### Bugs / Logic Errors

**1. `patch-file` still replaces ALL occurrences**
The prompt now explicitly documents this behavior:
> `patch-file` performs a literal string replacement — no fuzzy matching

But `String.replace` replaces *every* occurrence. If the search string appears twice in a file (common in code — think repeated method signatures, import patterns), both are silently replaced. The prompt doesn't warn the AI about this, and there's no runtime guard. At minimum, the implementation should check the match count and return an error if >1 match is found, so the AI can provide a more specific search string.

**2. `cat` before `patch-file` in `runTools` is problematic**
The prompt recommends:
```json
{"tool": "cat", ...},
{"tool": "patch-file", ...}
```
Both in the same `runTools` array. But if these execute sequentially, the AI had to write the `patch-file` search string *before* seeing the `cat` output — making the `cat` result useless for that patch. The AI would need to issue `cat` in one round, then `patch-file` in the next. Either the prompt should clarify this, or `cat` in `runTools` should be executed before `patch-file` in the same batch (which requires ordering guarantees and result-passing within a single round that may not exist).

**3. `delete-file` silently succeeds on missing files**
The prompt says:
> *Delete a file (silently succeeds if it does not exist)*

This could mask AI mistakes (wrong path, typo) without feedback. A warning result would be more useful.

---

### Security

**4. Path traversal protection — verify implementation**
The doc now states:
> *File tools reject any path that escapes the workspace root*

The diff doesn't show the `resolveWorkspacePath` implementation. From the previous review, this still needs verification: the resolved path must be canonicalized (to resolve `..` and symlinks) before checking it's under `workspaceDir`. A `Files.toRealPath()` or equivalent is required — `Path.normalize()` alone is insufficient against symlink attacks.

---

### Prompt / Documentation Issues

**5. Hardcoded example UUIDs used throughout the prompt**
The prompt uses `550e8400-e29b-41d4-a716-44665544000X` as examples in *multiple places*. Some models may latch onto these as a valid prefix pattern and reuse them, causing ID collisions. Using clearly placeholder values like `<uuid-1>`, `<uuid-2>` would be cleaner.

**6. `runTool` (singular) backward compatibility note in the prompt is misleading**
```json
"runTool": {"tool": "mvn", "args": ["compile", "-q"]}
```
The prompt marks this as deprecated but still supported. If the implementation is being cleaned up, keeping deprecated formats in the prompt trains the AI to sometimes use the old format. Consider removing it from the prompt entirely while keeping the parsing code for backward compatibility.

**7. Validation tool mandate is weakened**
The old prompt had:
> **NEVER omit `runTool` when providing `fileChanges`** — validation is mandatory

The new rules say:
> Always include at least one validation tool (`mvn`, `gradle`, `npm`, etc.) in `runTools`

The urgency is lower. Given that validation is a core correctness guarantee, this deserves equally strong emphasis (caps, bold) as the old prompt had.

---

### Code Style

**8. Alignment-style spacing is inconsistent**
```java
String owner     = payload.getRepository().getOwner().getLogin();
String repo      = payload.getRepository().getName();
Long   issueNumber
```
This alignment style is non-standard for Java and will conflict with most auto-formatters (Checkstyle, google-java-format). It's a minor issue but creates noisy diffs on reformatting.

**9. `IntStream` import added — verify it's actually used**
`import java.util.stream.IntStream;` is added but not visible in the truncated diff. If unused, it should be removed.

---

### Good Changes

- Removing `DiffApplyService` (558 lines of complex fuzzy matching logic) is a net win — the complexity wasn't reliable and the new approach is simpler and more predictable.
- Documentation is substantially improved and accurately reflects the new tool-based architecture.
- The `AGENT_MAX_FILES` config removal is appropriate since file ops are now tool-based.
- Security section update (path traversal protection vs. file limit) is more accurate.
- The `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL` constants are a good addition for traceability.

---

### Summary

| Severity | Issue |
|---|---|
| 🔴 Logic | `cat` in `runTools` before `patch-file` in same batch is architecturally inconsistent |
| 🟠 Bug | `patch-file` silently replaces all occurrences — needs count guard |
| 🟠 Security | Path traversal protection unverified in this diff |
| 🟡 Prompt | Hardcoded example UUIDs may cause collisions |
| 🟡 Prompt | Deprecated `runTool` in prompt will confuse AI |
| 🟡 Style | Alignment spacing will conflict with auto-formatters |

---
*Automated review by AI Gitea Bot*

max-california (commented):
## 🤖 AI Code Review

## Code Review: bugfix/filechange-json-replacement (Final Update)

### Summary of Changes from Previous Round
All three previously flagged issues in `prompts/agent.md` have been addressed:
- ✅ Hardcoded example UUIDs replaced with `<uuid-N>` placeholders
- ✅ `cat`+`patch-file` same-batch problem explicitly documented with a two-round example
- ✅ Validation mandate restored with CAPS/bold emphasis
- ✅ `delete-file` now returns a warning on missing file
- ✅ `patch-file` "must match exactly once" constraint documented

The Java changes (`BotWebhookService`, `DiffApplyService` deletion, `IssueImplementationService`) are consistent and clean.

---

### Remaining Issues

**1. `patch-file` "exactly once" constraint is only in the prompt, not enforced in code**
The prompt correctly states the search text *must match exactly once*, and the tool returns an error if it matches more than once. But this is not visible in the diff — `executePatchFileTool` (not shown) uses `String.replace` which silently replaces all occurrences. The enforcement needs to be in the implementation:
```java
long count = countOccurrences(content, searchText);
if (count == 0) return ToolResult.error("Search text not found");
if (count > 1)  return ToolResult.error("Search text matches " + count + " locations — provide more context");
```
Without this, the runtime behavior contradicts the documented contract.

**2. `IntStream` import added but not visible as used**
`import java.util.stream.IntStream;` appears in the diff. The truncated diff doesn't show it being used. If it's unused this should be removed to avoid compiler warnings.

**3. Path traversal verification still absent from diff**
Security consideration #3 in the doc claims path traversal protection exists, but `resolveWorkspacePath` implementation is not in this diff. This remains unverified — please confirm it uses `toRealPath()` or `toAbsolutePath().normalize()` followed by a `startsWith(workspaceDir)` check.

**4. Minor: alignment-style spacing still present**
```java
String owner     = ...
String repo      = ...
Long   issueNumber
```
Non-standard for Java, will conflict with any auto-formatter. Low priority but worth cleaning up.

---

### Prompt Quality: Good

The two-round `cat`→`patch-file` workflow example is clear and correct. The `<uuid>` placeholder approach is much better than the previous hardcoded values. The Rules section now has appropriately strong language for validation.

One small nit: the prompt's `// Round 1` and `// Round 2` comments are not valid JSON — if any tooling validates the prompt examples as JSON snippets, these will fail. Plain prose separation would be safer.

---

### Overall Assessment

This is a solid, well-reasoned refactor. The core issues from previous rounds are resolved. The only **blocking** item is **#1** — the `patch-file` uniqueness check must be enforced in code, not just documented. Everything else is minor.

---
*Automated review by AI Gitea Bot*

PR #54, opened 2026-04-24
AI Agent: Other authors can talk also to the bot in a PR (fixes #20)

max-california:
Fixes #20

## Summary

Need to see BotWebhookService to understand the routing and how PR author check is done

---
*This PR was automatically generated by the AI implementation agent. Please review carefully before merging.*


max-california (commented):
## 🤖 AI Code Review

A few important issues in this change:

- **Main logic bug:** `userMessage` is computed for side clarification, but the AI call still uses `commentBody`:
  ```java
  String response = aiClient.chat(history, commentBody, systemPrompt, null);
  ```
  So the model never receives the “Another PR participant…” / “side clarification” instruction. The tests expect that behavior, but the implementation doesn’t do it. This likely makes the feature ineffective and may cause the new tests to fail. Pass `userMessage` to `aiClient.chat(...)`.

- **Behavioral mismatch / likely incorrect detection:** `isSideClarificationRequest(commentAuthor)` treats **any non-bot author** as a side clarification. Unless routing guarantees only the PR author previously reached this method, this will also mark the PR author’s own bot requests as “side clarification,” which doesn’t match the issue wording. The check should probably compare `commentAuthor` to the **PR author**, not to `botUsername`.

- **Potential null-safety issue:** `Long commentId = payload.getComment().getId();` and `String commentBody = payload.getComment().getBody();` are accessed before the null-safe author extraction. If this method can be called for malformed/unexpected webhook payloads, it will NPE before hitting the safe branch. Consider validating `payload.getComment()` once up front.

- **Security / prompt-injection surface:** `commentAuthor` is interpolated directly into the prompt text. Usernames are usually constrained by the provider, but if not, this is user-controlled prompt content. At minimum, sanitize/quote defensively or avoid embedding raw usernames in instruction-like text.

- **Test coverage gap:** there should be an assertion that `sessionService.addMessage(..., "user", userMessage)` stores the side-clarification prompt, and a direct verify that `aiClient.chat(...)` receives the rewritten prompt. Right now the tests are aimed at this behavior, but given the implementation bug, they should explicitly catch it.

- **Style:** the null check for `commentAuthor` is redundant after dereferencing `payload.getComment()` above. Better to extract `var comment = payload.getComment();` once and use it consistently.

Overall, the formatting addition looks fine, but the core feature currently appears only partially wired.

---

tmseidel:
..wrong branch

PR #60, opened 2026-04-24
AI Agent: Other authors can talk also to the bot in a PR (fixes #59)

max-california:
Fixes #59

## Summary

Need more context on webhook handling, session reuse, and comment reply generation for PR discussions on develop

---
*This PR was automatically generated by the AI implementation agent. Please review carefully before merging.*


max-california (commented):
## 🤖 AI Code Review

A few important issues to address before merging:

- **Migration breakage / duplicate DDL risk**
  - Moving `review_sessions`, `agent_sessions`, `conversation_messages`, and `agent_file_changes` from **H2 V1** into **H2 V3** is dangerous for existing H2 databases. Flyway will not re-run V1, so environments already on V1/V2 but not V3 may now depend on V3 to recreate tables that were historically created by V1. That creates schema drift across instances/tests.
  - Also, in both H2 and PostgreSQL migrations, `alter table ... add constraint fk_review_session_participants_session ...` is not guarded with `IF NOT EXISTS`. On a partially migrated DB this can fail hard if the constraint already exists.

- **Session uniqueness bug**
  - `ReviewSession` uniqueness is still only `(repoOwner, repoName, prNumber)`, but sessions are looked up with `promptName` passed into `getOrCreateSession(...)`. If the same PR is reviewed with different prompts, they’ll share one session and participant set, causing context/prompt contamination. Either include `promptName` in the unique key + lookup, or stop accepting it in session creation.

- **Possible compile/runtime issue in `compactContextWindow`**
  - `messages` is declared as `List<ConversationMessage>`, but the code uses `removeFirst()` / `addFirst()`. Those are not `List` methods. Unless `messages` was changed to a `LinkedList` elsewhere, this won’t compile.

- **Incorrect “known participant” semantics**
  - A participant is only remembered **after** they first talk to the bot. That means the original PR author/reviewer is treated as a “new participant” on their first follow-up, so responses may be marked as side-clarifying even when they shouldn’t be. If “known participant” means someone already involved in the review, seed participants from PR author / review author, or clarify the behavior.

- **Concurrency / duplicate session creation**
  - `getOrCreateSession()` does a read-then-insert without handling races. Concurrent webhook deliveries for the same PR can create duplicate inserts and hit the unique constraint. Consider catching `DataIntegrityViolationException` and reloading, or using a DB-backed upsert/locking strategy.

- **Repository loading pattern**
  - `findByRepoOwnerAndRepoNameAndPrNumber()` fetch-joins `messages`, while `hasParticipant()`/`rememberParticipant()` do a second query for participants. This is okay functionally, but it adds extra round-trips per webhook. If this path is hot, consider fetching both collections for this use case, or redesign to avoid repeated reloads.

- **DB schema best practice**
  - `review_session_participants` should ideally have a uniqueness constraint on `(session_id, participant_login)` to enforce set semantics at the DB level, not just in Java.

Otherwise the feature direction looks good, and the tests added around new-participant behavior are useful.

---

PR #62, opened 2026-04-25
feature/backmerge-to-develop

tmseidel:


max-california (commented):
## 🤖 AI Code Review

Mostly good; the branch-context fix in `IssueImplementationService` and the added tests look valuable. A few things to watch:

- **Potential PR base bug in follow-up comments**
  - In `handleIssueComment`, PR creation now uses `createNew ? selectedContextBranch : defaultBranch`.
  - If the AI uses `branch-switcher` only for **context discovery** during a follow-up comment, `selectedContextBranch` may become a different branch than the actual branch the workspace changes were committed against (`branchName` / working branch).
  - That can create a PR from `ai-agent/issue-42` into the wrong base branch. At minimum, ensure branch switching is intended to mutate the eventual PR base in comment flows, or track PR base separately from “context branch”.

- **State consistency**
  - `runToolImplementationLoop` now returns the selected branch, but I don’t see session persistence of this updated base branch. If later comments/retries depend on the latest chosen base, behavior may drift across requests unless that branch is stored explicitly.

- **Security / trust boundary**
  - `branch-switcher` lets the model influence which branch is read from and, indirectly, which branch may become PR base. That’s powerful. Consider restricting allowed target branches (e.g. existing repo branches only, maybe disallow protected branches as PR targets unless explicitly configured) and logging/auditing branch switches clearly.

- **Performance**
  - In the loop, branch changes can trigger repeated `getRepositoryTree` / `getFileContent` calls across rounds. Probably fine at current scale, but caching per `(branch, path)` within a single run would reduce redundant API traffic.

- **Tests**
  - Nice coverage addition. I’d add one more test for the risky case above: follow-up comment where `branch-switcher` changes context branch, but changes are still committed to existing working branch, to assert the intended PR base behavior.

- **Provider metadata**
  - Refreshing suggested models is fine, but these names are effectively product-facing defaults. If they aren’t guaranteed available, consider validating model availability separately or keeping conservative/stable suggestions to avoid setup failures/confusion.

Overall: good fix and test additions, but I’d revisit the PR base semantics introduced in `handleIssueComment`.

---

PR #85, opened 2026-05-03
bugfix/82-distinguish-comments

tmseidel:


max-california (commented):
## 🤖 AI Code Review

### Diff chunk 1/3
1. **Blocking issues**

- **`WriterAgentService` imports `tools.jackson.databind.ObjectMapper` instead of `com.fasterxml.jackson.databind.ObjectMapper`**
  
  This looks like a hard compile failure unless the project intentionally provides a different Jackson package, which would be unusual. Given the rest of the codebase is standard Spring Boot, this should almost certainly be `com.fasterxml.jackson.databind.ObjectMapper`.

- **`BotController.save(...)` does not reliably repopulate the submitted `Bot` on validation/error paths**
  
  On failure, the method returns `"bots/form"` after adding form attributes, but it never re-adds `bot` to the model explicitly. Depending on Spring MVC behavior with `@ModelAttribute`, this may work implicitly, but it’s fragile and easy to break when the method evolves. More importantly, if binding of associations failed, the form may re-render without the expected selected values. Safer fix: explicitly `model.addAttribute("bot", bot);` before returning the form.

- **Possible nullability / migration regression around `Bot.systemPrompt`**
  
  `Bot.systemPrompt` is now `@ManyToOne(optional = false)` and several runtime paths throw if it is null. That’s fine only if the DB migration guarantees every existing bot gets a valid `system_prompt_id` before this code runs. I don’t see the migration in this chunk, so this needs verification before merge. Without it, existing rows will break form rendering and webhook handling. At minimum, confirm both H2 and PostgreSQL migrations backfill all existing bots and enforce the FK safely.

- **`AgentSessionService.claimSessionForUpdate(...)` excludes `ISSUE_CREATED` and `FAILED`, but not `COMPLETED` / `PR_CREATED` semantics should be verified**
  
  This may be intentional, but it changes locking/continuation behavior for shared session handling. For writer sessions, allowing continuation from `COMPLETED` would be odd; for coding sessions, allowing continuation from `PR_CREATED` seems intended. Since this method is now central to concurrency control, verify callers don’t accidentally allow updates from terminal states other than the ones explicitly blocked. If `COMPLETED` is terminal, it should probably also be rejected here.

- **Centralize bot invariants closer to the entity/service boundary**
  
  `BotService.save()` silently forces `agentEnabled=false` for writer bots, which is good, but other invariants now exist too: writer bots ignore PR events; bots require a system prompt; prompt text is mode-specific. Consider a small validator/helper to enforce all bot invariants in one place so controller, service, and webhook code don’t each need partial checks.

- **`AgentErrorNotificationService.sanitizeErrorMessage(...)` may still leak internal details**
  
  It strips formatting characters but still posts raw exception messages publicly. Some exception messages can expose internal paths, API details, SQL fragments, or provider responses. Consider posting a generic user-facing message and logging the detailed exception server-side only.

- **Search/find emulation behavior in `ToolExecutionService` should be documented or narrowed**
  
  The new parsing for `rg`/`grep`/`find` is useful, but it only supports a subset of flags and now interprets unknown `-...` args generically. That may produce surprising results for model-generated commands. If intentional, document the supported subset clearly and consider returning an error on unsupported flags instead of silently ignoring them.

- **`executeFindTool(...)` now filters to regular files only**
  
  This is a behavior change. If the writer/coding workflows rely on discovering directories (for example, to understand package/module layout), this is a regression. If the intent is file-only matching, consider renaming/documenting that behavior; otherwise preserve previous directory matches.

- **`BotWebhookService.createCodeReviewService(...)` still receives `promptService` indirectly via constructor removal changes**
  
  The code now constructs `CodeReviewService` with raw prompt content and key instead of `PromptService`. That’s a reasonable direction, but the transition is partial: issue services still use `PromptService` fallback names. Consider standardizing prompt resolution before service construction to avoid dual behavior and make future caching/versioning easier.

3. **Tests**

- Add tests for **bot form save failures**:
  - missing/invalid `systemPromptId`
  - writer bot save forcing `agentEnabled=false`
  - re-rendering form with selected bot/system prompt/integrations preserved

- Add migration/integration tests covering **existing bots upgraded to non-null `system_prompt_id`** for both H2 and PostgreSQL.

- Add tests for **writer bot event filtering** in `BotWebhookService`:
  - ignores PR review/comment/review-submitted/closed events
  - handles issue assignment/comment events only

- Add tests for **session-type conflict handling**:
  - coding bot assigned when writer session already exists
  - coding issue comment on writer-owned issue
  - writer continuation restricted to expected session types/states

- Add tests for **`ToolExecutionService` search/find parsing**:
  - `rg pattern . --glob *.java`
  - `grep -i foo src`
  - `rg foo --include=*.md`
  - escaped alternation retry (`a\|b`)
  - `find -name "*.java" src`
  - behavior for directory matches vs file-only matches

- Add tests for **`AgentErrorNotificationService`** to verify sanitization and avoid accidental markdown/code-block breakage.

4. **Overall assessment**

Good direction overall: the bot-type split and prompt-system refactor make sense, and most changes are cohesive. But there’s at least one likely compile issue (`tools.jackson...`), plus migration/nullability safety and a few behavior changes that should be verified with tests before merge.

### Diff chunk 2/3
1. Blocking issues

- **H2 migration is inconsistent with the new entity schema and likely breaks startup for fresh DBs.**  
  `SystemPrompt` now requires `writer_agent_system_prompt` (`@Column(nullable = false, columnDefinition = "TEXT")`), but `src/main/resources/db/migration/h2/V3__system_prompts.sql` creates `system_prompts` **without** that column and inserts rows without it. The column is only added later in `V5`. On a fresh H2 database, Hibernate/Flyway validation can fail after V3/V4 because the table shape doesn’t match the entity yet.  
  **Fix:** either include `writer_agent_system_prompt` in `V3` from the start (matching PostgreSQL once V5 is squashed into the base migration), or ensure Hibernate validation does not run against intermediate migration states. At minimum, verify fresh H2 startup end-to-end.

- **`CodeReviewService` changed session key semantics in a way that can break existing conversation continuity.**  
  Previously sessions were keyed with `promptName`; now `getOrCreateSession(..., sessionPromptKey)` is used while the method signature still accepts `promptName`, which is ignored for session lookup and prompt resolution. If `sessionPromptKey` is not guaranteed to remain stable per bot/system-prompt combination, existing review conversations may fork or collide unexpectedly after edits/migrations.  
  **Fix:** confirm the intended session identity and make it explicit. If sessions should remain bound to the selected prompt set/version, use a deterministic key derived from that selection and remove the now-misleading `promptName` parameter. Otherwise keep using the previous key.

- **Writer loop can create an issue with an empty or null draft.**  
  In `runWriterLoop`, once `plan.isReadyToCreate()` is true and there are no questions/context requests, `createIssue(...)` is called even if `revisedIssueDraft` is missing/blank. That depends entirely on model compliance and can create junk issues.  
  **Fix:** require `plan.getRevisedIssueDraft()` to be non-blank before `createIssue`; otherwise treat it as an invalid AI response and ask for another round or fail gracefully.

2. Non-blocking suggestions

- **`WriterResponseParser` imports look wrong/suspicious.**  
  It uses `tools.jackson.core.JacksonException` and `tools.jackson.databind.ObjectMapper`, while elsewhere the project uses `com.fasterxml.jackson...`. If intentional, it’s unusual enough to verify; if not, this is a compile/runtime issue waiting to happen. Prefer consistent Jackson imports.

- **`handleIssueComment` re-queries the session in `isIssueAuthor(...)`.**  
  The method already has the session, but `isIssueAuthor` fetches it again. This adds unnecessary DB work and slightly weakens reasoning around the claimed session. Passing the already loaded/claimed session into the author check would simplify the flow.

- **Bitbucket issue APIs should be verified before merge.**  
  `BitbucketApiClient` now implements real `postIssueComment`, but Bitbucket Cloud support is described elsewhere as PR-review-focused. I’m not sure issue comments / issue creation / issue search are actually supported in the same way for the intended integration. If writer bots are not meant to run on Bitbucket, consider explicitly rejecting unsupported operations rather than exposing partially implemented defaults.

- **`SystemSettingsController.save()` prevents changing `defaultEntry` through the form for existing records.**  
  For edits, it forcefully copies `existing.isDefaultEntry()` back onto the submitted object. If there is a UI/API path intended to change which prompt is default, this blocks it. If intentional, the UI should make that explicit; otherwise add a dedicated safe path for switching the default.

3. Tests

- Add a **fresh-migration test** for both H2 and PostgreSQL covering the new `system_prompts` schema, especially presence/non-nullability of `writer_agent_system_prompt`.
- Add tests for `WriterAgentService.runWriterLoop` covering:
  - `readyToCreate=true` with blank `revisedIssueDraft` does **not** create an issue
  - non-author comments are ignored
  - branch-switch requests update session branch state correctly
  - failure to prepare workspace posts the correct issue comment/status
- Add parser tests for `WriterResponseParser`:
  - fenced JSON / unfenced JSON / malformed JSON fallback
  - multiple tool requests
  - missing `id` generation
- Add/adjust tests for the repository client split:
  - `postPullRequestComment` vs `postIssueComment` for GitLab specifically
  - regression coverage in `CodeReviewService` to ensure PR replies still use the PR-comment API

4. Overall assessment

Promising change set with clear separation of issue vs PR comments and a solid foundation for writer bots, but I would not merge until the migration/entity mismatch and the writer issue-creation guard are addressed, and the new session-key behavior is verified.

### Diff chunk 3/3
1. **Blocking issues**

- **`SystemPromptServiceTest` appears incomplete for current validation rules**
  - The added UI and tests indicate `writerAgentSystemPrompt` is now required, but `save_requiresBothPromptTexts()` only asserts failure on missing `issueAgentSystemPrompt`. There’s no corresponding test for missing `writerAgentSystemPrompt`.
  - This is only blocking if the production service actually forgot to validate writer prompts; from this chunk alone I can’t confirm. Please verify `SystemPromptService.save(...)` rejects missing/blank `writerAgentSystemPrompt` as well, since the form marks it `required` and bots now depend on it.

- **Potential mismatch between session status expectation and test intent**
  - `writerBot_clarifyingQuestionsResetSessionToWaiting()` verifies:
    ```java
    verify(agentSessionService).setStatus(session, AgentSession.AgentSessionStatus.IN_PROGRESS);
    ```
    but the test name says “ResetSessionToWaiting”.
  - Same pattern in `writerBot_contextRoundLimitResetsSessionAndPostsNotice()` / `writerBot_followUpFailurePostsVisibleErrorCommentAndResetsSession()`: names imply a reset, assertions check `IN_PROGRESS`.
  - If the implementation really leaves sessions `IN_PROGRESS`, that risks stuck writer sessions that should instead be resumable / waiting for user input. If this is just a test naming issue, rename the tests; otherwise fix the implementation/status assertions to reflect the intended lifecycle.

- **Add stronger coverage for the new navigation/UI path**
  - `layout.html` now exposes `/system-settings`, but there’s no MVC/template-level test proving the page renders and the nav is visible to authenticated users. Not critical, but easy to regress.

- **Test wording vs behavior should be aligned**
  - Several new webhook tests are good and valuable, but the naming inconsistencies around “resets session” make maintenance harder. Tightening those names would improve readability and avoid future false assumptions about session lifecycle.

- **Docs-only changes should be checked for accuracy**
  - README still contains a truncated Docker snippet in this chunk showing:
    ```
    POSTGRES_PASSWORD: $
    ```
    I assume that’s just truncation in the review payload, but worth verifying the actual file wasn’t accidentally committed malformed.

3. **Tests**

Recommended missing coverage:

- **`SystemPromptService.save(...)`**
  - missing/blank `writerAgentSystemPrompt`
  - duplicate-name handling on update vs create for unchanged names
  - trimming/blank-string validation for all three prompt fields, not just null/missing

- **`SystemSettingsController`**
  - save/edit/delete/clone flows
  - error-path assertions when delete is blocked because prompt is default or referenced by bots

- **Integration coverage**
  - a web/integration test proving a bot without an explicitly assigned system prompt still behaves correctly via default prompt resolution, if that fallback is supported
  - if not supported, a validation test proving bot save fails without `systemPrompt`

4. **Overall assessment**

This chunk looks solid overall: the test coverage added around writer-bot behavior and explicit issue/PR comment APIs is substantial and useful. The only thing I’d want resolved before merge is the apparent inconsistency between writer-session test names and asserted status transitions, plus verification that `writerAgentSystemPrompt` is enforced in the service layer.

---

PR #107, opened 2026-05-13
feature/add-comments-to-context

### LINKED ISSUE #106: Coding agent does not add comments to the context
**Current behavior**
The coding agent takes only the head and the description of an issue into account for creating the context. If useful information for the implementation are added after the issue-creation they won't be recognized by the agent

**Expected behavior**
Add all upcoming comments (excluding the coding agents thinking comments) to the context.

tmseidel:


max-california (commented):
## 🤖 AI Code Review

1. **Blocking issues**

- **`isPullRequestAuthor()` now returns true when the commenter is missing**
  
  In `BotWebhookService`, this changed from:
  ```java
  return author != null && commenter != null && author.equalsIgnoreCase(commenter);
  ```
  to:
  ```java
  return author != null && author.equalsIgnoreCase(commenter);
  ```
  Since `String#equalsIgnoreCase(null)` returns `false`, this is not an NPE risk, but it does silently change the guard semantics and makes the method depend on JDK behavior rather than an explicit null check. More importantly, this looks accidental rather than intentional and weakens readability around an auth-sensitive gate. Please restore the explicit `commenter != null` check unless there is a documented reason not to.

- **MCP client version was not bumped consistently with the project version**
  
  `pom.xml` moves the app to `1.6.0-SNAPSHOT`, but `McpOrchestrationService` still reports:
  ```java
  new McpSchema.Implementation("ai-git-bot", "1.5.0")
  ```
  in two places. This is easy to miss, but it creates inconsistent runtime metadata and can complicate debugging / interoperability if servers key behavior off client version. Please either update this to `1.6.0`/`1.6.0-SNAPSHOT` or centralize the version from build metadata.

- **Issue comment fetching is hard-capped to first page only**
  
  All provider implementations fetch only 50 comments. For older issues with long discussions, the agent may miss the most relevant clarifications depending on API default ordering. If this is intentional for prompt-budget reasons, consider explicitly requesting the latest comments (or documenting the ordering assumption), since “first 50” can easily mean oldest 50.

- **Failure text is injected directly into prompt context**
  
  `fetchIssueCommentsContext()` returns:
  ```java
  "Issue comments could not be loaded: " + e.getMessage()
  ```
  on errors. That leaks raw upstream/API error messages into the LLM prompt, which is usually low value and can expose transport/provider details unnecessarily. Prefer a generic string like `"Issue comments could not be loaded."` and keep the detailed error in logs only.

- **`AgentPromptBuilder` may include HTML content in prompts**
  
  `extractCommentBody()` falls back to `content.html` for Bitbucket-style comments. Sending HTML markup to the model is likely noisier than useful and may waste prompt budget. If possible, prefer raw/plaintext fields only; only use HTML if there is truly no raw representation, and then consider stripping tags first.

3. **Tests**

- Good additions around issue-comment context inclusion and bot-authored comment filtering.
- I’d still recommend adding:
  - a test for `fetchIssueCommentsContext()` / prompt building when comment retrieval fails, to ensure raw exception text is not included if you adopt the suggestion above;
  - provider-specific tests for `getIssueComments()` translation/shape, especially Bitbucket and GitLab where field names differ more;
  - a regression test around `isPullRequestAuthor()` null-commenter behavior, since this method gates who can trigger bot actions.

4. **Overall assessment**

The core feature—adding issue comment context to implementation prompts—looks solid and the new tests cover the main behavior well. Before merge, I’d fix the inconsistent MCP version metadata and clarify/restore the `isPullRequestAuthor()` null check.

---

max-california (commented):
## 🤖 AI Code Review

1. **Blocking issues**

- **`isPullRequestAuthor` logic changed in a way that can authorize null/unknown commenters**
  
  In `BotWebhookService`, the check changed from:
  ```java
  return author != null && commenter != null && author.equalsIgnoreCase(commenter);
  ```
  to:
  ```java
  return author != null && author.equalsIgnoreCase(commenter);
  ```
  This relies on `String.equalsIgnoreCase(null)` returning `false`, so it won’t throw, but it does weaken the guard semantics and makes the method treat “missing commenter” as just another mismatch instead of an invalid payload state. More importantly, this looks like an incidental change unrelated to the feature and could mask payload translation bugs from providers where `sender`/comment author is absent or mapped differently.

  If this was intentional to support cases where `commenter` can be null, please add a comment/test showing the provider payload that requires it. Otherwise I’d keep the explicit `commenter != null` check for clarity and safety.

- **MCP client version was not bumped consistently**
  
  `pom.xml` is now `1.6.0-SNAPSHOT`, but `McpOrchestrationService` still reports:
  ```java
  new McpSchema.Implementation("ai-git-bot", "1.5.0")
  ```
  in both normal initialization and diagnostics.

  This can make debugging interoperability problems harder and produces incorrect client metadata to remote MCP servers. Update this to the current app version, or better, source it from build metadata so it stays in sync automatically.

- **Issue comment fetching is capped to the first page only**
  
  All `getIssueComments(...)` implementations request only 50 comments and do not paginate. Since this feature is specifically about using issue discussion as agent context, older but still relevant comments may be silently omitted on long-lived issues.

  If the intent is to limit context size anyway, consider fetching the *most recent* comments explicitly where the provider supports sorting, or paginating until you hit the configured context limits. Right now the subset is provider-default and may not be the most useful one.

- **Potential prompt-quality issue: bot-authored filtering only matches exact username**
  
  `filterBotAuthoredComments` excludes comments only when `botUsername.equalsIgnoreCase(extractCommentAuthor(comment))`. Depending on provider payloads, authored comments may come back as display names rather than login/usernames, and those comments would still be included. That won’t break correctness, but it may re-feed the agent’s own progress comments into the prompt.

  Worth considering a more robust filter where possible (e.g. provider-specific IDs, or excluding comments with the standard agent prefix as a fallback).

- **`fetchIssueCommentsContext` returns provider error text directly into prompts**
  
  On failure it returns:
  ```java
  "Issue comments could not be loaded: " + e.getMessage();
  ```
  That means transport/API details can be injected into the LLM prompt. This is not a security issue by itself, but it’s noisy and can reduce prompt quality.

  Prefer a neutral message like `"Issue comments could not be loaded."` and keep the detailed error only in logs.

3. **Tests**

- Good additions around:
  - issue comments being included in assignment/continuation prompts
  - filtering bot-authored comments
  - configurable prompt limits in `AgentPromptBuilder`

- Still recommended:
  - **BotWebhookService tests** covering the `isPullRequestAuthor` behavior after the change, especially null/missing commenter cases.
  - **Repository client tests** for each new `getIssueComments(...)` implementation, to verify response-shape parsing across providers.
  - A test for **failure to fetch issue comments** ensuring the agent continues gracefully and the fallback prompt text is acceptable.
  - If pagination is intentionally not implemented, a test or comment documenting that only the first page is used.

4. **Overall assessment**

The main feature addition looks solid and the prompt-builder/test coverage is generally good. Before merge, I’d fix the MCP version mismatch and verify the `isPullRequestAuthor` change is intentional and covered by tests.

---

PR #117, opened 2026-05-17
refactor(tooling): introduce ToolCatalog as single source of truth an…

tmseidel:
…d decouple execution

Add ToolCatalog with STATIC_TOOLS as the single declaration point for all tools (Entry(name, kind, roles, description, schema)) Enable config-driven validation tools via agent.validation.available-tools; unknown tools get generic native descriptors (e.g., kubectl) Remove duplication:

Delete AgentNativeTools.java and AgentNativeToolsTest.java Move schemas & descriptions into catalog entries
Migrate schema assertions to ToolCatalogTest

Decouple responsibilities:

ToolExecutionService becomes a pure executor (remove is*Tool, getAvailable*Tools, catalog() forwarders) IssueNotificationService depends on ToolCatalog (not TES) CodingAgentStrategy classifies via ToolCatalog, executes via AgentToolRouter AgentToolRouter receives ToolCatalog via DI (no longer passed through TES)

Extensibility:

Add built-in tool by appending one entry in STATIC_TOOLS Add validation tool via config only
Adjust roles via EnumSet<Role> per tool

Net change: ~−250 LOC; tests pass without brittle classification mocks

BREAKING CHANGE: removal of AgentNativeTools and TES catalog forwarders; consumers must use ToolCatalog directly.


max-california (commented):
## 🤖 AI Code Review

1. **Blocking issues**

- **Potential NPE/regression in `ToolCatalog.validationToolNames()` / constructor paths**
  
  `ToolCatalog.validationToolNames()` unconditionally does `agentConfig.getValidation().getAvailableTools()`, and multiple call sites now depend on `ToolCatalog` for core execution/classification. This looks safe only if `AgentConfigProperties` always initializes `validation` and `availableTools` non-null. From the diff alone that guarantee isn’t obvious, and some tests instantiate `new AgentConfigProperties()` directly. If either field can be null in production binding edge cases, this refactor turns benign classification calls into runtime failures across issue handling, writer flow, and descriptor generation.

  **Suggested fix:** defensively normalize in `ToolCatalog`, e.g. return `List.of()` when config/validation/tools is null, or populate defaults in `AgentConfigProperties` and add a test that verifies null-safe behavior.

- **Behavior change for unknown validation tools may break native tool execution compatibility**
  
  `ToolCatalog.nativeDescriptors()` now exposes config-only validation tools with a generic varargs schema, which is good, but `CodingAgentStrategy.toRequest()` only has special flattening for known typed tools and otherwise falls back to raw JSON if no recognized fields were found. That means if a provider/tool-call emits an object not using `args` exactly as expected for an unknown validation tool, it will be passed as a single JSON blob string into `ToolExecutionService.executeTool(...)`, which expects positional CLI args. This is probably fine when the model honors the schema, but it’s a fragile compatibility point introduced by the new “config-only tools” extensibility.

  **Suggested fix:** in `toRequest()`, explicitly treat any tool classified as validation/context/repository with an `args` array schema as varargs-first, rather than relying on name-specific knowledge plus fallback. At minimum add a test for a config-added validation tool invoked natively.

- **`ToolExecutionService` now depends on `ToolCatalog`, which depends on `AgentConfigProperties`**
  
  This is fine today, but it creates a tighter cycle in responsibilities than the PR description implies for a “pure executor.” The executor still consults catalog classification/name lists for allowlists. That may be acceptable, but if you want stricter separation, consider letting callers validate/classify and keep `ToolExecutionService` focused purely on executing known commands.

- **`ToolCatalog.validationToolNames()` does repeated list lookups in hot classification paths**
  
  `kindOf()` calls `validationToolNames().contains(n)` each time, rebuilding/reading the config list repeatedly. This is unlikely to matter at current scale, but since classification is used broadly, caching normalized validation names in a `Set<String>` in the constructor would simplify semantics and avoid repeated linear scans.

- **Silent aliases are included in `contextToolNames()`**
  
  `namesOf(ToolKind.CONTEXT)` iterates `STATIC_TOOLS` and includes `ripgrep`/`grep`, even though comments say these aliases are “never advertised to the LLM.” That’s okay for runtime-facing “available tools” messages, but it slightly conflicts with the documentation intent. If you want a clean distinction, consider separating “classified runtime names” from “advertised descriptor names.”

3. **Tests**

- Add a test covering **null/empty validation config safety** for `ToolCatalog` and any code path that calls `validationToolNames()`.
- Add a test for **native execution of a config-added validation tool** (e.g. `kubectl`) to verify `ToolCall -> ToolRequest -> executeTool(...)` preserves positional args correctly.
- Add a test for **unknown tool bucketing/notification behavior** in `IssueNotificationService` to lock in the new “unknown => Will run / non-silent” semantics.
- If external consumers rely on removed TES forwarders, consider adding/adjusting integration tests around Spring wiring and startup to catch DI regressions from the new constructor signatures.

4. **Overall assessment**

Good refactor overall: consolidating tool taxonomy into `ToolCatalog` removes real duplication and makes the system easier to extend. I’d want the null-safety around validation config and the config-added native validation-tool path verified before merge.

---

PR #122, opened 2026-05-17
feat(tooling): per-bot tool configurations with whitelist enforcement

### LINKED ISSUE #119: AI Created Issue: Available tools per Bot
Originates from #118

## Quality assessment

The issue intent is clear and broadly feasible, and it aligns with existing patterns for MCP configurations and system prompts. However, it is incomplete in a few implementation-critical areas: it references `org.remus.giteabot.agent.tools.ToolCatalog`, which does not appear to exist in the current codebase; it does not define whether tool configurations are global reusable entities like MCP configurations or bot-local settings; and it does not state expected behavior for existing bots after migration. These gaps can be handled as open questions/assumptions without blocking a draft, because the overall requirement is still actionable.

## Summary
Add reusable **Tool Configurations** so admins can control which built-in agent tools are exposed to each bot, similar to how MCP configurations and MCP tool selections are managed today.

## Problem
Today, all built-in tools are made available to the AI, either:
- as native tool calls, or
- via legacy system-prompt instructions.

This exposes tools that are irrelevant for a given repository or bot setup. For example, a Java/Maven project should not need npm or .NET-related tools.

## Goal
Allow admins to define reusable tool configurations and assign exactly one configuration to each bot. Only tools enabled by the selected configuration must be exposed to the AI and executable during a conversation.

### 1. System Settings: Tool Configurations
Add a new **Tools** section under **System settings**.

In this section, an admin can:
- view all tool configurations
- create a new configuration
- edit an existing configuration
- clone an existing configuration

Requirements:
- A non-deletable **default configuration** must exist.
- The default configuration must be created automatically.
- The default configuration must have **all available built-in tools enabled**.

### 2. Tool Configuration editor
When editing a tool configuration, show all registered built-in tools in a selection UI.

Requirements:
- List all registered built-in tools from the application tool catalog/registry.
- Show tool metadata at least including:
  - name / identifier
  - description
  - grouping by purpose/category, if such metadata exists
- The selection UI should match the MCP tool-selection UX as closely as practical, including:
  - filtering
  - paging
  - selection state
  - preview/summary behavior where applicable

Note: the issue originally references `org.remus.giteabot.agent.tools.ToolCatalog`, but that type does not appear to exist in the current codebase. The implementation should use the actual built-in tool registry/catalog used by the application.

### 3. Bot Configuration integration
Add a mandatory **Tool Configuration** selector to the Bot Configuration page.

Requirements:
- Place it alongside the existing MCP Configuration selection.
- Selection is required for every bot.
- The UI should provide a preview/summary similar to the MCP Configuration selector where practical.

### 4. Runtime tool exposure
When a bot runs, only tools enabled by the selected Tool Configuration must be available to the AI.

This applies to both exposure mechanisms currently used by the application:
- native tool-calling
- legacy system-prompt-based tool instructions

Requirements:
- Disabled built-in tools must not be advertised to the model.
- Disabled built-in tools must not appear in any generated legacy prompt/tool instructions.

### 5. Runtime enforcement
Tool execution must be validated against the selected Tool Configuration.

Requirements:
- If the AI attempts to call a built-in tool that is not enabled in the bot's Tool Configuration, execution must be rejected.
- The rejection should fail safely and return a clear tool error/result to the agent flow rather than executing the tool.
- This enforcement must happen even if a disabled tool call is somehow generated despite prompt/tool filtering.

## Data model / backend expectations
Add persistent support for reusable Tool Configurations, analogous to existing MCP configuration handling.

Expected capabilities:
- persist tool configurations
- persist selected tools per configuration
- associate each bot with one tool configuration
- provide a default configuration with all tools enabled

## Migration / compatibility
Existing bots must remain valid after this feature is introduced.

Minimum expectation:
- Existing bots should be assigned the default Tool Configuration during migration or fallback handling.
- After migration, all bots must have a valid Tool Configuration because the selection is mandatory.

## Acceptance criteria
- A new **Tools** section exists under **System settings**.
- A default, non-deletable Tool Configuration exists automatically.
- The default Tool Configuration contains all built-in tools.
- Admins can create, edit, and clone Tool Configurations.
- The Tool Configuration editor lists all available built-in tools with descriptions.
- The editor supports filtering and paging comparable to the MCP tool-selection UI.
- Bot Configuration requires selecting a Tool Configuration.
- Existing bots are assigned a valid Tool Configuration after upgrade.
- For a bot with a restricted Tool Configuration, only enabled built-in tools are exposed to the AI.
- Disabled built-in tools are excluded from both native tool-calling and legacy prompt-based tool instructions.
- If the AI tries to execute a disabled built-in tool, the system rejects the call and does not execute it.
- The default Tool Configuration preserves current behavior by making all tools available.

## Out of scope
- Changing how MCP tool selection works
- Changing individual tool behavior
- Adding new built-in tools

## Implementation notes
- Reuse the existing MCP configuration/tool-selection patterns where possible for entity modeling, UI behavior, and filtering logic.
- Ensure tool filtering affects both prompt construction and actual tool execution paths.
- Be explicit about how built-in tools are identified so that filtering is stable even if display names change.

## Open questions
- What is the canonical built-in tool registry/catalog to use, since the referenced `org.remus.giteabot.agent.tools.ToolCatalog` is not present?
- Do built-in tools already have category/purpose metadata for grouping, or is that grouping unavailable/inferred?
- Should cloning a Tool Configuration also clone the exact enabled/disabled selection snapshot, including metadata shown in the UI? (Assumed yes.)

## Assumptions

- Tool Configurations are reusable global system settings, analogous to MCP configurations, not bot-local one-off settings.
- The default Tool Configuration should preserve current behavior by enabling all built-in tools.
- Existing bots should be migrated or defaulted to the default Tool Configuration so the new mandatory association does not break them.
- This feature applies to built-in application tools, not MCP tools, which already have separate configuration and selection handling.
- Runtime enforcement should reject disallowed tool calls even if the model somehow produces them.

## Open questions

- Which concrete class/service is the source of truth for the built-in tool catalog?
- Is grouping by purpose already supported in built-in tool metadata, or should the first version fall back to a flat list if not available?
- What exact error/result shape should be returned when a disabled tool is requested at runtime?

---

tmseidel:
Introduce reusable, named "Bot tool configurations" that whitelist the built-in agent tools available to a bot, and wire the whitelist through the runtime for both native function-calling and the legacy prompt-based tool protocol.

Highlights
----------
* Data model
  - New `BotToolConfiguration` (named, reusable) and `BotToolSelection` (enabled built-in tool names) JPA entities with repositories.
  - `Bot` gets a mandatory `toolConfigurationId` FK; every bot is always associated with exactly one configuration.
  - Flyway migrations V11/V12 for H2 and PostgreSQL create the tables, backfill a Default configuration and add the FK on `bots`.

* Default configuration
  - `DefaultBotToolConfigurationInitializer` ensures a "Default" entry exists on startup, adds newly shipped built-in tools to it, and protects it from rename/deletion/default-flag changes.

* Catalog & runtime enforcement
  - `BuiltinToolRegistry` exposes the catalog as the single source of truth for selectable built-ins (MCP tools remain out of scope and are still managed via MCP configurations).
  - `ToolCatalog` and `AgentToolRouter` filter advertised and executed tools against the bot's whitelist; `ToolExecutionService` rejects disabled tools at execution time.
  - `SystemPromptAssembler` / `LegacyToolProtocolRenderer` only render whitelisted tools into the legacy prompt protocol.

* Admin UI
  - New "Bot tool configurations" area under System settings (`bot-tools-selection.html`, `bot-tools-form.html`, list entry).
  - Bot form lets operators pick the tool configuration; controller and service layer validate the association.

* Cleanup
  - Remove obsolete `prompts/legacy/*-tool-protocol.md` resources in favor of the renderer-driven output.

* Docs
  - New `doc/BOT_TOOL_CONFIGURATIONS.md`; updates to README, `doc/ARCHITECTURE.md` and `doc/USER_GUIDE.md`.

* Tests
  - Unit tests for `BotToolConfigurationService`, `BotToolSelectionService`, `BuiltinToolRegistry`, `DefaultBotToolConfigurationInitializer`, `AgentToolRouterWhitelistTest`, plus updates to existing controller/service tests to cover the mandatory FK and whitelist behavior.


max-california (commented):
## 🤖 AI Code Review

### Diff chunk 1/3
1. **Blocking issues**

- **Migration/bootstrap ordering looks unsafe for existing installations.**  
  The docs say V12 “backfills every existing bot to the Default configuration created by `DefaultBotToolConfigurationInitializer`”. That initializer runs at application startup, but Flyway migrations run before the application is fully started. If V12 depends on a row inserted by the initializer, upgrades can fail on a real DB because the default config will not exist yet when the migration executes.  
  **Why it matters:** existing instances may fail startup during migration, or the `NOT NULL`/FK step may leave bots unassigned.  
  **Suggested fix:** make the migrations fully self-contained: insert the default configuration and its selections inside Flyway SQL (or in a Java migration), then backfill `bots.bot_tool_configuration_id`, then add the `NOT NULL`/FK constraint. The runtime initializer can still be additive for newly shipped tools, but migration correctness should not depend on it.

- **`BotService.save()` can still persist an invalid bot with `toolConfiguration == null`.**  
  The new fallback is:
  ```java
  botToolConfigurationRepository.findByDefaultEntryTrue().ifPresent(bot::setToolConfiguration);
  return botRepository.save(bot);
  ```
  If the default row is missing for any reason (bad migration, test slice, manual DB damage), the bot is saved with a null mandatory association and fails later at JPA/DB level with a generic exception.  
  **Why it matters:** this turns a domain validation error into a late persistence failure and makes failures harder to diagnose.  
  **Suggested fix:** after the fallback lookup, explicitly fail fast:
  ```java
  if (bot.getToolConfiguration() == null) {
      throw new IllegalStateException("No tool configuration assigned and no default tool configuration exists");
  }
  ```
  Ideally also validate this in controller/service input paths.

- **Potential lazy-loading bug in `cloneConfiguration()`.**  
  `BotToolConfigurationService.cloneConfiguration()` loads the source via `configurationRepository.findById(sourceId)` and then iterates `source.getSelectedTools()`. Unless the collection is eagerly fetched or Open Session in View is relied on, this is fragile and can throw `LazyInitializationException` if called outside the transactional boundary or if service wiring changes.  
  **Why it matters:** clone is part of admin CRUD and should not depend on persistence-context subtleties.  
  **Suggested fix:** fetch selections explicitly, e.g. via repository method with `JOIN FETCH`, or load rows from `selectionRepository.findByConfigurationId(sourceId)` and clone from that list.

- **Be explicit about ordering in `findAll()` for tool configurations.**  
  `BotToolConfigurationService.findAll()` delegates to `findAll()` with no sort. For admin dropdowns and list pages, nondeterministic ordering is awkward and can vary by DB. Consider repository methods ordered by `defaultEntry desc, name asc` or at least `name asc`.

- **`loadSelectedTools()` loses descriptions for known tools.**  
  The selected-tools endpoint returns empty descriptions because `loadSelectedTools()` maps persisted rows directly and sets `description` to `""`. Since the UI “Details” modal appears intended to show grouped built-in tools with descriptions, it would be better to enrich known tools from `BuiltinToolRegistry`, falling back to empty only for persisted-but-unknown tools.

- **Whitelist semantics for `null` vs empty set should stay very clearly separated.**  
  A lot of runtime behavior now depends on:
  - `null` whitelist => no enforcement
  - empty set => all built-ins disabled  
  That’s fine, but it’s easy to regress because `allowedBuiltinTools()` currently returns `Set.of()` for null configuration. Since prod should never pass null, this is probably okay, but I’d consider documenting/asserting that runtime callers must not pass null, because `Set.of()` is stricter than the “legacy/test bypass” semantics described elsewhere.

- **`normalizeWhitespaceForCompare()` uses regex per line.**  
  The new patch matching is a correctness improvement, but this method is called in nested loops and uses `replaceAll`, which recompiles/applies regex repeatedly. For very large files/search windows this can get expensive. Probably not urgent, but if patching large files is common, consider a small non-regex normalizer.

3. **Tests**

Recommended or missing coverage based on the changed code:

- **Migration/integration test for upgrade path**
  - existing bots before V11/V12
  - migration creates default config
  - all existing bots get backfilled
  - application starts successfully with FK/NOT NULL enforced

- **`BotService.save()` failure path**
  - saving without tool configuration when no default exists should raise a clear exception

- **`BotToolConfigurationService.cloneConfiguration()`**
  - clone copies selections even when source was loaded fresh from DB
  - clone name uniqueness when “Copy of X” already exists

- **`BotToolSelectionService.loadSelectedTools()` / controller endpoint**
  - selected built-in tools endpoint returns expected rows and, if intended, descriptions/kinds for known tools

- **`ToolExecutionService` patch-file edge cases**
  - CRLF/LF normalization patch succeeds and preserves dominant line ending
  - whitespace-tolerant unique match succeeds
  - whitespace-tolerant ambiguous match fails
  - trailing newline edge cases don’t corrupt file content

4. **Overall assessment**

The runtime/tooling direction looks solid, and the whitelist enforcement is sensibly applied at multiple layers. The main concern is the migration/bootstrap dependency for the mandatory new FK; that should be made self-contained before merge. Other than that, this is close.

### Diff chunk 2/3
1. **Blocking issues**

- **Potential startup/migration inconsistency around the default configuration**
  - `V12` inserts a default row only if no row with `default_entry = TRUE` exists, and `DefaultBotToolConfigurationInitializer` then loads the default via `findByDefaultEntryTrue().orElseGet(...)`.
  - If the DB ever ends up with:
    - a non-default row named `Default`, or
    - multiple rows with `default_entry = TRUE`
    
    the initializer/service behavior becomes ambiguous and could fail or pick an arbitrary row depending on repository implementation. The migrations also do not enforce uniqueness of `default_entry = TRUE`.
  - This matters because `bots.bot_tool_configuration_id` is backfilled from that row and the whole feature assumes exactly one protected default configuration.
  - **Suggested fix:** enforce the invariant at the DB level where possible (e.g. partial unique index on `default_entry = TRUE` in PostgreSQL; for H2, at least validate at startup and fail fast with a clear error if multiple defaults exist). Also consider resolving by name+flag or repairing invalid states explicitly.

- **`BuiltinToolRegistry` lowercases tool names with default locale**
  - `name.trim().toLowerCase()` uses the JVM default locale. That can mis-normalize identifiers in locales like Turkish. Tool names are protocol identifiers and should be locale-independent.
  - **Suggested fix:** use `toLowerCase(Locale.ROOT)` everywhere tool identifiers are normalized.

- **Unused `Ordered` implementation in bootstrap**
  - `DefaultBotToolConfigurationBootstrap implements Ordered`, but the `ApplicationRunner` bean returned does not itself implement/consume that ordering. So `getOrder()` on the inner class likely has no effect.
  - If runner ordering matters, return an `ApplicationRunner` that also implements `Ordered`, or use `@Order` on the bean.
  - Not necessarily wrong if there is only one relevant runner, but the current code is misleading.

- **Controller catches overly broad `Exception`**
  - New `SystemSettingsController` endpoints catch generic `Exception` for save/delete flows. That can hide programming errors and make user-facing behavior harder to reason about.
  - **Suggested fix:** catch expected validation/state exceptions (`IllegalArgumentException`, `IllegalStateException`) and let unexpected exceptions propagate to the global error handler.

- **UI/JS scalability**
  - `bot-tools-selection.html` renders and filters the full tool list client-side. That is fine today, but if validation tools/MCP-like built-ins grow significantly, DOM rebuilds on every interaction may become less pleasant.
  - Probably acceptable now given current tool counts, just worth keeping in mind.

3. **Tests**

- Add a test for **invalid default configuration states**:
  - multiple `defaultEntry=true` rows
  - existing row named `Default` but `defaultEntry=false`
  - startup behavior when such corruption exists

- Add a test for **locale-stable normalization** of tool names if normalization is kept in `BuiltinToolRegistry`.

- Given the migration importance, consider an integration-style migration test verifying:
  - an existing bot is backfilled to a valid configuration after V11/V12
  - startup initializer populates selections for the SQL-created default row

4. **Overall assessment**

Good change overall: the whitelist enforcement appears thoughtfully layered and the test coverage is solid. I’d want the default-configuration invariant tightened up and the locale-sensitive identifier normalization fixed before merge.

### Diff chunk 3/3
1. **Blocking issues**

- **Potential migration/startup race for existing installations**
  The docs say `V12` backfills `bots.bot_tool_configuration_id` to the Default configuration created by `DefaultBotToolConfigurationInitializer`, but that initializer runs at application startup, not during Flyway migration. If `V12` assumes the `Default` row already exists, upgrades can fail because Flyway runs before Spring beans initialize. Please verify the SQL migration is self-contained and creates/locates the default configuration itself, rather than depending on application startup logic.

- **Whitelist semantics for `null` appear inconsistent**
  The tests here assert `allowedBuiltinTools(null)` returns an empty set, while the docs state: “`null` as the whitelist disables enforcement entirely.” Those are very different semantics unless every caller distinguishes `null` config from empty whitelist explicitly somewhere else. This is a correctness/regression risk: a missing configuration could accidentally disable all built-in tools instead of bypassing enforcement in legacy/test paths. I’d recommend making the contract explicit in code:
  - either return `null`/`Optional<Set<String>>` for “no enforcement”
  - or keep returning a set but ensure all callers treat `null config` as a separate branch.
  At minimum, add tests at the enforcement points (`ToolCatalog`, `AgentToolRouter`, `ToolExecutionService`) for the `null` case.

- **Preserving unknown persisted tools is reasonable, but should be deliberate and documented in service/API behavior**
  The new test `saveSelection_preservesKindOfPersistedButUnknownTool` suggests the service retains previously persisted tool names that are no longer in the built-in registry. That can help during upgrades/renames, but it also means stale/removed tools may live indefinitely in DB/UI state. If intentional, consider documenting this behavior in the service and perhaps adding cleanup/migration strategy for removed tools.

- **Controller tests around flash/model errors could assert the user-visible result more fully**
  Several controller tests verify the redirect/view name but not that an error flash/model attribute is actually populated on redirect cases. Since the feature relies on surfacing validation failures in the UI, it would be useful to assert the relevant redirect attributes as well.

3. **Tests**

Recommended additional coverage:

- **Migration/integration test for upgrade path**
  A Flyway-backed test starting from a pre-V11 schema with existing bots and verifying migration to V12 succeeds and results in non-null `bot_tool_configuration_id` for all bots.

- **End-to-end whitelist enforcement tests for `null` vs empty**
  Add explicit tests showing behavior when:
  - whitelist is `null` / no config available
  - whitelist is empty
  - whitelist contains mixed-case names
  This should be covered at runtime enforcement layers, not just selection service level.

- **Default configuration initializer duplicate/concurrency behavior**
  If `ensureDefault()` can run more than once or across clustered startups, a test around duplicate default creation / unique-name handling would help. Even if the DB unique constraint is relied on, that should be validated.

4. **Overall assessment**

The test additions are generally solid and the feature shape looks good, but I’d want the migration dependency and `null`-whitelist semantics clarified before merge. Those are the main areas that could cause upgrade failures or unintended tool disabling.

---

PR #126, opened 2026-05-18
Feature/review agentic workflows m1

### LINKED ISSUE #124: M1 — Foundation: `PrWorkflow` SPI + `ReviewWorkflow` extraction
### Goal
Refactor the existing PR-review path into a pluggable architecture **without
changing any observable behaviour**.

### Tasks

**1.1 SPI**
- Create `PrWorkflow` interface in `org.remus.giteabot.prworkflow`:
  ```
  String key();                       // e.g. "review"
  String displayName();
  PrWorkflowCategory category();      // REVIEW | TESTING | SECURITY | DOCS | CUSTOM
  WorkflowParamsSchema paramsSchema();
  WorkflowResult run(PrWorkflowContext ctx);
  default boolean supportsCallback() { return false; }
  default WorkflowResult onCallback(PrWorkflowContext ctx, CallbackPayload payload) { ... }
  ```
- Create `PrWorkflowContext` (immutable record): bot, pull request payload, repo
  client, AI client, `PrWorkflowRun` id, allowed built-in tools set, MCP catalog,
  deployment target (nullable), workflow params (JSON node).
- Create `PrWorkflowRegistry` (`@Service`) — Spring DI auto-collects all
  `PrWorkflow` beans; lookup by `key()`. Mirrors `AiProviderRegistry`.
- Add `PrWorkflowCategory` enum and `WorkflowResult` (status + summary +
  artifacts list).

**1.2 Extract `ReviewWorkflow`**
- Move the body of `BotWebhookService.reviewPullRequest(...)` into a new
  `ReviewWorkflow implements PrWorkflow` under
  `org.remus.giteabot.prworkflow.review`. Keep all existing chunking, retry,
  comment-posting logic.
- `BotWebhookService.reviewPullRequest(...)` becomes a thin call to
  `PrWorkflowOrchestrator.run(bot, payload, "review")`.

**1.3 Orchestrator**
- New `PrWorkflowOrchestrator` (`@Service`):
  - Resolves the bot's `WorkflowConfiguration` (M2 adds the column; for now,
    default to `["review"]`).
  - Persists `PrWorkflowRun` rows (status transitions).
  - Catches exceptions per workflow, persists `FAILED` + first 2 KB of stack
    trace into `PrWorkflowStep.logExcerpt`.
  - Emits Micrometer metrics.
- Concurrency: per `(botId, repoOwner, repoName, prNumber, workflowKey)` only
  one `RUNNING`/`WAITING_DEPLOY` row at a time. On a new PR-synchronize event,
  cancel the previous run (transition to `CANCELLED`).

**1.4 Persistence**
- Flyway `V13__prworkflow_runs.sql` creates tables `pr_workflow_runs`,
  `pr_workflow_steps` (see ER diagram in concept doc).
- JPA entities `PrWorkflowRun`, `PrWorkflowStep`; repositories with finders by
  `(botId, repoOwner, repoName, prNumber, status)`.
- `PrWorkflowRunService` with `start()`, `appendStep()`, `complete()`,
  `fail()`, `cancel()`.

**1.5 Tests**
- `PrWorkflowOrchestratorTest`: registry lookup, exception handling, metrics,
  cancel-on-resync semantics.
- `ReviewWorkflowTest`: assert byte-for-byte identical comment payload as the
  pre-refactor `BotWebhookService` for a fixture diff.
- Regression: enable the full existing `BotWebhookServiceTest` suite —
  expectation is **zero changes** to those tests.

**1.6 Docs**
- New `doc/PR_WORKFLOWS.md` (operator-facing top-level doc).
- Update `doc/ARCHITECTURE.md` § "Components" with a paragraph + diagram node
  for `PrWorkflowOrchestrator` + `PrWorkflowRegistry`.

### Acceptance criteria
- All existing tests pass unchanged.
- The legacy PR-review behaviour now flows through the orchestrator.
- `pr_workflow_runs` rows are written for every PR review.
- New metrics visible at `/actuator/prometheus`.

tmseidel:
## What's new

- **PrWorkflow SPI**: New interface for registrable PR follow-up workflows
  - `PrWorkflow.key()` — stable lowercase kebab-case identifier
  - `PrWorkflow.run(PrWorkflowContext)` — execution contract
  - `PrWorkflowRegistry` — auto-discovery via Spring DI, key validation
  - `PrWorkflowOrchestrator` — single entry point for all PR workflows
  - `PrWorkflowContext` — immutable run context (bot, payload, runId, step appender)
  - `WorkflowResult` — outcome type (SUCCESS, FAILED, SKIPPED, WAITING_DEPLOY)

- **Persistence (Flyway V13)**:
  - `pr_workflow_runs` — track each workflow invocation across (bot, repo, PR, workflow)
  - `pr_workflow_steps` — append-only progress log per run
  - Cancel-on-resync: supersede in-flight runs when PR receives new push

- **ReviewWorkflow (first implementation)**:
  - Extracted legacy code-review path into PrWorkflow
  - Behaviour byte-for-byte identical — zero breaking changes
  - Wrapped CodeReviewService via new CodeReviewServiceFactory

- **Observability**:
  - Micrometer: `prworkflow.run_total{workflow,status}` counter
  - Micrometer: `prworkflow.run_duration_seconds{workflow}` timer
  - Exception handling: capture, persist as FAILED, rethrow for admin error tracking

- **Documentation**:
  - doc/PR_WORKFLOWS.md — SPI overview, lifecycle, writing new workflows
  - doc/ARCHITECTURE.md — updated with PrWorkflowOrchestrator component

## Changes

- `org.remus.giteabot.prworkflow.*` — new package (SPI, registry, orchestrator, entities, services)
- `org.remus.giteabot.prworkflow.review.*` — ReviewWorkflow + CodeReviewServiceFactory
- `BotWebhookService` — `reviewPullRequest()` now delegates to orchestrator
- Flyway migrations: `V13__prworkflow_runs.sql` (H2 + PostgreSQL)

## Testing

- 19 new unit tests (PrWorkflowRegistry, PrWorkflowOrchestrator, PrWorkflowRunService, ReviewWorkflow)
- Full regression suite: 560/560 tests pass
- BotWebhookServiceTest: 33/33 pass (zero behaviour change)

## Backwards compatibility

✅ Fully backwards compatible
- Existing bots unaffected; ReviewWorkflow runs by default via orchestrator
- Legacy code-review handlers (bot commands, inline replies, reviews) continue unchanged
- Flyway V13 schema-only; no data migration needed


max-california (commented):
## 🤖 AI Code Review

1. **Blocking issues**

- **Race condition in cancel-on-resync can allow duplicate active runs/comments**
  - `PrWorkflowRunService.start()` does `cancelActiveRunsForPr(...)` and then inserts a new run, but there’s no locking or uniqueness guarantee at the DB level. Two webhook deliveries for the same `(bot, repo, pr, workflow)` arriving close together can both observe no active rows (or the same active rows), both create `RUNNING` rows, and both proceed to post reviews.
  - This matters because the feature’s core promise is “cancel superseded in-flight runs”; without serialization, duplicate reviews are still possible under concurrent async webhook handling.
  - Suggested fix:
    - Add a DB-backed concurrency control strategy, e.g.:
      - pessimistic lock on rows selected for the tuple, or
      - a dedicated per-PR/workflow lock table, or
      - a partial unique index for active statuses (Postgres can do this; H2 may need a different approach) plus retry logic.
    - At minimum, document/verify how concurrent deliveries are prevented elsewhere if that already exists.

- **Cancelled runs are not actually prevented from continuing to side effects**
  - `cancelActiveRunsForPr()` marks older runs `CANCELLED`, but neither `PrWorkflowOrchestrator` nor `PrWorkflowContext.appendStep(...)` checks whether the current run has since been superseded. If an older async run is already inside `ReviewWorkflow`, it will still call `reviewPullRequest(...)` and `postReviewAction(...)` after being marked cancelled.
  - That undermines the stated goal of preventing “racing comments/reviews against an outdated diff”.
  - Suggested fix:
    - Add cooperative cancellation checks before/after major workflow steps and before external side effects. For example, expose `runService.getById(runId).getStatus()` or `isActive(runId)` and abort if no longer active.
    - For `ReviewWorkflow`, check cancellation at least before posting the review action, and ideally before invoking the review itself if feasible.

- **Registry validation does not enforce kebab-case, only lowercase**
  - `PrWorkflowRegistry.index()` rejects non-lowercase keys, but it still accepts values like `review_1`, `review!`, or `review test`, despite the docs/API contract saying “lower-case kebab-case”.
  - Suggest validating against an explicit regex like `^[a-z0-9]+(?:-[a-z0-9]+)*$`.

- **`appendStep()` step ordering is not safe under concurrent writes**
  - `stepOrder` is derived from `run.getSteps().size()`. If steps are appended concurrently for the same run, duplicate or out-of-order values are possible.
  - If workflows are strictly single-threaded today this may be acceptable, but future callback/log streaming work mentioned in docs will likely make this real.
  - Consider deriving the next order in the database (`max(step_order)+1`) or using a timestamp/ID-only ordering instead.

- **Minor terminology mismatch in lifecycle docs vs implementation**
  - Docs mention `QUEUED`, but `start()` immediately creates runs in `RUNNING`. That’s not necessarily wrong, but the documentation and status model should match the actual behavior to avoid confusion.

3. **Tests**

- Add a test covering **cancellation behavior during an in-flight run**, not just that prior rows are marked `CANCELLED`. Right now the most important behavioral guarantee isn’t verified.
- Add a concurrency-oriented test (service/integration level if possible) for **two near-simultaneous starts** of the same `(bot, repo, pr, workflow)` tuple.
- Add a registry test for **invalid non-kebab-case lowercase keys** (e.g. `review_test`, `review.test`) so the contract is enforced.
- Consider an integration test verifying the Flyway/JPA mapping around `pr_workflow_runs` / `pr_workflow_steps`, since this PR introduces new persistence entities and migrations.

4. **Overall assessment**

Good structural refactor and the extraction is easy to follow, but the cancellation/concurrency story looks incomplete for a feature whose main new behavior is “cancel superseded in-flight runs.” I’d fix that before merge.

---

max-california (commented):
## 🤖 AI Code Review

1. **Blocking issues**

- **Docs still describe a `QUEUED` state that the implementation may not have**
  - In `doc/PR_WORKFLOWS.md`, the lifecycle diagram still says `[*] -> QUEUED: PrWorkflowRunService.start(...)`, but in the earlier implementation notes/tests you described runs as starting immediately in `RUNNING`.
  - If the code still creates `RUNNING` directly, this is now inaccurate documentation for a newly introduced persistence/lifecycle API and will mislead future implementors.
  - Fix either the docs or the code/status model so they match exactly.

- **Potential API/behavior mismatch around cancellation semantics needs verification**
  - The new `PrWorkflowContext.requireActive(...)` and `WorkflowCancelledException` are a good addition, but the important question is whether `ReviewWorkflow` now actually calls `requireActive(...)` before external side effects.
  - If `ReviewWorkflow` still performs `reviewPullRequest(...)` / `postReviewAction(...)` without these checks, the original race remains: superseded runs can still post reviews/comments after being marked cancelled.
  - Since the diff is truncated before `ReviewWorkflow`, I can’t confirm. Please verify and add/adjust the implementation before merge if not already done.

- **Locking strategy correctness cannot be assumed from `PrWorkflowRunLockManager` alone**
  - `PrWorkflowOrchestrator` now depends on `PrWorkflowRunLockManager`, which suggests the concurrency issue is being addressed, but the diff is truncated before the actual use.
  - A lock manager only fixes the duplicate-run issue if it is applied around the full critical section for the same `(bot, repo, pr, workflow)` key and is safe in the actual deployment model. If it’s only in-memory, it won’t protect multi-instance deployments.
  - Please verify:
    - the lock wraps cancel + create + transition into execution,
    - the key includes all relevant dimensions,
    - behavior is acceptable in clustered deployments, or this limitation is explicitly documented.
  - If this is only JVM-local and the app can run multiple instances, that remains a blocking correctness issue.

- **Enforce kebab-case in code, not just in docs/Javadoc**
  - The new docs and `PrWorkflow.key()` Javadoc clearly say “lowercase, kebab-case identifier”. If `PrWorkflowRegistry` still only checks lowercase/uniqueness, it should validate against a regex such as `^[a-z0-9]+(?:-[a-z0-9]+)*$`.
  - This is low-effort and prevents persisted invalid keys from creeping into config/migrations later.

- **Be careful with metric tag cardinality and normalization drift**
  - `PrWorkflowMetrics.normalise()` lowercases and trims but doesn’t otherwise constrain values. That’s probably fine for workflow/status values controlled by code, but if any future workflow keys become externally configurable, it would be safer to validate at registration time and keep metrics tags strictly bounded there.
  - Not an immediate issue if registry validation is strict.

- **Deprecation path in `BotWebhookService` is reasonable, but dead code should be short-lived**
  - The deprecated private `createCodeReviewService(Bot)` wrapper is fine for transition, but since this class is already being refactored, it’d be good to remove it as soon as remaining handlers are migrated to avoid keeping two mental models around.

- **Docs mention “orchestrator already deduplicates concurrent in-flight runs”**
  - In `PrWorkflow` Javadoc, that statement is fairly strong. If deduplication is only best-effort or single-node scoped, soften the wording to avoid overpromising.

3. **Tests**

- Add/confirm a test that a **superseded in-flight `ReviewWorkflow` does not post review side effects** once cancellation is detected.
- Add/confirm a test for the new lock manager covering **two concurrent starts for the same tuple** and asserting only one active run proceeds.
- If `PrWorkflowRunLockManager` is in-memory, add a test and documentation clarifying its scope; if clustered deployment is supported, add an integration test or design note for cross-node behavior.
- Add a registry test for **non-kebab-case lowercase keys** if not already present.
- Add tests for `PrWorkflowContext.requireActive(...)` / `WorkflowCancelledException` to ensure orchestrator maps cancellation to `CANCELLED` rather than `FAILED`.

4. **Overall assessment**

The update moves in the right direction: cooperative cancellation and explicit locking are exactly the right areas to address. I’d still want confirmation in code/tests that cancellation actually blocks review side effects, and that the locking approach is correct for the deployment model, before calling this ready to merge.

---

PR #127, opened 2026-05-18
docs(prworkflow): document M2 workflow configurations and tighten reg…

### LINKED ISSUE #125: M2 — Workflow Configurations CRUD + Bot Assignment
### Goal
Operators can create reusable workflow configurations (named whitelists of
workflow keys + params) and assign them to bots.

### Tasks

**2.1 Persistence**
- Flyway `V14__workflow_configurations.sql`:
  - `workflow_configurations(id, name UK, default_entry, created_at, updated_at)`
  - `workflow_selections(id, workflow_configuration_id FK, workflow_key, params_json)`
    UNIQUE `(workflow_configuration_id, workflow_key)`
  - `bots.workflow_configuration_id` (nullable FK)
- Entities + repositories under `org.remus.giteabot.prworkflow.config`.

**2.2 Services**
- `WorkflowConfigurationService`: CRUD + clone (mirrors
  `BotToolConfigurationService`).
  - Guard: cannot delete a configuration in use by any bot.
  - Guard: cannot delete or rename the `defaultEntry`.
- `WorkflowSelectionService`: add/remove/update selections, JSON-schema
  validation of `params_json` against `PrWorkflow.paramsSchema()`.
- `DefaultWorkflowConfigurationInitializer` (`ApplicationRunner`): on startup,
  ensure a `Default` configuration exists with `review` enabled. Additively
  enable newly registered workflows on each startup (never auto-enable
  workflows whose `category != REVIEW` — operators must opt in).

**2.3 Orchestrator integration**
- `PrWorkflowOrchestrator.runAll(bot, payload)` iterates over the bot's
  `WorkflowConfiguration → WorkflowSelection` rows and invokes each
  `PrWorkflow` sequentially. Order: stable by `workflow_key`.
- If `Bot.workflowConfigurationId` is null, behave like Default.

**2.4 Controllers + UI**
- `WorkflowConfigurationController` under `/system-settings/workflow-configurations`:
  list, new, edit, delete, clone, "select workflows" sub-page (mirror MCP UX).
- Thymeleaf templates under `src/main/resources/templates/system-settings/workflow-configurations/`.
- Per-workflow params editor: dynamic form fields rendered from
  `paramsSchema()` (JSON Schema → simple HTML inputs).
- Update `bots/new.html` and `bots/edit.html`: new dropdown
  **Workflow configuration** + **Details** modal showing enabled workflow keys
  and their params (secrets masked).
- Backfill on upgrade: Flyway sets `workflow_configuration_id` of all existing
  bots to the Default row created by the initializer.

**2.5 Tests**
- `WorkflowConfigurationServiceTest`: CRUD, guards, clone.
- `WorkflowConfigurationControllerTest`: MockMvc happy-path + validation errors.
- `DefaultWorkflowConfigurationInitializerTest`: idempotency, additive enable.

**2.6 Docs**
- `doc/PR_WORKFLOWS.md` § "Workflow configurations".
- Screenshot for `doc/screenshots/prworkflow/workflow-configurations.png`.

### Acceptance criteria
- Operators can create/clone/edit/delete workflow configurations via the UI.
- New bots default to the Default configuration; existing bots are backfilled
  on startup.
- The orchestrator iterates over selected workflows in deterministic order.

tmseidel:
Wrap up milestone M2 ("Workflow Configurations CRUD + Bot Assignment") by aligning the developer documentation with the shipped code and by simplifying the PrWorkflowRegistry bootstrap so package-external tests can exercise it without reflection.

Registry
- Move PrWorkflow indexing from @PostConstruct into the constructor. The registry is now eagerly consistent the moment Spring (or a test) hands it the list of PrWorkflow beans; no package-private hook is needed anymore.
- Drop the jakarta.annotation.PostConstruct import.
- Update PrWorkflowRegistryTest, PrWorkflowOrchestratorTest, DefaultWorkflowConfigurationInitializerTest and WorkflowSelectionServiceTest to stop calling the removed index() method; rejection tests now wrap the constructor call in assertThrows.

Docs
- doc/PR_WORKFLOWS.md
  - Update status banner: M1 + M2, available since 1.7.0-preview1.
  - Refresh the components mermaid diagram: add WorkflowSelectionService and the workflow_configurations / workflow_selections tables; drop the "M1 — only workflow" caption on ReviewWorkflow.
  - New section "Workflow configurations (M2)" covering:
    * data model (ER diagram, V14 migration, unique constraint),
    * services (WorkflowConfigurationService, WorkflowSelectionService, WorkflowParamsValidator, DefaultWorkflowConfigurationInitializer with REVIEW-only auto-enable policy),
    * orchestrator dispatch via runAll(...) with deterministic ordering, per-workflow failure isolation and skip-with-WARN for unknown keys,
    * admin UI touchpoints (system-settings list/edit/clone/delete, workflows sub-page, bot form dropdown + JSON-backed Details modal),
    * upgrade behaviour (column added nullable by Flyway, one-time backfill performed by the initializer on first startup).


max-california (commented):
## 🤖 AI Code Review

### Diff chunk 1/2
1. **Blocking issues**

- **Workflow parameter form does not preserve persisted values on edit**
  
  In `workflows.html`, the inputs are populated from `field.defaultValue`, not from `row.paramsJson` / the saved selection. That means editing an existing configuration will render defaults/blank values instead of the actual persisted params, and saving can unintentionally overwrite existing settings (especially bad for non-default text/integer values and secrets).

  Why it matters:
  - Existing workflow params are effectively invisible in the UI.
  - A no-op edit can mutate configuration.
  - Secret fields are especially problematic: they render blank, and if selected/saved, may clear an existing secret unless the service preserves old values.

  Suggested fix:
  - Parse `row.paramsJson` server-side into a map and expose per-field current values in `WorkflowSelectionRow` (or a view model).
  - Render current values preferentially, falling back to `field.defaultValue` only when no persisted value exists.
  - For `SECRET`, preserve the existing value unless the user explicitly changes it.

- **Checkbox boolean params can be saved incorrectly due to duplicate request params being collapsed**
  
  The form posts both:
  - a hidden `params__...=false`
  - a checkbox `params__...=true`

  But `saveWorkflowSelection(..., @RequestParam Map<String, String> allParams, ...)` only captures a single value per key. With duplicate parameter names, Spring binds one value arbitrarily/implementation-specifically; you can't reliably know whether the checkbox was checked. This can cause checked booleans to persist as `false`.

  Suggested fix:
  - Accept `MultiValueMap<String, String>` or `HttpServletRequest` and handle repeated params explicitly.
  - Or avoid the hidden+checkbox pattern here and encode booleans differently.

- **Manual JSON construction in `WorkflowConfigurationController.toJson()` breaks typed params**
  
  `extractWorkflowParams()` serializes all submitted field values as JSON strings. Although `WorkflowParamsValidator` later coerces them, this relies on a lossy intermediate encoding and makes booleans/numbers/string edge cases harder to reason about. Combined with the checkbox issue above, it increases risk of incorrect persisted values.

  Suggested fix:
  - Build a typed `Map<String, Object>` and call `WorkflowParamsValidator.validateAndCanonicalise(Map<String, ?>, schema)` directly, or use `ObjectMapper` instead of manual string concatenation.
  - If the service must accept JSON strings, still use `ObjectMapper.writeValueAsString(...)` rather than hand-rolled JSON.

- **`DefaultWorkflowConfigurationInitializer.backfillExistingBots()` saves one bot at a time**
  
  This is fine for small datasets, but it’s an N-update startup path. If installations can have many bots, consider collecting changed bots and using `saveAll(...)`.

- **`WorkflowConfigurationService.cloneConfiguration()` depends on lazy-loaded selections**
  
  It iterates `source.getSelectedWorkflows()` after `findById`. Because the service is transactional, this likely works, but it’s somewhat implicit. A dedicated fetch query (or loading selections via repository) would make it more robust/readable.

- **`runAll()` silently returns empty when a configuration has no enabled workflows**
  
  This may be intended, but it’s worth verifying operationally. A bot configured with zero workflows will effectively ignore PR events. If that’s allowed, maybe log at `info`/`warn` instead of `debug` so misconfiguration is easier to spot.

3. **Tests**

Recommended missing coverage:

- **Workflow selection form round-trip with existing params**
  - existing persisted string/integer/boolean params render in the form
  - saving without changes preserves them

- **Boolean param submission**
  - checked checkbox persists `true`
  - unchecked checkbox persists `false`

- **Secret param preservation**
  - existing secret remains unchanged when editing unrelated fields
  - explicit replacement updates it

- **`PrWorkflowOrchestrator.runAll()`**
  - falls back to `ReviewWorkflow.KEY` when `bot.workflowConfiguration == null`
  - continues after one workflow throws
  - skips unknown workflow keys and still runs the rest
  - respects deterministic ordering from `enabledWorkflowKeys(...)`

4. **Overall assessment**

Good direction overall: moving registry indexing into the constructor is a clean simplification, and the workflow-configuration model fits the existing patterns well. However, the workflow-params UI/save path has correctness issues around persisted values and boolean handling that should be fixed before merge.

### Diff chunk 2/2
1. **Blocking issues**
- None evident from the provided diff chunk.

- `WorkflowSelectionServiceTest.saveSelection_validatesParams_andPersistsCanonicalJson` uses `ArgumentCaptor.forClass(Iterable.class)`, which works but is fairly weakly typed and can hide mistakes. If practical, capture a concrete collection type or assert the saved iterable contents more explicitly to make the test less brittle.
- In `WorkflowConfigurationServiceTest.cloneConfiguration_copiesNameAndSelections`, consider also asserting whether the clone is actually persisted (if `cloneConfiguration(...)` is supposed to save it rather than just return a detached copy). The current test makes the intended contract a bit ambiguous.
- The doc update mentions deterministic ordering, skip-on-WARN for unknown keys, and default/backfill behavior. Those are important behavioral guarantees; it would be helpful if the tests explicitly named those behaviors where they already exist so the contract stays clear.

3. **Tests**
- Good additions overall around:
  - workflow configuration save/delete/clone invariants
  - selection validation and canonicalization
  - controller redirect/error behavior
- Recommended if not already covered elsewhere in the omitted files:
  - a test that `WorkflowConfigurationController.saveWorkflowSelection(...)` actually forwards only `params__<workflowKey>__...` inputs into the expected per-workflow JSON map, not just that `saveSelection(...)` is invoked
  - service-level coverage for unknown/orphaned workflow keys being preserved/listed appropriately and skipped safely at execution time
  - explicit coverage for the default configuration protections described in docs: cannot delete, cannot rename, cannot clear default flag, and new configs cannot self-assign default (most of this is covered already)

4. **Overall assessment**
- This diff looks solid. The tests added here cover the main invariants introduced by M2 well, and I don’t see a correctness or security concern in the provided changes.

---

PR #129, opened 2026-05-19
feat(prworkflow): add M3 deployment targets + async callback channel

### LINKED ISSUE #128: M3 — Deployment Targets (Webhook + Static) + Callback Endpoints
Provide the deployment SPI and the two simplest strategies (webhook trigger + static preview URL). Add the async callback channel.

**Key Deliverables:**
- `DeploymentStrategy` interface (methods: `trigger()`, `poll()`, `teardown()`, `awaitsCallback()`)
- `WebhookTriggerStrategy` (signed POST with HMAC-SHA256; awaits callback)
- `StaticPreviewUrlStrategy` (template resolution; optional readiness probe)
- `WorkflowCallbackController` endpoints:
  - `POST /api/workflow-callback/{runId}/{secret}` (verify HMAC, resume orchestrator)
  - `POST /api/workflow-log/{runId}/{secret}` (optional log stream append)
- Flyway `V15__deployment_targets.sql` (tables: `deployment_targets` with encrypted `config_json`)
- New UI: **System settings → Deployment Targets** (list, new, edit, delete, per-strategy form)
- Updated bot form: dropdown **Deployment target** + **Details** modal
- Callback URL recipe snippet (copy-to-clipboard) + operator docs

**Tests:**
- `WebhookTriggerStrategyTest` (WireMock-based)
- `StaticPreviewUrlStrategyTest` (template + probe)
- `WorkflowCallbackControllerTest` (HMAC validation, replay protection)

**Documentation:**
- `doc/PR_WORKFLOWS.md` § "Deployment targets" (JSON schemas, HMAC recipe)
- `doc/PR_WORKFLOWS_WEBHOOK_RECIPES.md` (Jenkins, GitLab, Argo, GitHub Actions)

**Definition of Done:**
- A test webhook callback completes a `PrWorkflowRun` transition
- Forged callbacks rejected with 401
- Operators can copy/paste the callback URL into their CI system

tmseidel:
Introduces the deployment SPI and two no-extra-deps strategies (WEBHOOK, STATIC) so PR workflows can request a per-PR preview environment from an operator-managed CI/CD pipeline and resume on an inbound callback. No production workflow consumes this yet; the E2ETestWorkflow in M4 will be the first caller.

Persistence (Flyway V16, H2 + PostgreSQL)
- deployment_targets(id, name UK, strategy_type, config_json, preview_url_template, timeout_seconds, created_at, updated_at)
- bots.deployment_target_id nullable FK
- pr_workflow_runs: preview_url, callback_secret, deployment_handle_json

Deployment SPI (org.remus.giteabot.prworkflow.deployment)
- DeploymentStrategy, DeploymentStrategyType (WEBHOOK | STATIC | CI_ACTION | MCP), DeploymentRequest, DeploymentResult, DeploymentStatus
- DeploymentStrategyRegistry: discovers strategies by typeKey, enforces uniqueness
- DeploymentOrchestrator: trigger + await/poll loop, persists handle JSON, bounded by per-target timeoutSeconds
- DeploymentCallbackNotifier: in-process SynchronousQueue keyed by runId (documented multi-instance caveat in doc/PR_WORKFLOWS.md)

Strategies
- WebhookTriggerStrategy: POSTs signed JSON envelope (HMAC-SHA256 via X-AI-Bot-Signature), returns PENDING, awaits async callback
- StaticPreviewUrlStrategy: resolves URL template ({prNumber}, {sha}, {branch}) and optionally probes /healthz until reachable; returns READY synchronously

Inbound callback channel
- WorkflowCallbackController exposes POST /api/workflow-callback/{runId}/{secret} (status + previewUrl) and POST /api/workflow-log/{runId}/{secret} (4 KB-truncated log chunks); constant-time secret comparison + optional HMAC body verification; no-op (HTTP 409) once the run reached a terminal status

Services + Admin UI
- DeploymentTargetService: CRUD, encrypts config_json via EncryptionService, guards name uniqueness, blocks delete when any bot still references the target, exposes M3 strategy subset to the UI
- DeploymentTargetController under /system-settings/deployment-targets: new/edit/save/delete, surfaces the per-instance callback URL pattern
- New card on the System settings landing page (list + actions)
- Bot form: optional "Deployment target" dropdown; BotController binds deploymentTargetId

Project conventions
- ObjectMapper is instantiated directly (`private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper()`), never injected, in all new deployment classes.

Tests (+39, total 624 → all green)
- WebhookTriggerStrategyTest (8): missing config rejection, signed payload shape + headers, custom headers, non-2xx → FAILED, IO error → FAILED, HMAC determinism
- StaticPreviewUrlStrategyTest (7): missing template, immediate READY without probe, placeholder resolution, probe success after retries, probe timeout → FAILED
- WorkflowCallbackControllerTest (12): unknown run / wrong secret / bad signature → 401, terminal-state replay → 409, bad JSON / missing / unknown status → 400, READY persists previewUrl + notifies, FAILED marks run failed, log endpoint truncates >4 KB
- DeploymentTargetServiceTest (12): encrypt-on-save / decrypt-on-read, name uniqueness on insert + update, defaults, delete blocked while referenced, available strategy subset
- SystemSettingsControllerTest: constructor updated for new dependency

Docs
- doc/PR_WORKFLOWS.md: new "Deployment targets (M3)" section with lifecycle state diagram, per-strategy JSON schemas, callback contract, HMAC verification recipe, multi-instance caveat, V16 upgrade notes; status banner bumped to 1.7.0-preview2
- doc/PR_WORKFLOWS_WEBHOOK_RECIPES.md (new): Jenkins, GitLab CI, GitHub Actions repository_dispatch, Argo CD ApplicationSet recipes; common-pitfalls cheat sheet


max-california (commented):
## 🤖 AI Code Review

### Diff chunk 1/2
1. **Blocking issues**

- **Callback HMAC is verified with the wrong secret**
  `WorkflowCallbackController.verifySignature(...)` uses `run.getCallbackSecret()` as the HMAC key, but the outbound webhook docs and `WebhookTriggerStrategy` use the target’s configured `sharedSecret` to sign requests. A CI system following the documented recipe will sign the callback with `sharedSecret`, so verification here will always fail when the header is present.
  - Why it matters: valid callbacks get rejected with 401, breaking the async flow.
  - Fix: verify callback signatures with the deployment target’s configured shared secret (decrypting target config as needed), or explicitly change the contract/docs to say callbacks must be signed with `callbackSecret` and update the outbound payload/examples accordingly.

- **Run state is left in `WAITING_DEPLOY` for immediate failures/rejections**
  In `DeploymentOrchestrator.requestDeployment(...)`, any non-null `handleJson` causes `runService.markWaitingDeploy(...)` before status handling. If `trigger()` returns `FAILED`/`REJECTED` with a handle, the run is persisted as `WAITING_DEPLOY` and then returned to the caller without being restored/completed.
  - Why it matters: persisted state can become incorrect and misleading, and later logic may treat the run as still awaiting a callback.
  - Fix: only call `markWaitingDeploy(...)` when the result is actually `PENDING`, or explicitly complete/fail the run for terminal trigger outcomes.

- **Async success callback does not transition the run out of `WAITING_DEPLOY` unless an orchestrator thread is currently waiting**
  `WorkflowCallbackController.callback(...)` persists preview URL for `READY`, but does not persist the status transition back to `RUNNING`; it only notifies the in-process queue. If no thread is waiting (documented multi-instance case, restart, or callback arrives slightly after timeout), the run remains `WAITING_DEPLOY`.
  - Why it matters: the DB state can get stuck indefinitely even though the deployment is ready.
  - Fix: persist the state transition in the controller as well (e.g. `resumeFromDeploy(...)` on `READY`) before/independent of notifier delivery.

- **`DeploymentTargetService.findAll/findById` mutates managed entities with decrypted secrets in a read-only transaction**
  Both methods load entities from JPA and then `decryptInPlace(target)`. Because these are still managed entities, later flushes in the same persistence context can accidentally write decrypted plaintext back to the DB.
  - Why it matters: this can silently defeat the “encrypted at rest” guarantee.
  - Fix: return detached copies/DTOs for reads, or explicitly detach entities before mutating, or use attribute conversion/listener-based encryption rather than in-place mutation.

- **Template form contains nested `<form>` tags**
  `templates/system-settings/deployment-targets/form.html` nests a delete `<form>` inside the main save `<form>`, which is invalid HTML and can lead to broken submission behavior depending on browser.
  - Why it matters: save/delete actions may not behave reliably in the admin UI.
  - Fix: close the main form before rendering the delete form, or place the delete button in a separate sibling form.

- `DeploymentTargetService.deleteById()` scans `botRepository.findAll()` to check references. That’s workable now, but it won’t scale well. Prefer a repository method like `existsByDeploymentTargetId(...)` or `countByDeploymentTargetId(...)`.

- There’s a property-name inconsistency:
  - `DeploymentTargetController` uses `ai-git-bot.public-base-url`
  - `DeploymentOrchestrator` uses `app.public-url`
  This is likely to confuse operators and produce different callback URLs between UI and runtime. Consider standardizing on one property.

- `StaticPreviewUrlStrategy` directly substitutes `{branch}` into URLs. Branch names can contain `/` and other characters that aren’t safe in hostnames/paths depending on template placement. At minimum, document constraints clearly; better, encode/sanitize placeholders appropriately or support separate raw/url-encoded placeholders.

- `WebhookTriggerStrategy` uses `Math.clamp(...)`, which is only available on newer JDKs. If this project targets an older LTS/runtime, that will break compilation. Worth verifying against the project’s configured Java version.

3. **Tests**

Recommended missing coverage for meaningful behavior changes:

- `DeploymentOrchestrator` tests for:
  - `FAILED`/`REJECTED` trigger results not leaving the run in `WAITING_DEPLOY`
  - timeout behavior and persisted run status
  - `READY` synchronous result persistence

- `WorkflowCallbackController` tests for:
  - successful `READY` callback persisting state transition when no waiter is present
  - callback signature verification using the intended secret source
  - replay behavior for a run stuck in `WAITING_DEPLOY`

- `DeploymentTargetService` test proving decrypted reads cannot be re-persisted as plaintext (or, after fix, a test around returned detached copies).

- UI/controller test for the deployment target form rendering/submission path, since the nested form issue is easy to miss in unit tests.

4. **Overall assessment**

Good feature slice overall, and the added tests are solid around strategy behavior and callback validation. But there are a few correctness issues in the async callback/state flow and one important secret-handling mismatch that should be fixed before merge.

### Diff chunk 2/2
1. **Blocking issues**

- **Callback auth reuses the same secret for URL path auth and HMAC verification**
  - From the tests and PR description, `callback(42L, "topsecret", "sha256=...", body)` uses the run’s stored `callbackSecret` both as the `{secret}` path segment and as the HMAC key. That means the path itself exposes the HMAC secret to anything that can observe URLs (reverse-proxy access logs, browser history if ever invoked manually, monitoring tooling, etc.). Once leaked, the optional body signature adds no security because it’s keyed by the same value already present in the URL.
  - This materially weakens the callback channel.
  - **Suggested fix:** separate these concerns:
    - keep a random opaque callback token for the URL path, and
    - use a distinct shared HMAC secret from the deployment target config for body signing/verification.
    - If you want backward compatibility, accept unsigned callbacks when no HMAC secret is configured, but never derive HMAC trust from the path token.

- **401 for unknown run may complicate operator debugging**
  - Returning 401 for both “unknown run” and “bad secret” is defensible to avoid run enumeration, but it will make diagnosing legitimate callback issues harder.
  - Consider logging these cases with enough context (runId, remote addr if available, signature present/absent) at warn level, while keeping the external response generic.

- **Replay/idempotency behavior should be verified for duplicate READY callbacks**
  - The description says terminal-state replays return 409, but `READY` appears to notify an in-process waiter rather than explicitly persisting terminal completion in the controller. That may be fine if the orchestrator transitions the run afterward, but it leaves a small uncertainty around duplicate callbacks arriving before the waiter updates the run status.
  - **Suggested fix:** if not already handled atomically elsewhere, consider guarding callback acceptance with a compare-and-set style transition on the run status.

- **Direct `new ObjectMapper()` in new deployment classes is brittle**
  - The PR notes this follows project convention, so not blocking. Still, for security-sensitive request signing/parsing flows, duplicated mapper configuration can drift over time.
  - Consider at least centralizing a shared static helper/utility for these deployment classes so serialization used for signatures and persisted handle JSON stays consistent.

3. **Tests**

- **Add tests for callback secret / signature separation**
  - Given the security concern above, add coverage that the URL token and HMAC secret are distinct and that:
    - correct path token + wrong HMAC fails,
    - wrong path token + correct HMAC fails,
    - correct both succeeds.

- **Add duplicate-callback race/idempotency coverage**
  - Recommended tests for:
    - two READY callbacks in quick succession,
    - READY after FAILED and vice versa,
    - callback arriving after orchestrator timeout/cancel.
  - This is especially important with the in-process notifier and async flow.

- **Add tests for log endpoint terminal-state behavior**
  - The description mentions terminal-state callbacks are no-op/409. I’d also verify whether `/api/workflow-log/{runId}/{secret}` rejects logs once a run is terminal, to avoid post-completion log spam or confusing step history.

4. **Overall assessment**

Good test coverage and the feature shape looks solid overall, but I’d treat the callback auth design as a merge blocker: using the URL path secret as the HMAC key undermines the intended integrity protection. Once that’s separated, this looks close.

---

Issue #123, opened 2026-05-18
Epic: PR-Review Agentic Workflows — Extend reviews with test generation, deployment, and E2E execution


tmseidel:
Currently, AI-Git-Bot's PR-review workflow is linear: fetch diff → send to LLM → post review comment. This epic extends the bot with a pluggable **workflow orchestration system** that lets operators chain multiple follow-up actions:

- **Generate E2E tests** from the PR diff (intelligent test planning).
- **Deploy** to a preview/test environment (callback-driven, strategy-agnostic).
- **Execute tests** against the live preview (using the AI agent + Playwright MCP).
- **Report results** back as PR comments and status checks.
- **Future workflows:** security scans, performance smoke tests, doc diffs, migration plans (enabled by the extensible `PrWorkflow` SPI).

This is **backwards-compatible** — existing bots continue to work unchanged; new workflows are opt-in per bot via the UI.

## Problem Statement

1. **PR reviews are passive:** Today's bot only *comments* on code changes; it doesn't verify the changes work end-to-end.
2. **Testing is manual:** Teams must write E2E tests manually, run them against a deployed version, and reconcile results.
3. **Deployment is operator-specific:** There's no standard way to trigger a preview environment; each company has a different CI/CD setup (Jenkins, GitHub Actions, Argo, Vercel, etc.).
4. **No observability:** Operators have no centralized place to see which workflows ran on which PRs and why they passed or failed.

## Solution Overview

We introduce two new first-class concepts:

1. **`PrWorkflow`** — a reusable, configurable follow-up action that runs on PR open/sync (registry-based SPI, like `AiProviderRegistry`).
2. **`DeploymentTarget`** — an abstraction layer for hooking into existing deployment infrastructure (webhook trigger, GitHub Actions dispatch, MCP tool, or static preview URL).

All new workflows are **built on the existing `AgentLoop` infrastructure** (see AGENT.md) so the AI can iteratively generate, validate, and refine E2E tests using tool calls.

**Key design decisions:**
- The legacy review becomes the first `ReviewWorkflow` (refactored in M1). No behaviour change for existing bots.
- Per-PR test suites live in the database, not the repo. They can optionally "graduate" to the repo via a follow-up PR (M7).
- Deployment is non-invasive: the bot **triggers** existing CI, then **awaits a callback** with the preview URL. No custom deploy logic needed.

---

## Acceptance Criteria (Epic-level)

[...]

PR #135, opened 2026-05-20
docs(m6): mark CI_ACTION deployment shipped + add laptop systemtest scenario

### LINKED ISSUE #133: M6 — CI Action Strategy (GitHub/Gitea/GitLab/Bitbucket native)
Reuse each Git provider's native CI instead of a generic webhook.

**Key Deliverables:**
- New `RepositoryApiClient` methods per provider:
  - `dispatchWorkflow(req)` → provider run id
  - `getWorkflowRun(runId)` → status
  - `getWorkflowRunOutputs(runId)` → extract preview URL
- Per-provider implementations (GitHub/Gitea/GitLab/Bitbucket)
- `CiActionTriggerStrategy` (poll-based, not callback-based)
- `CiActionPoller` scheduled task (scans all `WAITING_DEPLOY` runs)
- Per-provider recipe docs (exact step markup to expose `preview_url` output)

**Tests:**
- `CiActionTriggerStrategyTest` per provider (WireMock fixtures)
- `CiActionPollerTest` (many-in-flight, timeouts, status transitions)

**Documentation:**
- `doc/PR_WORKFLOWS_CI_ACTIONS.md` (GitHub, Gitea, GitLab, Bitbucket recipes)

**Definition of Done:**
- GitHub PR triggers a workflow dispatch → bot polls until complete → consumes preview URL → proceeds to testing

tmseidel:


max-california (commented):
## 🤖 AI Code Review

### Diff chunk 1/7
1. **Blocking issues**

- **README / DockerHub README still claim `CI_ACTION` is planned, not shipped.**  
  The PR title says “mark CI_ACTION deployment shipped”, and new docs (`doc/PR_WORKFLOWS.md`, `doc/PR_WORKFLOWS_CI_ACTIONS.md`, `doc/refactoring/CI_ACTION_DEPLOYMENT_USER_STORY.md`) consistently describe it as shipped in 1.7.0. But both `README.md` and `doc/DOCKERHUB_README.md` still show `CI_ACTION` as `⏳ planned (M6)` / `⏳ ... (planned M6)`. That’s a direct contradiction in the user-facing entry points.  
  **Fix:** update both top-level tables to mark `CI_ACTION` as shipped and link to the new CI actions guide.

- **Several new docs describe stale/incorrect strategy names or milestone states, which will mislead operators.**  
  In `doc/PR_WORKFLOWS_E2E.md`, the “available strategies” text says `STATIC`, `WEBHOOK_TRIGGER`, and the upcoming `MCP_BASED`, which conflicts with the actual enum/docs elsewhere (`STATIC`, `WEBHOOK`, `MCP`, `CI_ACTION`) and with the newly added “shipped” docs. This looks like copied pre-implementation text that no longer matches the product.  
  **Fix:** align `doc/PR_WORKFLOWS_E2E.md` terminology and milestone status with the current implementation/docs.

- **The new conceptual/refactoring doc is marked “not yet implemented” even though the same PR adds docs asserting these features are shipped.**  
  `doc/refactoring/PR_REVIEW_AGENTIC_WORKFLOWS.md` starts with “Concept / architecture sketch — not yet implemented”, but the repository already contains the implementation classes and other docs in this PR refer to M1–M6 as shipped. This is likely to confuse readers and contributors about what is design-only vs. actual behavior.  
  **Fix:** either clearly archive this as historical design context (e.g. “original design sketch, parts now implemented”) or update/remove the “not yet implemented” status banner.

- **Avoid hardcoding old repository URLs in DockerHub docs.**  
  `doc/DOCKERHUB_README.md` links to `https://github.com/tmseidel/anthropic-gitea-bot/...` while the repo is now clearly `ai-git-bot`. If this old path still redirects today, it may still break later and is inconsistent branding. Prefer relative links where possible in repo docs, and for DockerHub-facing absolute links, update them to the current repo slug.

- **Be careful with version/test-count claims in docs.**  
  A few docs include very specific numbers like “Test bar: 798 / 798 green” and dated generated timestamps. These age quickly and become maintenance burdens. Consider reducing to qualitative statements or moving exact counts to release notes instead.

- **Document the single-instance limitation closer to top-level operator docs.**  
  `doc/PR_WORKFLOWS_CI_ACTIONS.md` mentions `CiActionPoller` / `DeploymentCallbackNotifier` are in-process and effectively single-instance. That’s operationally important enough that it may deserve a short note in the main PR workflows or deployment docs too, not only deep in the CI_ACTION strategy page.

3. **Tests**

- No code changes are shown in this chunk, so no blocking test gaps here.
- For docs changes specifically, I’d recommend at least a lightweight link/consistency check in CI if you don’t already have one:
  - broken relative links
  - references to nonexistent files (`systemtest/...`)
  - maybe a simple grep-based guard for stale strategy names (`WEBHOOK_TRIGGER`, `MCP_BASED`) if those are no longer valid terms

4. **Overall assessment**

Good documentation expansion overall, but there are a few important consistency issues across the changed docs that should be fixed before merge—especially the contradictory shipped/planned status for `CI_ACTION` and the stale strategy names in the new E2E doc.

### Diff chunk 2/7
1. **Blocking issues**

- **`@EnableScheduling` is a behavior change in a docs-heavy PR and may activate unrelated scheduled jobs globally.**  
  `GiteaBotApplication` now enables scheduling for the entire app. That will start *all* `@Scheduled` beans, not just the CI action poller. Given this PR title/chunk is mostly documentation, this is a meaningful runtime change and could alter production behavior immediately after deploy.  
  Why it matters: if there are any existing `@Scheduled` methods already present but previously dormant, they’ll now run in all environments. That can cause duplicate processing, unexpected polling load, or side effects.  
  Suggested fix: either:
  - move this code change into the implementation PR that introduces/owns the scheduler, with clear release notes and tests, or
  - gate scheduling behind a property, e.g. `@EnableScheduling` via a config class with `@ConditionalOnProperty`, or property-gate the scheduled beans themselves.

- **`BotWebhookService.handlePrClosed` can skip E2E cleanup if review cleanup throws.**  
  The new `e2eTestPrCloseHandler.onPrClosed(...)` call happens *after* `createCodeReviewService(bot).handlePrClosed(payload)` and only the E2E call is wrapped in `try/catch`. If the review close handler throws, the E2E cleanup won’t run.  
  Why it matters: leaked preview environments / stale test suites / temp workspaces on PR close. The added comment explicitly says this should be best-effort.  
  Suggested fix: wrap both close handlers independently, e.g. call review cleanup in its own `try/catch`, then always attempt E2E cleanup in a second `try/catch`.

- **GitHub/Gitea workflow dispatch run-id discovery is race-prone and can attach to the wrong run.**  
  Both `GitHubApiClient.dispatchWorkflow` and `GiteaApiClient.dispatchWorkflow` infer the run id by recording the “highest run id before” and then polling until a newer highest id appears. This is not scoped to the specific ref/inputs/request and can pick up an unrelated concurrently-triggered run of the same workflow.  
  Why it matters: the deployment poller may monitor someone else’s run, leading to wrong preview URL, false success/failure, or cross-PR contamination.  
  Suggested fix: filter candidate runs by ref/head branch and, if possible, a unique input/correlation value (`runId`) passed in workflow inputs and surfaced in the run metadata/logs. If provider APIs don’t support reliable correlation, this limitation should be explicitly handled and tested.

- **`handlePrComment` only checks E2E slash commands when there is no existing agent session.**  
  In the “session exists” path, PR comments won’t go through `e2eTestSlashCommandHandler.tryHandle(...)`. If that’s intentional, it should be documented; otherwise `@bot rerun-tests` / `@bot regenerate-tests` may behave inconsistently depending on session state. Please verify the intended precedence.

- **Gitea/GitHub dispatch loops block request threads for up to 15s.**  
  The synchronous `Thread.sleep` polling in API clients is simple, but it increases webhook latency and ties up threads under load. If many deployments are triggered concurrently, this can reduce throughput. Consider moving “resolve run id” into the poller with a pending handle state, or using shorter async retries outside the request path.

- **Bitbucket download URL construction should be verified against `cloneUrl` variants.**  
  `buildDownloadUrl()` derives a browser URL from `credentials.cloneUrl()`. If `cloneUrl` includes `.git`, credentials, or an API host rather than the web host, the generated link may be invalid. This may already be fine in your data model, but it’s worth confirming and normalizing if needed.

- **Logging in `handlePrClosed` drops the stack trace.**  
  The new warning logs only `e.toString()`. For operational debugging of leaked resources, a full stack trace would be more useful:
  ```java
  log.warn("[Bot '{}'] E2eTestPrCloseHandler failed", bot.getName(), e);
  ```

3. **Tests**

- Add/ensure a test that **E2E PR-close cleanup still runs when review cleanup throws**.
- Add tests for **dispatch run-id correlation under concurrent workflow runs** for GitHub/Gitea, or at least document/test the current limitation.
- Add tests covering **slash-command handling precedence** in `BotWebhookService`:
  - PR comment with no agent session
  - PR comment with existing agent session
  - top-level bot command path
- Add a test proving that enabling scheduling does not unintentionally start unrelated jobs, or gate the scheduler and test the gate.

4. **Overall assessment**

Mostly solid changes, but there are a few merge-blocking reliability concerns in the runtime code mixed into this docs chunk—especially global scheduling enablement, the PR-close cleanup ordering, and the workflow run-id correlation approach for GitHub/Gitea. I’d fix those before merge.

### Diff chunk 3/7
1. **Blocking issues**

- **`CiActionPoller.lookupTargetForRun()` can resolve the wrong deployment target**
  
  It currently does:
  ```java
  return deploymentTargetService.findAll().stream()
      .filter(t -> t.getStrategyType() == DeploymentStrategyType.CI_ACTION)
      .findFirst()
      .orElse(null);
  ```
  This ignores which deployment target was actually used for the run and just picks the first CI_ACTION target in the system. In an installation with multiple CI_ACTION targets, preview URL fallback can become incorrect or nondeterministic, and a later target edit can affect unrelated in-flight runs.

  **Why it matters:** wrong preview URLs will be posted for the PR, and behavior depends on DB ordering rather than run state.

  **Suggested fix:** persist the deployment target id (or enough immutable preview-template data) on `PrWorkflowRun` / handle JSON at trigger time, and resolve that specific target in the poller. Avoid consulting “current first CI_ACTION target”.

- **`E2eTestPrCloseHandler` deletes suites across repositories/bots for the same PR number**
  
  Cleanup uses:
  ```java
  suiteRepository.findByPrNumberAndLifecycleMode(prNumber, mode)
  ```
  with no bot/repo/run scoping. PR numbers are only unique within a repository, not globally.

  **Why it matters:** closing PR #123 in one repo could delete test suites for PR #123 in another repo/bot if they share the same lifecycle mode.

  **Suggested fix:** scope suite lookup by the runs already resolved for `(botId, repoOwner, repoName, prNumber, workflowKey)` — e.g. delete by `runId in (...)`, or add repository/bot identity to `PrTestSuite` and query on that.

- **`E2eTestPrCloseHandler` may promote suites from other repositories with the same PR number**
  
  Similar issue in:
  ```java
  suiteRepository.findByPrNumberAndLifecycleMode(prNumber, SuiteLifecycleMode.PROMOTE_ON_MERGE)
  ```
  Promotion is then attempted using the closed PR’s `repoOwner/repoName`, even if the suite came from another repo.

  **Why it matters:** this can create follow-up PRs in the wrong repository or fail in confusing ways.

  **Suggested fix:** same as above — only consider suites tied to the runs for this specific PR/repo/bot.

- **`CiActionPoller.publish()` mutates `finishedAt` while run is still `WAITING_DEPLOY`**
  
  The comment says it does not transition status, but it does:
  ```java
  r.setFinishedAt(Instant.now());
  runRepository.save(r);
  ```
  while explicitly leaving status unchanged.

  **Why it matters:** `finishedAt` semantically implies terminal completion. Setting it before the orchestrator resumes creates inconsistent run state (`WAITING_DEPLOY` + finished timestamp), which can break dashboards, duration metrics, timeout logic, or any code assuming `finishedAt != null` means terminal.

  **Suggested fix:** don’t set `finishedAt` here. If you want freshness, add a dedicated `updatedAt` field or only persist `previewUrl`.

- **`DeploymentTargetService` should not instantiate its own `ObjectMapper`**
  
  `private final ObjectMapper objectMapper = new ObjectMapper();`
  
  This bypasses any shared Jackson configuration/modules used elsewhere in the app. Prefer constructor injection of the Spring-managed `ObjectMapper`.

- **`CiActionPoller` does an O(n²) cleanup pass**
  
  ```java
  lastPolledAt.keySet().removeIf(id -> runs.stream().noneMatch(...))
  ```
  Fine for small counts, but easy to make linear by precomputing a `Set<Long>` of active run ids.

- **`E2eTestSlashCommandHandler` pattern is broader than the Javadoc suggests**
  
  The docs mention “literal `@bot` marker OR actual bot alias”, but the regex accepts any `@\S+`. That may be intentional, but if not, it will trigger on mentions of other users/bots. Worth tightening or updating the comment to match behavior.

- **Minor documentation drift in `DeploymentTargetController`**
  
  The class comment now says CI_ACTION will appear automatically “once M6 wires its strategy bean”, but `availableStrategyTypes()` already includes `CI_ACTION`. The comment appears stale.

3. **Tests**

Recommended missing coverage for meaningful behavior changes:

- **`CiActionPoller`**
  - multiple CI_ACTION deployment targets present → verify preview URL fallback uses the correct target for the run
  - terminal READY callback does **not** set `finishedAt` prematurely (or add test for the corrected behavior)
  - runs with missing/invalid handle JSON are skipped safely
  - provider status/output lookup error paths and retry behavior

- **`E2eTestPrCloseHandler`**
  - closing PR in repo A does not delete/promote suites from repo B with the same PR number
  - cleanup/promotion only acts on suites associated with the matching workflow runs
  - merged vs non-merged behavior for `PROMOTE_ON_MERGE`

- **`E2eTestSlashCommandHandler`**
  - command recognition for intended mentions only
  - feedback is passed via `PrWorkflowContext.HINT_E2E_FEEDBACK`

- **`DeploymentTargetService.validateMcpConfig()`**
  - rejects non-whitelisted deploy/status/teardown tools
  - accepts valid config with optional tools omitted

4. **Overall assessment**

There’s useful functionality here, but I see a couple of correctness issues around run/suite scoping and CI_ACTION target resolution that should be fixed before merge. The rest looks reasonable, with good attention to validation and extensibility.

### Diff chunk 4/7
1. **Blocking issues**

- **`WorkspaceProcessRunner` does not provide `BASE_URL` to the test process, but prompts/config assume it exists.**  
  `PrTestWorkspaceManager.scaffoldPlaywright()` config reads `process.env.BASE_URL`, and `E2ePromptLibrary.authorSystemPrompt()` explicitly instructs agents to target that env var. However, `PrWorkflowToolExecutor.prTestRun()` / `WorkspaceProcessRunner.run()` never inject the preview URL into the child process environment. In practice, generated tests will run against the fallback `http://localhost:3000` instead of the deployed preview, which is a correctness-breaking regression for the whole E2E flow.  
  **Fix:** pass the preview URL from `PrWorkflowToolContext.previewUrl()` into the process environment (at least `BASE_URL`, and equivalents if supporting other frameworks).

- **Playwright JSON parsing likely won’t find the report with the configured reporter setup.**  
  In `buildCommand()` you run `playwright test --reporter=json`, which prints JSON to stdout. But the scaffolded `playwright.config.ts` configures the JSON reporter as `['json', { outputFile: 'playwright-report/report.json' }]` plus `['list']`. Depending on Playwright reporter precedence/behavior, stdout may not contain the JSON object that `updatePlaywrightCases()` expects, causing zero per-case updates and `attempted == 0` => error outcome even when tests ran.  
  **Fix:** choose one source of truth and make it consistent:
  - either rely on stdout JSON and remove the config-based JSON reporter/output file, or
  - read `playwright-report/report.json` from disk instead of scraping stdout.

- **`SuitePromotionService` is wired through `GiteaClientFactory`, which appears provider-specific for a feature described as repository-generic.**  
  `promote()` obtains a `RepositoryApiClient` via `giteaClientFactory.getApiClient(bot.getGitIntegration())`. That naming strongly suggests a Gitea-specific factory, while the rest of the codebase has provider-agnostic `RepositoryApiClient` abstractions. If this service is reachable for GitHub/GitLab/Bitbucket bots, it will either fail or create hidden provider coupling.  
  **Fix:** use the repository/provider-agnostic factory used elsewhere for `RepositoryApiClient`, or clearly constrain this feature to Gitea-only with validation and docs.

- **Path sanitization in `SuitePromotionService` is insufficient against nested traversal segments.**  
  `sanitisePath()` strips leading `/`, `./`, and some prefixes, but it does not normalize or reject embedded `../` segments. A generated `PrTestCase.path` like `foo/../../bar.spec.ts` would survive and then be resolved under `baseDir`, potentially escaping `targetDir`. This is especially risky because these paths originate from AI-generated content persisted earlier.  
  **Fix:** normalize with `Path.of(trimmed).normalize()` and reject any absolute path or normalized path starting with `..` (or containing parent traversal after normalization) before resolving/writing.

- **`E2eAgentRunner` comment/docs around error handling are misleading.**  
  The Javadoc says the “exception path returns a `Result` flagged as `budgetExhausted=true`”, but only AI-call exceptions are handled that way. Tool execution exceptions are converted to `"ERROR: ..."` strings by `PrWorkflowToolExecutor`, which is fine, but worth documenting more precisely to avoid confusing future maintainers.

- **`TestPlannerAgent` injects `PrWorkflowToolExecutor` although the planner is tool-free.**  
  It’s harmless, but unnecessary coupling. If this is only to satisfy `E2eAgentRunner` construction, consider allowing a null/no-op executor more explicitly or overloading the runner constructor.

- **`PrTestSuiteRepository.findByPrNumberAndLifecycleMode(...)` may be too broad.**  
  PR numbers are typically only unique within a repository. If this query is used for cleanup/promotion logic, it probably needs repo owner/name (or bot/run scope) to avoid cross-repo collisions.

- **`WorkspaceProcessRunner` truncation is character-based, not byte-based.**  
  `maxOutputBytes` is treated as a character count via `StringBuilder.length()`, which diverges for multibyte UTF-8 output. Probably acceptable for logs, but the naming is inaccurate and could surprise operators.

3. **Tests**

Recommended missing coverage for meaningful behavior in this diff:

- **End-to-end runner env propagation:** verify `pr-test-run` / `WorkspaceProcessRunner` receives `BASE_URL` from `PrWorkflowToolContext.previewUrl()`.
- **Playwright report parsing source:** a test that matches the actual configured reporter behavior, proving per-case statuses are updated from whichever source is intended (stdout vs file).
- **`SuitePromotionService.sanitisePath` / file writing traversal cases:** inputs like `../x.spec.ts`, `foo/../../x.spec.ts`, absolute paths, backslash-based traversal.
- **Cross-provider promotion behavior or validation:** either tests for provider-agnostic client resolution or tests asserting promotion is rejected for unsupported providers.
- **`PrTestSuiteRepository` usage scope:** if used in cleanup/promotion flows, add tests covering two repos with the same PR number to prevent accidental cross-repo matches.

4. **Overall assessment**

There’s good structure here overall—the E2E runner/tooling split is readable and testable—but I see a few correctness issues that would likely break real runs, especially the missing preview URL env injection and the Playwright report/parsing mismatch. I’d fix those before merge.

### Diff chunk 5/7
1. **Blocking issues**

- **Potential compile/runtime issue from new Flyway migrations being added in this PR chunk**
  - The diff adds `V17`, `V18`, and `V19` migrations under both `h2` and `postgresql`, but these version numbers already appear in the repository structure you provided. If these files are truly being added as `new file mode` in the PR, that implies duplicate Flyway versions, which will fail migration validation/startup.
  - Please verify whether these are actually new in this PR or already present on the target branch. If they already exist upstream, this chunk is likely stale/rebased incorrectly and should not re-introduce them. If they are being modified, the diff should be edits, not new files.

- **`ArtifactUploadSupport.linkMarkdown(...)` does not escape the URL, which can break Markdown links or allow malformed rendering**
  - `linkMarkdown` escapes the link text but interpolates `url` directly into `[...](` + url + `)`. URLs containing `)` or whitespace can break Markdown; some providers may also return relative URLs or user-controlled filenames that end up in the rendered link.
  - This matters because the helper is intended to centralize safe link rendering across providers. Right now only some provider tests cover URL encoding for one case.
  - Suggested fix: percent-encode or otherwise sanitize URLs before embedding, or render as autolink `<...>` where possible. At minimum escape `)` and whitespace-sensitive characters consistently.

- **Inline text rendering can prematurely terminate fenced code blocks**
  - `ArtifactCommentRenderer.render(...)` inserts raw `body` inside triple backticks. If the artifact content itself contains ``````, the fence can be closed early and the remainder rendered as normal Markdown.
  - This is a correctness/safety issue for logs and text artifacts because comment rendering becomes corrupted and could unintentionally format arbitrary content.
  - Suggested fix: choose a fence length longer than any backtick run found in the payload (common Markdown strategy), or fall back to indent-based code blocks / summary-only for content containing long backtick runs.

- **Decouple `ArtifactCommentRenderer` from `ArtifactUploadSupport`**
  - `ArtifactCommentRenderer.humanBytes()` delegates to `ArtifactUploadSupport.humanBytes()`, while the class doc says the renderer is reusable/stateless. It’s not a big problem, but it creates an unnecessary dependency between two utility classes. Consider moving `humanBytes` into the renderer as a private helper or a small shared utility.

- **Be explicit about accepted textual MIME types**
  - `looksLikeText(...)` handles a few types, but common textual artifacts like `application/javascript`, `application/problem+json`, `application/ld+json`, `image/svg+xml`, etc. may be treated as non-inline unexpectedly. If this is intentional, fine; otherwise consider broadening the allowlist or keying more off UTF-8 detection for small payloads.

- **`CiActionPollerTest.successWithoutOutputFallsBackToTemplate()` may be masking target selection ambiguity**
  - The test uses `targetService.findAll()` and a single target with a matching strategy type. If production code also resolves fallback preview URL templates by scanning all targets rather than by the run’s actual deployment target, that’s error-prone when multiple CI targets exist.
  - I can’t confirm from this chunk alone whether the implementation does that, but it’s worth verifying the poller resolves the exact target associated with the run, not just “some CI_ACTION target”.

- **Repository interface growth is getting broad**
  - `RepositoryApiClient` now includes artifact upload and CI workflow dispatch/polling concerns in addition to PR/issue/review operations. That may be acceptable, but it’s starting to span multiple distinct capabilities with provider-specific optional support. Longer-term, consider capability-specific sub-interfaces to reduce unsupported-operation paths.

3. **Tests**

- **Add tests for Markdown fence escaping in `ArtifactCommentRenderer`**
  - Case: text artifact containing triple backticks.
  - Expected: comment remains a valid code block and does not break formatting.

- **Add tests for unsafe/broken URL characters in `ArtifactUploadSupport.linkMarkdown(...)`**
  - Cases: `)`, spaces, maybe `#`.
  - This is especially useful since the helper is meant to normalize provider output.

- **Add direct unit tests for `ArtifactCommentRenderer`**
  - The provider tests are good integration-style coverage, but renderer-specific behavior should be tested directly:
    - inline image vs inline text vs summary-only
    - null/blank filename and null payload handling
    - UTF-8 heuristic with NUL bytes / invalid sequences
    - exact byte-threshold boundaries

- **If the migration files are intentional, add/verify migration validation coverage**
  - At least one integration test that boots Flyway against both H2 and PostgreSQL migration sets would catch duplicate-version regressions immediately.

4. **Overall assessment**

Good direction overall: the new tests add useful coverage, and the repository abstraction changes are coherent with the CI/deployment work. However, I’d block on verifying the Flyway migration version situation and on hardening the Markdown rendering helpers (`linkMarkdown` URL handling and fenced code rendering) before merge.

### Diff chunk 6/7
1. **Blocking issues**

- **The README / DockerHub docs appear to claim `CI_ACTION` is still planned, despite the PR title saying it’s shipped.**  
  In both `README.md` and `doc/DOCKERHUB_README.md`, the `CI_ACTION` deployment strategy is still shown as `⏳ planned (M6)` / `(planned M6)`. If this PR is meant to “mark CI_ACTION deployment shipped”, these docs are inconsistent and will mislead users. Please update those status rows to match the actual shipped state.

- **The new systemtest docs reference a likely wrong internal hostname/port for the sample app.**  
  In `systemtest/README-ci-action.md`, step 3 says `previewUrlTemplate = http://sample-e2e-app:3000`, but the compose file defines:
  - service name: `sample-e2e-app`
  - container name: `ai-git-bot-sample-e2e-app`
  - host port mapping: `3030:3000`
  
  This only works if the bot is on the same Docker network and resolves the *service name* `sample-e2e-app`. The doc earlier says that’s the intended setup, but later examples/logs also mix `sample-e2e-app:3000` and `localhost:3030`. Please make the networking assumption explicit at the field/value table, otherwise operators running the bot outside Docker will configure a broken preview URL.

- **Potential compile risk in `StubAiClient`: `switch` on `Object` with type patterns requires newer Java settings.**  
  `StubAiClient.toJson(...)` uses:
  ```java
  switch (v) {
      case null -> ...
      case Number n -> ...
      case Boolean b -> ...
      case List<?> list -> ...
      default -> ...
  }
  ```
  Pattern matching for switch depends on the project’s Java level/toolchain. If the repo is not already on a version that supports this in your build configuration, tests won’t compile. Please verify the project source/target level supports it; otherwise replace with plain `if/else instanceof` to keep tests portable.

- **A few test comments/assertions are slightly misleading.**  
  In `PlaywrightTestSuiteRunnerTest.capsJourneysAtMaxTestCases()`, the assertion comment says “exactly 2 tool calls”, but the code only checks `isGreaterThanOrEqualTo(2)` and counts author invocations rather than tool calls. Consider tightening the assertion or rewording the comment to avoid future confusion.

- **The in-memory repository stub is fairly heavy and duplicates patterns.**  
  `InMemorySuiteRepo` implements a large repository surface with many unsupported methods. If this pattern is starting to repeat, it may be worth extracting a small reusable fake in test utilities to reduce maintenance burden.

- **Docs should consistently distinguish “service name inside Docker network” vs “host port from outside Docker”.**  
  This comes up in the new `README-ci-action.md` and `README-mcp-deployment.md`. The docs are mostly correct, but a short repeated note near each URL value would make operator setup less error-prone.

3. **Tests**

- Good coverage was added across the new E2E flow pieces. A few worthwhile additions:
  - **Slash command parsing edge cases**: bot-name case differences, punctuation-only suffixes, multiple bot mentions, and commands embedded in longer prose.
  - **Summary renderer escaping** beyond pipe characters: backticks / angle brackets / markdown table-breaking content in titles.
  - **Artifact upload/rendering**: filenames with brackets/parentheses and non-UTF8 text payloads.
  - **Promotion service**: behavior when `createPullRequest(...)` fails/throws after a successful push.

4. **Overall assessment**

- The added tests are substantial and targeted, which is good.  
- Before merge, I’d fix the documentation inconsistencies around `CI_ACTION` status and verify the Java compatibility of `StubAiClient`’s pattern-switch usage.

### Diff chunk 7/7
1. **Blocking issues**

- **`sample-e2e-app` image will fail to start in Docker**
  - `systemtest/sample-e2e-app/Dockerfile` does not run `npm install`, but it also doesn’t need npm packages. However, the image’s default command is `node server.js`, which is fine; the problem is that the Dockerfile copies only `package.json` and `server.js`, so that part is okay. The actual risk is subtler: there is no `RUN npm install`, but unlike the other sample servers this one has no dependencies, so that’s not inherently broken.
  - The blocking part is instead the **compose build context path**: in `systemtest/docker-compose-mcp-deployment.yml`, `sample-e2e-app` uses `context: ./sample-e2e-app`. Because the compose file itself lives under `systemtest/`, Compose resolves relative paths from the compose file’s directory in most normal invocations, so this is likely fine. But the preceding diff hunk shows another compose snippet being added without its filename, and if that file is not also under `systemtest/`, the same relative path pattern may break. Please verify all newly added compose files resolve contexts correctly from their own location.
  - If any of these compose files are meant to be runnable from repo root exactly as documented, use explicit stable paths or validate with CI.

- **Mock CI Action server may not satisfy the bot’s expected API surface**
  - The comment says it implements the subset used by `GitHubApiClient.dispatchWorkflow / getWorkflowRun / getWorkflowRunOutputs`, but the server only exposes:
    - dispatch
    - workflow runs list
    - run status
    - healthz
  - There is **no endpoint for workflow run outputs/artifacts/jobs/logs**, despite the comment explicitly naming `getWorkflowRunOutputs`.
  - If the production code now relies on outputs for `CI_ACTION` deployments, this laptop scenario will fail or give misleading docs. Either:
    - add the missing endpoint(s), or
    - update the documentation/comment to match the actual methods used by the CI_ACTION path.

- **Add basic request size / body validation to sample servers**
  - `sample-ci-action-server` and `sample-mcp-deploy-server` are just system-test helpers, so this is not a product security concern, but they are exposed on host ports. A tiny bit more validation would avoid confusing failures:
    - validate `workflow`, `owner`, `repo` presence/shape
    - reject malformed JSON cleanly
    - validate `RUN_DURATION_MS`, `DEPLOY_DELAY_MS`, `PORT` are finite non-negative integers

- **Document memory-only behavior more prominently**
  - Both sample servers keep all state in memory. That’s fine for demos, but worth calling out in the README/walkthrough as a limitation: restarting the container loses runs/deployments and can confuse polling/teardown behavior.

- **Potential resource cleanup in MCP sample**
  - In `sample-mcp-deploy-server`, a fresh `McpServer` + `StreamableHTTPServerTransport` is created for every `/mcp` request. That’s okay for a tiny demo, but if this is used repeatedly during local testing it would be safer to explicitly close/dispose the `server` as well if the SDK requires it. I’m not sure from the diff whether `transport.close()` is sufficient; worth verifying against the MCP SDK lifecycle expectations.

- **Use exact dependency versions or lockfiles for reproducibility**
  - The new sample Node projects use caret ranges and don’t include lockfiles. For system-test/demo assets, reproducibility matters more than freshness. Consider committing `package-lock.json` and using `npm ci`.

3. **Tests**

- Since this chunk adds runnable system-test infrastructure rather than app code, unit tests are less critical, but I’d recommend at least:
  - a lightweight CI smoke check that each new Docker image builds successfully
  - a smoke test for `sample-ci-action-server` covering:
    - dispatch returns `204`
    - latest workflow run can be listed
    - run transitions from `in_progress` to `completed`
    - fail-next path returns `failure`
  - a smoke test for `sample-mcp-deploy-server` covering:
    - tool discovery
    - deploy returns `handle` or `previewUrl`
    - status transitions to ready
    - teardown removes the deployment

- Also worth verifying docs/examples against reality:
  - if README and DockerHub docs now mark `CI_ACTION` as shipped, ensure there is at least one integration or end-to-end test for the actual `CI_ACTION` deployment flow in the main Java code path, not just the sample server.

4. **Overall assessment**

Good addition overall: the sample infra is useful and the docs move in the right direction. Before merge, I’d verify the CI_ACTION mock actually matches the bot’s current API expectations and that the documented compose scenarios build/run exactly as written.

---

PR #136, opened 2026-05-20
feat(e2e/promotion): add PromotedSuiteGarbageCollector for stale promoted suites

### LINKED ISSUE #134: M7 — Suite Promotion Workflow (optional terminal feature)
Let per-PR generated test suites graduate from ephemeral to permanent (repo-level) status.

**Key Deliverables:**
- New `E2ETestWorkflow` param: `suiteLifecycle` ∈ {`ephemeral`, `offer-as-pr`, `promote-on-merge`}
- `SuitePromotionService`:
  - `offer-as-pr`: create follow-up PR with tests under `tests/e2e/pr-{n}/`
  - `promote-on-merge`: on parent PR merge, create PR against default branch (conflict handling: suffix numeric suffix)
- Nightly GC job (delete orphaned `PrTestSuite` rows after 30 days)
- UI: radio group for suite lifecycle in workflow params; link to generated follow-up PR on workflow-run detail view
- Security notes in docs (promoted tests run in standard CI; may need manual secret review)

**Tests:**
- `SuitePromotionServiceTest` (branch creation, conflict handling, idempotency)
- Integration: PR open → tests pass → merged → follow-up PR created exactly once

**Documentation:**
- Extend `doc/PR_WORKFLOWS_E2E.md` (lifecycle modes, security note)

**Definition of Done:**
- Merged PR with `promote-on-merge` results in exactly one follow-up PR containing generated tests
- Ephemeral run leaves no DB rows or temp files after PR close

tmseidel:


max-california (commented):
## 🤖 AI Code Review

### Diff chunk 1/2
1. **Blocking issues**

- None in the provided diff. This chunk is documentation-only, and I don’t see a merge-blocking correctness or security issue from the changes shown.

- `doc/DOCKERHUB_README.md`: the link text still says **“PR Workflows roadmap + user stories”** while now pointing at `doc/agentic-workflows/README.md`, which appears to be broader feature docs rather than a roadmap. Consider renaming the label to avoid stale wording.
- `doc/PR_WORKFLOWS_E2E.md`: there’s no blank line after the H1 before body text. Not important functionally, but some Markdown renderers/linting setups are picky.
- Several docs now reference placeholder screenshots under `doc/screenshots/prworkflow/...` marked `(TODO)`. If those files are not added in the follow-up, users may hit dead references once the placeholders become real links/images. Worth verifying before release.

3. **Tests**

- No code changes in this chunk, so no additional automated test coverage seems necessary here.
- If the repo has any docs-link validation in CI, it would be good to run it against the renamed `doc/refactoring/` → `doc/agentic-workflows/` paths, since this PR updates many internal links and deletes old files.

4. **Overall assessment**

- Looks good for this diff chunk. The documentation update is coherent, and the path migration from `doc/refactoring/` to `doc/agentic-workflows/` appears consistently applied in the files shown.

### Diff chunk 2/2
1. **Blocking issues**

- **Potential data loss: GC is not scoped to E2E workflow runs.**  
  `PromotedSuiteGarbageCollector` queries all `PrWorkflowRun` rows with a non-null `followUpPrNumber` and old `finishedAt`, then deletes suites by `runId`. Today that may effectively only match E2E promotion runs, but the repository method itself does not constrain by `workflowKey`. If any other workflow later starts using `followUpPrNumber`, this job could unexpectedly delete its attached suites/artifacts.  
  **Suggested fix:** either add a repository method scoped to the E2E workflow key (e.g. `findByWorkflowKeyAndFollowUpPrNumberIsNotNullAndFinishedAtBefore(...)`) or explicitly filter in the GC before deleting.

- **GC is not transactional; partial cleanup is possible.**  
  `collectOnce()` performs multiple read/delete operations across runs without `@Transactional`. If deletion fails partway through, some suites are deleted and others remain, leaving a partial cleanup state. That may be acceptable operationally, but for scheduled maintenance code this usually should be explicit. Also, if `PrTestSuite.cases` relies on JPA cascade/orphan semantics, transactional boundaries matter for consistent persistence behavior.  
  **Suggested fix:** annotate `collectOnce()` or the class with `@Transactional`, or document/handle partial-failure semantics explicitly with per-run try/catch and logging.

- **Consider batching/deleting by query for scalability.**  
  The current GC does 1 query for stale runs and then 1 query per run for suites, followed by `deleteAll(suites)`. For large histories this becomes N+1-ish. If this job could process many promoted runs, a bulk repository delete/find by `runId in (...)` would be more efficient.

- **Defensive validation of retention config would help.**  
  If `prworkflow.e2e.promotion.retention` is configured as zero/negative, `cutoff = now.minus(retention)` could behave unexpectedly and delete too aggressively.  
  **Suggested fix:** validate retention > 0 at construction/startup.

- **The test seam is useful, but naming implies test-only mutability in production bean.**  
  `setNowSupplierForTest` is package-private, which is okay, but it still mutates runtime behavior on a singleton bean. Not a bug, just worth noting. A `Clock` dependency would be a more idiomatic and less stateful alternative.

3. **Tests**

- Add a test proving the GC only considers the intended workflow type / promotion runs, especially if `followUpPrNumber` is not guaranteed to be exclusive to E2E forever.
- Add a failure-path test for deletion exceptions to define expected behavior: continue with later runs vs fail the whole pass.
- If you keep non-transactional behavior, add a test/documentation around partial cleanup semantics.
- Good to see explicit tests for cutoff calculation and preserving `PrWorkflowRun`.

4. **Overall assessment**

Good addition overall, and the `SuitePromotionService` switch hardening is a nice correctness improvement. The main thing to fix before merge is tightening the GC scope so it only targets the intended E2E promotion runs.

---

PR #139, opened 2026-05-21
fix(e2e): harden full-stack QA pipeline, normalize workflow params, polish UX

tmseidel:


max-california (commented):
## 🤖 AI Code Review

### Diff chunk 1/3
1. **Blocking issues**

- **Dockerfile may fail at runtime/build because `curl` is used in `HEALTHCHECK` after switching to non-root, but not guaranteed available in minimal final PATH context**
  - `curl` is installed earlier, so this is probably fine, but what is more concerning is the healthcheck target: `http://localhost:8080/actuator/health`.
  - I don’t see any indication in this diff that Spring Actuator is enabled/exposed in the runtime image/profile. If `/actuator/health` is not available in the `docker` profile, containers will be marked unhealthy even though the app works.
  - **Fix:** verify actuator is present and exposed on that path in the docker profile, or switch the healthcheck to an endpoint guaranteed to exist.

- **`WorkflowSelection.params` is marked `fetch = FetchType.EAGER`, which is a risky regression for a one-to-many collection**
  - Changing from a single text column to a child table is reasonable, but eagerly loading a collection on every `WorkflowSelection` fetch can create unnecessary joins / row multiplication and makes repository/service behavior more fragile as usage grows.
  - This is especially risky because selection lists are used in several UI/service paths and could now silently become much more expensive.
  - **Fix:** prefer `LAZY` and fetch explicitly where needed via query/entity graph. If EAGER is intentional, it should be justified and covered by tests around the affected repository/service access patterns.

- **Potential provider-specific bug in `E2eTestSlashCommandHandler`: reaction client is hardwired through `GiteaClientFactory`**
  - The class now acknowledges slash commands by creating a client via `GiteaClientFactory` and calling `addReaction(...)`. That looks wrong in a codebase that supports GitHub/GitLab/Bitbucket too.
  - If this handler is used outside Gitea-triggered flows, the reaction path may break or use the wrong API implementation.
  - The payload type imported here is also `org.remus.giteabot.gitea.model.WebhookPayload`, which suggests this may be Gitea-specific, but that should be verified before merge.
  - **Fix:** either confirm/limit this handler to Gitea-only usage, or resolve the client through the provider-agnostic `RepositoryProviderRegistry` / integration type instead of a Gitea factory.

- **Dockerfile: deleting any UID/GID 1000 account/group is pretty aggressive**
  - This can unexpectedly remove a base-image account with files outside `/home/appuser`. It’s pragmatic, but brittle.
  - Consider using a non-conflicting app UID/GID, or create the user with `useradd -u 1000 -o` only if you truly need 1000 specifically.

- **Dockerfile: external installer scripts/repos increase supply-chain risk**
  - `curl ... | bash` for NodeSource and live GPG key retrieval for k6 are common, but they do reduce reproducibility and trustability.
  - Consider pinning repo keys more explicitly or documenting the trust model. Not a blocker, but worth tightening.

- **`WorkflowSelectionService.saveSelection`: relying on DB-level cascade should be verified**
  - The code comments say child param rows are removed by DB-level `ON DELETE CASCADE`, but the JPA mapping itself doesn’t enforce that. If migrations/schema don’t match, deletes could fail or leave orphans.
  - Worth double-checking the migration in the remaining diff chunks.

- **`WorkflowParamsValidator.typed(...)` passes unknown keys through unchanged**
  - That matches the comment, but now that persistence is normalized into rows, this may keep stale/unknown params around indefinitely for downgraded workflows.
  - That may be intended; if so, fine. Otherwise consider pruning unknown keys at save time for registered workflows, which you mostly already do.

- **`NarratedToolCallParser` is useful, but best-effort parsing should stay tightly scoped**
  - It’s recovering model failures by executing text-described tool calls, which is reasonable here, but it’s also a trust boundary expansion: assistant text is now executable input.
  - Since execution still goes through `PrWorkflowToolExecutor`, this is somewhat contained. I’d still recommend making sure the executor validates tool names/args strictly and only allows the already-whitelisted E2E tools.

3. **Tests**

- Add/verify tests for the new **workflow params storage model**:
  - save/update existing selection preserves encrypted/raw values correctly
  - replacing params deletes removed keys
  - cloning configuration copies child param rows, not shared references
  - `maskSecrets(...)` works with row-based params
  - controller extraction for `params.<workflow>.<field>` works and no longer breaks on Thymeleaf naming

- Add tests for the **explicit flush fixes**:
  - `DeploymentTargetService.save(...)` updates an existing target correctly
  - `WorkflowSelectionService.saveSelection(...)` can replace an existing selection with the same workflow key without unique-constraint failure

- Add tests for **E2E workflow UX/runtime fixes**:
  - `E2ETestWorkflow` posts the “starting” comment before long-running execution
  - `findByIdWithCases(...)` avoids lazy init issues in the summary path
  - slash-command handler adds reaction only when comment metadata exists, and failures are swallowed

- Add tests for **legacy/narrated tool-call recovery**:
  - `LegacyToolProtocolRenderer.renderE2eAgent(...)` output shape
  - `E2eAgentRunner` legacy positional args are mapped back to named args correctly, especially array/object args
  - `NarratedToolCallParser` parses:
    - fenced JSON form
    - `<tool_call>` JSON form
    - XML `<invoke>/<parameter>` form
  - recovery path only executes `pr-test-write` and handles malformed blocks safely

- Add an integration/system test for the **Gitea actions endpoint fix**
  - ensure `listRecentRuns(...)` queries `/actions/runs?workflow=...` and can resolve dispatched runs on Gitea

4. **Overall assessment**

Promising set of fixes with several clearly useful corrections, especially around Gitea CI dispatch, workflow-param normalization, and E2E agent prompt/tooling alignment. The main things I’d want resolved or explicitly verified before merge are the provider-specific reaction handling, the new `EAGER` one-to-many fetch choice, and runtime validity of the new container healthcheck.

### Diff chunk 2/3
1. **Blocking issues**

- **Potentially destructive user deletion in the Dockerfile**
  - This block removes any existing UID/GID 1000 user/group from the base image:
    ```dockerfile
    if getent passwd 1000 >/dev/null; then
        userdel -r "$existing_user" ...
    fi;
    ```
  - That is risky in a general-purpose base image because `userdel -r` can remove the user’s home and owned files. If the image/vendor ever ships a real service user at 1000, this may delete files unexpectedly and break the image in hard-to-debug ways.
  - Safer fix: create `appuser` with the next free UID/GID, or only reuse the existing UID/GID owner rather than deleting it. If stable 1000 is required, prefer checking specifically for the known placeholder user/group names before deleting.

- **`app.public-url` defaulting to a concrete host IP in `application.properties`**
  - Setting:
    ```properties
    app.public-url=http://172.17.0.1:8080
    ```
    as the default app config is not safe for general environments. It is Linux/Docker-specific, wrong on many hosts, and can break callback URL generation outside the documented local system-test scenario.
  - This is especially concerning because it changes runtime behavior globally, not just in Docker/systemtest.
  - Fix: move this to `application-docker.properties`, compose env, or systemtest-specific config. The default in main `application.properties` should remain neutral or derived from request/base-url config.

- **DB migration compatibility risk for existing installations**
  - The migrations add `workflow_selection_params` and stop using `params_json`, but the shown migration files do **not** include a data backfill from existing `workflow_selections.params_json` into the new child table.
  - If deployed onto an instance that already has workflow params persisted, those params may disappear logically after upgrade unless another omitted change/migration handles the conversion.
  - This is a backward-compatibility/data-loss concern and should be verified before merge.
  - Fix: add explicit migration(s) for H2/Postgres to read existing `params_json` and populate `workflow_selection_params`, then keep or drop `params_json` only once migration is complete.

- **`prTestWrite` path restriction is good, but verify repository schema/path expectations**
  - Restricting writes to `tests/`, `cypress/e2e/`, etc. is a good hardening step. However, it may reject legitimate project layouts if the framework/scaffold or existing tests use another conventional directory.
  - Suggest verifying that all supported E2E flows and generated plans always target these prefixes, especially for Playwright projects that commonly use `e2e/`.

- **`runNpmInstall` scales poorly per workspace**
  - Installing packages into each workspace during scaffolding can become expensive on repeated runs and under concurrency, even with cache warmup.
  - Not necessarily wrong, but worth watching. If this path is hit often, consider a shared immutable scaffold/template or symlink/copy strategy from a pre-provisioned directory.

- **Logging may expose too much model/tool content**
  - New warnings/debug logs include full tool args and combined process output in some places:
    ```java
    log.warn("... args={} -> {}", call.args(), abbreviate(result));
    log.debug("Result: ExitCode: {}, Combined output: {}", ...)
    ```
  - Since test content and command output can contain secrets/tokens/app data, consider redacting known sensitive keys or reducing payload verbosity in non-test environments.

- **`seenSpecFiles.contains(...)` is O(n) in a loop**
  - In `updatePlaywrightCases`, this should probably be a `LinkedHashSet<String>` instead of `List<String>` since it is used only for uniqueness + later logging.

3. **Tests**

- Good additions overall around:
  - workflow param persistence/refactor
  - narrated tool-call recovery
  - Playwright report file parsing
  - preview URL env propagation
  - slash-command reaction behavior

- **Recommended missing coverage**
  - **Migration/backfill tests** for existing `params_json` data -> `workflow_selection_params`.
  - **Config tests** proving `app.public-url` behavior does not break non-Docker/default deployments.
  - **Executor tests** for allowed test-path prefixes per framework, especially if `e2e/` or other layouts are expected.
  - **Workspace scaffolding tests** for repeated runs ensuring `writeAlways(...)` and per-workspace `npm install` don’t regress behavior across reused workspaces.

4. **Overall assessment**

Useful hardening changes with solid test coverage in the Java code, but I’d block on the unsafe Docker user-deletion logic, the global `app.public-url` default, and the unverified migration path for existing workflow params. Once those are addressed/confirmed, this looks much closer to merge-ready.

### Diff chunk 3/3
1. **Blocking issues**

- **Hard-coded Gitea runner registration token in `systemtest/docker-compose-local-gitea.yml`**
  ```yaml
  GITEA_RUNNER_REGISTRATION_TOKEN=${GITEA_RUNNER_REGISTRATION_TOKEN:-IExQ2fu4RH4zQzxR2oJWxYByJp18zqFZvx7QA0og}
  ```
  This bakes what looks like a real registration token into the repo and also causes the runner to auto-register with that token by default. Even if it’s only intended for local system tests, committing credential-like material is risky and teaches an unsafe pattern. It also makes startup behavior surprising for users who expect registration to require explicit setup.

  **Suggested fix:** remove the fallback entirely and require the env var to be provided explicitly for auto-registration, e.g.:
  ```yaml
  - GITEA_RUNNER_REGISTRATION_TOKEN=${GITEA_RUNNER_REGISTRATION_TOKEN:-}
  ```
  and gate registration in entrypoint/config/docs accordingly.

- **Dockerfile installs NodeSource package without refreshing apt metadata**
  After the initial `apt-get update`, the image deletes `/var/lib/apt/lists/*`. The subsequent step:
  ```dockerfile
  RUN curl -fsSL https://deb.nodesource.com/setup_22.x | bash - && \
      apt-get install -y --no-install-recommends nodejs && \
  ```
  runs `apt-get install` without a fresh `apt-get update`. Since the package lists were already removed, this is prone to failing depending on what the NodeSource script does on the current base image. It’s not safe to rely on the external script for apt cache state.

  **Suggested fix:** add `apt-get update` before installing `nodejs`:
  ```dockerfile
  RUN curl -fsSL https://deb.nodesource.com/setup_22.x | bash - && \
      apt-get update && \
      apt-get install -y --no-install-recommends nodejs && \
      rm -rf /var/lib/apt/lists/*
  ```

- **Dockerfile build reproducibility / supply-chain**
  Several installs pull latest content from remote bootstrap scripts/repos (`nodesource`, `rustup`, apt feeds, pip). That may be intentional, but it makes container builds less reproducible and can break unexpectedly over time. Consider pinning more aggressively (or at least documenting expected versions for Node, k6, Rust, pytest dependencies).

- **`gpg -k` in Dockerfile appears unnecessary**
  In the k6 install step:
  ```dockerfile
  RUN gpg -k && \
      gpg --no-default-keyring ...
  ```
  `gpg -k` doesn’t seem to contribute to the setup and may fail/noise in minimal environments. It looks safe to remove unless it’s intentionally warming up GPG state.

- **Runner comments vs config mismatch**
  In `systemtest/gitea-runner/config.yaml`, the comment says:
  ```yaml
  # Run jobs in privileged mode ...
  privileged: false
  ```
  The comment is misleading as written. Worth correcting to avoid confusion during troubleshooting.

- **Potentially destructive UID/GID cleanup in Dockerfile**
  The runtime image deletes any existing user/group with uid/gid 1000. That may be necessary for Temurin Noble, but it’s a fairly invasive step and could become fragile if base-image internals change. If you keep this approach, I’d at least add a short note in docs/changelog since it affects downstream image extension assumptions.

3. **Tests**

- I don’t see test coverage corresponding to the Dockerfile/container behavior changes. For changes this operational, automated tests may be hard, but at minimum I’d recommend verifying in CI that the built image can:
  - start successfully,
  - run `playwright --version`, `cypress --version`, `dotnet --info`, `k6 version`,
  - execute as `appuser`.

- For the systemtest compose changes, a smoke-check in docs or CI would be useful to validate that the added runner sidecars don’t break baseline startup of the local Gitea/GitLab stacks.

4. **Overall assessment**

Good documentation/systemtest improvements overall, but I’d block on the committed runner token and the potentially broken Node.js install step in the Dockerfile before merge.

---

PR #149, opened 2026-05-22
fix: add dark mode toggle and structural dark theme support

### LINKED ISSUE #146: Implement dark mode
**Current behavior**
The current UI theme is light, the user's system-preference about dark/light mode is not requested.

**Expected behavior**
Based on the system-preference of the client the system should switch to the correct theme

PenceaV:
Fixes #146 

A structural dark theme switch in the navbar with persistent localStorage support. The solution utilizes Bootstrap 5’s native color modes (data-bs-theme) to ensure project-wide consistency for tables, cards, and forms.

Additionally, all list templates were refactored to remove hardcoded contrast classes, allowing the UI to adapt dynamically to the active theme without manual overrides.

Tests run with Java 21:
- mvn clean package

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is inspired by [Keep a Changelog](https://keepachangelog.com/en/1.1.0
 
 ### Added
 
+- Added Dark Theme support with a toggle switch in the navbar.
 - Added repository discovery metadata for search engines, LLMs, autonomous agents, and RAG systems.
 - Added `llms.txt` as a concise LLM entry point.
 - Added `llms-full.txt` as an expanded single-file project reference for larger context windows.
--- a/src/main/resources/static/js/theme-init.js
+++ b/src/main/resources/static/js/theme-init.js
@@ -1,0 +1,98 @@
+(function() {
+    const getStoredTheme = () => {
+        try {
+            return localStorage.getItem('theme');
+        } catch (e) {
+            return null;
+        }
+    };
+
+    const setStoredTheme = theme => {
+        try {
+            if (theme === 'auto') {
+                localStorage.removeItem('theme');
+            } else {
+                localStorage.setItem('theme', theme);
+            }
+        } catch (e) {}
+    };
+
+    const getPreferredTheme = () => {
+        const storedTheme = getStoredTheme();
+        if (storedTheme) {
+            return storedTheme;
+        }
+        return 'auto';
+    };
+
+    const setTheme = theme => {
+        let actualTheme = theme;
+        if (theme === 'auto') {
+            actualTheme = window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
+            document.documentElement.style.colorScheme = 'light dark';
+        } else {
+            document.documentElement.style.colorScheme = theme;
+        }
+        document.documentElement.setAttribute('data-bs-theme', actualTheme);
+        updateUI(theme, actualTheme);
+    };
+
+    const updateUI = (theme, actualTheme) => {
+        // Update icons and buttons if they exist
+        const themeIcon = document.getElementById('theme-icon');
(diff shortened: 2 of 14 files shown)


max-california (commented):
## 🤖 AI Code Review

1. **Blocking issues**

- **Navbar theming likely regresses readability/branding in both themes**
  - The navbar changed from `navbar-dark bg-dark` to `theme-aware-navbar`, but this diff does not add any CSS for `theme-aware-navbar`, nor any Bootstrap navbar color class such as `navbar-light`, `navbar-dark`, `bg-body-tertiary`, etc.
  - In Bootstrap, navbar link/icon colors depend on those classes/variables. Without them, text/icon contrast can become inconsistent or hard to read depending on defaults and the active theme.
  - **Fix:** use Bootstrap-supported theme-aware classes on the navbar, e.g. `navbar navbar-expand-lg bg-body-tertiary border-bottom shadow-sm mb-4`, and verify brand/nav/link/toggler colors in both light and dark modes. If `theme-aware-navbar` is intended to come from custom CSS elsewhere, that CSS should be included in this PR.

- **Theme toggle JS can throw and break on pages/fragments where the button is absent**
  - The script does `document.getElementById('bd-theme').addEventListener(...)` unconditionally. If this layout is ever reused in a context where that button is missing, this becomes a null dereference and stops subsequent JS.
  - Same risk for `updateThemeIcon()` if `theme-icon` is not present.
  - This may not happen today, but the layout is a shared template, so defensive code is warranted.
  - **Fix:** guard element lookups:
    ```js
    const themeToggle = document.getElementById('bd-theme');
    const themeIcon = document.getElementById('theme-icon');
    if (themeToggle) {
      themeToggle.addEventListener('click', ...);
    }
    function updateThemeIcon(theme) {
      if (!themeIcon) return;
      ...
    }
    ```

- **Support system preference changes when no explicit theme is stored**
  - The current implementation only applies system preference at initial page load. If the OS theme changes while the app is open, the UI won’t update unless the user reloads.
  - Consider listening to `matchMedia('(prefers-color-scheme: dark)')` changes and only applying them when no `localStorage` override exists.

- **Accessibility for the toggle button**
  - The toggle button has only a `title`. Consider adding an accessible label and state, e.g. `aria-label="Toggle color theme"` and update `aria-pressed` based on the active mode.

- **Avoid raw localStorage access assumptions**
  - In some privacy-restricted environments, `localStorage` access can throw. Rare, but easy to harden with `try/catch` around reads/writes.

- **Table header styling may be too subtle after removing `table-light`**
  - Removing `table-light` makes the theme adapt better, but it also removes explicit visual separation for headers. If readability suffers, consider replacing with a Bootstrap theme-aware table variant rather than leaving headers completely unstyled.

3. **Tests**

- Recommend UI/integration coverage for:
  - **Initial theme selection**: stored preference overrides system preference; otherwise system preference is used.
  - **Persistence**: clicking the toggle updates `data-bs-theme` and persists to `localStorage`.
  - **Theme icon state**: correct icon is shown on initial render and after toggling.
  - **Visual smoke test for navbar/table readability** in both light and dark modes, since this change intentionally removes hardcoded classes and alters navbar styling.

4. **Overall assessment**

Good direction and the Bootstrap `data-bs-theme` approach is sensible, but I’d fix the navbar styling regression risk and harden the toggle script before merge.

---

tmseidel (changes requested):
@PenceaV thanks for your contribution. I've looked over the code and have some suggestions before we can merge:
* The Login-Screen is still white, although my OS-preference is dark. The login screen should select the OS-selected theme, no button on this page
* In dark mode are several ui-glitches that breaks readibilty:
1. Selection of built-in tools (see screenshot)
image
2. The preview of the system-prompt of the bot-edit-page (see screenshot)
image

The issue "Theme toggle JS can throw and break on pages/fragments where the button is absent" from max-california should also be implemented.

tmseidel (approved):
Looks good to me 🚀

PR #151, opened 2026-05-22
feature/make-new-system-prompts-editable

### LINKED ISSUE #150: AI Created Issue: Make System-Prompts for E2E PR Workflow editable
Originates from #142

## Quality assessment

The issue intent is clear and feasible, and the added context resolves the main implementation ambiguity. The original issue was missing structure, migration expectations, and testable acceptance criteria.

## Summary
Make the three E2E PR workflow system prompts editable through the existing **System settings → System prompts** UI instead of keeping them hard-coded in `org.remus.giteabot.prworkflow.e2e.agents.E2ePromptLibrary`.

## Current behavior
The E2E PR workflow uses three hard-coded system prompts defined in `org.remus.giteabot.prworkflow.e2e.agents.E2ePromptLibrary`:
- planner prompt: plans the required E2E test case(s) based on the PR changes
- author prompt: turns the planner output into executable test cases
- runner prompt: executes the test cases

These prompts are not stored as editable `SystemPrompt` entries and therefore cannot be managed through the existing system prompt UI.

## Expected behavior
The three E2E workflow prompts should be handled like other system prompts:
- stored and managed through the existing `SystemPrompt` / `SystemPromptService` mechanism
- visible and editable in **System settings → System prompts**
- used by the E2E PR workflow at runtime instead of the hard-coded prompt text

## Scope
- Replace hard-coded E2E prompt usage with prompt lookup via the existing system prompt infrastructure.
- Add a Flyway migration to create/seed the required E2E prompt entries.
- Ensure the existing UI for system prompts allows these prompts to be viewed and edited.

## Implementation notes
Relevant existing components mentioned by the reporter:
- `org.remus.giteabot.systemsettings.SystemPromptService`
- `org.remus.giteabot.systemsettings.SystemPrompt`
- `src/main/resources/templates/system-settings/list.html`
- `org.remus.giteabot.prworkflow.e2e.agents.E2ePromptLibrary`

## Acceptance criteria
- [ ] The E2E planner, author, and runner prompts exist as system prompt records managed by the existing system prompt infrastructure.
- [ ] A Flyway migration creates the three E2E system prompt entries for existing installations.
- [ ] The prompts are visible in **System settings → System prompts**.
- [ ] A user can edit and save each of the three E2E prompts through the existing UI.
- [ ] The E2E PR workflow uses the stored prompt values at runtime.
- [ ] Existing installations continue to work after migration without requiring manual prompt creation.
- [ ] If a prompt record is missing, the application behavior is defined and handled gracefully (for example via seeded defaults or existing fallback behavior). If no fallback is desired, fail fast with a clear error.

## Testing notes
- Verify migration creates the three prompt entries on an existing database.
- Verify a fresh installation contains the three prompt entries.
- Verify editing each prompt in the UI persists the updated content.
- Verify the E2E workflow reads the edited prompt content rather than the previous hard-coded text.
- Verify behavior when one of the prompt records is absent or invalid.

## Out of scope
- Changing the semantics of the E2E planner/author/runner workflow itself.
- Redesigning the system prompt UI beyond what is needed to expose these prompts consistently.

## Target branch
`develop`

## Assumptions

- The existing `SystemPrompt` model and `SystemPromptService` are already the intended mechanism for storing prompt text.
- The current system prompt UI can display these prompts without requiring a new dedicated editor flow.
- The Flyway migration should seed default values equivalent or close to the current hard-coded E2E prompts.
- Using the existing system prompt infrastructure is the preferred approach rather than introducing a separate E2E-specific prompt store.

## Open questions

- Should missing E2E prompt records be automatically recreated from defaults at runtime, or should the application surface a configuration error?

---

tmseidel:


max-california (commented):
## 🤖 AI Code Review

### Diff chunk 1/2
1. **Blocking issues**

- **`pom.xml`: invalid project version string**
  - The version was changed to `1.8.0-SNASPSHOT`, which looks like a typo for `SNAPSHOT`.
  - This can break artifact publishing/version parsing and will leak an incorrect version into generated metadata.
  - **Fix:** change to `1.8.0-SNAPSHOT`.

- **Flyway migration version conflict risk (`V20__system_prompts_e2e.sql`)**
  - The repo already contains `src/main/resources/db/migration/*/V20__system_prompts_e2e.sql` in the provided tree, and this PR adds files with that same version/name as “new”. If this is accurate relative to target branch, Flyway will fail or the migration history is inconsistent.
  - I can’t fully verify branch state from the chunk alone, but this needs confirmation before merge.
  - **Fix:** ensure the migration version is unique relative to the target branch, and if `V20` already shipped, create a new migration instead of modifying/re-adding it.

- **`SystemPromptService`: removed public method may break compilation/runtime in the remaining diff**
  - `countBotsUsing(Long id)` was removed from `SystemPromptService`, but this chunk does not show corresponding call-site updates. Given the system settings UI lists prompt usage, this is a likely compile break or functional regression in the omitted half.
  - **Fix:** either keep the method or update all call sites in the PR. Please verify with a full build and controller/UI tests.

- **Theme script: `matchMedia(...).addEventListener` browser compatibility**
  - `MediaQueryList.addEventListener('change', ...)` is not supported in some older browsers/WebViews; `addListener` fallback is sometimes still needed if broad compatibility matters.
  - Probably fine for a modern admin UI, but worth deciding explicitly.

- **Theme script swallows storage errors silently**
  - The `catch (e) {}` blocks make failures impossible to diagnose.
  - **Suggestion:** at least leave a comment that storage errors are intentionally ignored, or optionally `console.debug` in development.

- **`E2ePromptLibrary.render(...)` silently substitutes empty framework when `framework == null`**
  - That hides programming errors and can generate malformed prompts.
  - **Suggestion:** consider `Objects.requireNonNull(framework, "framework")` instead. All current call sites appear to expect a real framework.

- **Docs/version metadata consistency**
  - `CITATION.cff` and `codemeta.json` say `1.6.0`, while `pom.xml` is now `1.8.0-...`. If this PR is not strictly a docs release, that may be intentional, but it’s easy for metadata to drift and confuse consumers.
  - **Suggestion:** align the intended release/versioning story across these files.

3. **Tests**

- **Missing tests for editable E2E prompt fallback behavior**
  - Add unit tests for:
    - default prompt used when `SystemPrompt` is `null`
    - default prompt used when E2E prompt fields are blank
    - protocol suffix is always appended
    - framework placeholder substitution works correctly

- **Missing migration/entity persistence coverage**
  - This change adds three new non-null DB columns and new entity fields. Recommend an integration test proving an existing prompt row migrates correctly and a `SystemPrompt` can be saved/loaded with the new fields.

- **Missing controller/form tests for new system prompt fields**
  - Since the UI now edits/clones/previews the three E2E fields, add/update MVC tests covering create/edit/clone/preview flows.

- **Missing workflow-level regression test**
  - Recommend at least one test around `PlaywrightTestSuiteRunner` proving the bot’s `SystemPrompt` is passed through to planner/author/runner and that blank fields don’t break execution.

4. **Overall assessment**

Good direction overall: splitting editable role text from non-editable protocol is a solid design for preserving pipeline correctness while allowing customization. However, I would **not merge yet** because of the `pom.xml` version typo, the likely Flyway migration/versioning concern, and the need to confirm the removed `countBotsUsing` method doesn’t break the rest of the PR.

### Diff chunk 2/2
1. **Blocking issues**

- **Potential backward-compatibility break in `SystemPromptService` validation**
  
  The new tests assert that `save()` now requires `e2ePlannerSystemPrompt`, `e2eAuthorSystemPrompt`, and `e2eRunnerSystemPrompt`. If existing `system_prompt` rows created before this feature can still exist without those fields populated, this will break editing/saving older records unless a migration backfills them first. I can’t see the migration/entity/controller changes in this chunk, so this needs verification before merge.

  **Why it matters:** a user opening and saving an older prompt entry could start getting validation errors unexpectedly, or startup/bootstrap paths that persist prompts could fail.

  **Suggested fix:** ensure there is a DB migration that backfills non-null/default values for all existing rows, and add an integration test covering upgrade/edit of a legacy `SystemPrompt` record.

- **Exercise the new editable prompt path for author/runner too**
  
  `TestPlannerAgentTest` now verifies that the edited system prompt is prepended and the non-editable suffix remains appended. That’s good, but the same risk exists for `TestAuthorAgent` and `TestRunnerAgent`, which also gained a new `SystemPrompt` parameter in their public methods. A similar assertion there would help catch regressions where one agent accidentally ignores the editable DB prompt or fails to append its protocol suffix.

- **Avoid relying on brittle built-in prompt text**
  
  In `fallsBackToBuiltInPromptWhenSystemPromptSlotIsBlank()`, asserting `.contains("TestPlannerAgent")` is a bit implementation-specific. If the bundled prompt text changes, this test may fail without a behavior regression. Prefer asserting something stable about fallback behavior, e.g. that the custom text is absent and the required protocol suffix is still present.

3. **Tests**

- **Missing upgrade/backfill coverage for legacy prompts**
  
  Given the stronger validation, add a test that simulates a legacy `SystemPrompt` record missing the new E2E fields and verifies either:
  - migration/backfill populates them, or
  - the service/controller can still load and update safely after upgrade.

- **Missing editable prompt tests for author/runner agents**
  
  Add tests parallel to `usesEditedSystemPromptFromSystemPromptEntity()` for:
  - `TestAuthorAgent.write(..., systemPrompt)`
  - `TestRunnerAgent.execute(..., systemPrompt)`
  
  Specifically verify:
  - edited prompt text is included,
  - runtime-owned/non-editable protocol instructions are still appended,
  - blank custom fields fall back to built-in defaults.

- **Good addition:** the new `GitLabApiClient.getIssueComments()` test is useful and should help guard the API path/query encoding.

4. **Overall assessment**

Looks mostly good from this chunk, and the added tests are moving in the right direction. The main thing I’d want confirmed before merge is safe handling of pre-existing `SystemPrompt` data now that the three E2E prompt fields appear to be mandatory.

---

PR #153, opened 2026-05-23
feature/configure-e2e-pr-workflow

### LINKED ISSUE #147: Configure E2E PR-workflow for this project

**Expected behavior**

tmseidel:


max-california (commented):
## 🤖 AI Code Review

1. **Blocking issues**

- **`preview.yml` is using PR-only context in a `workflow_dispatch` workflow, so the deployed revision/slot will be wrong in normal dispatches.**  
  In this workflow, `github.event.pull_request.head.sha` and `github.event.pull_request.number` are referenced multiple times, but `workflow_dispatch` events do not include a `pull_request` payload. That means:
  - checkout falls back to `github.sha`, which for a workflow-dispatch run is the workflow ref/default branch commit, not the PR head commit;
  - `PR_NUMBER` will be empty, so slot/remote dir/port become `run-${RUN_NUMBER}` instead of PR-specific;
  - preview URLs won’t be stable per PR.

  Given this PR is specifically for PR preview/e2e workflows, this looks like a functional mismatch unless the bot dispatches extra inputs and the workflow consumes them.  
  **Suggested fix:** pass the PR number / head SHA explicitly as workflow inputs (e.g. `prNumber`, `refSha`, maybe `branch`) and use those instead of `github.event.pull_request.*`. If the bot already sends them, update the workflow to read them.

- **`DEPLOY_HOST` is set from `vars.DEPLOY_USER`, which looks like a likely copy/paste bug and will break SSH/scp target resolution.**  
  The workflow currently has:
  ```yaml
  DEPLOY_HOST: ${{ vars.DEPLOY_USER }}
  DEPLOY_USER: ${{ vars.DEPLOY_USER }}
  ```
  Unless the repository variable is intentionally overloaded, this will attempt to deploy to a host equal to the username.  
  **Suggested fix:** introduce/use a separate `vars.DEPLOY_HOST` and keep `DEPLOY_USER` mapped to `vars.DEPLOY_USER`.

- **The callback step ignores the resolved fallback callback URL and only uses the direct input, causing silent callback loss.**  
  `Resolve deployment metadata` computes `CALLBACK_URL` from either `inputs.callbackUrl` or `vars.CALLBACK_URL`, and exports it as `steps.meta.outputs.callback_url`. But `Notify bot` uses:
  ```yaml
  CALLBACK_URL: ${{ inputs.callbackUrl }}
  ```
  So when the input is empty and fallback was intended, the notify step skips entirely. This is inconsistent with the earlier step and likely breaks the documented fallback behavior.  
  **Suggested fix:** use `CALLBACK_URL: ${{ steps.meta.outputs.callback_url }}` in the notify step.

- **SSH host key verification is disabled, which is not safe for a deployment workflow handling a private key and shipping artifacts.**  
  `StrictHostKeyChecking=no` plus `UserKnownHostsFile=/dev/null` makes MITM of the deploy target trivial. For anything beyond throwaway local testing, this is not acceptable.  
  **Suggested fix:** pin the server host key via `known_hosts` (repo/org secret or checked-in fingerprint for non-sensitive environments) and enable normal host key verification.

- **`tail -c1` newline check may fail under `set -e` for empty files.**  
  In practice `SSH_PRIVATE_KEY` is checked earlier, so this may be fine, but the current idiom is brittle. A safer pattern is:
  ```sh
  last_char="$(tail -c 1 "$SSH_KEY_FILE" || true)"
  if [ -n "$last_char" ]; then ...
  fi
  ```

- **The “repository already present in workspace” short-circuit can reuse stale contents.**  
  If the workspace persists across runs, this exits before verifying that the current checkout matches the requested SHA. That can produce surprising deployments.  
  **Suggested fix:** either always fetch/reset, or validate `HEAD` equals `REF_SHA` before reusing the workspace.

- **Port allocation by `prNumber % 1000` risks collisions once PR numbers exceed 1000 active previews.**  
  Maybe acceptable for the intended environment, but worth documenting as a limitation or adding collision detection on the target host.

- **`guard-main.yml` event type narrowing was removed.**  
  This is probably fine because the default `pull_request` activity types include the common cases, but if you intentionally wanted coverage for `ready_for_review` or `edited`, verify they’re still included as desired.

3. **Tests**

- For the new preview workflow logic, I’d recommend validating at least:
  - dispatch runs with no `pull_request` payload still deploy the intended PR revision/metadata;
  - callback URL fallback works end-to-end;
  - deployment host/user env wiring is correct.

  Since this is GitHub Actions YAML, that may be docs/systemtest coverage rather than unit tests, but right now these behaviors are unverified and central to the change.

4. **Overall assessment**

Good direction, and the Docker/public-url docs updates look useful. However, the new `preview.yml` has a few functional and security issues that should be fixed before merge, especially the use of nonexistent PR context in `workflow_dispatch`, the apparent `DEPLOY_HOST` misconfiguration, and the callback fallback bug.

---

tmseidel:
@max-california The worfklow is intended only to run on pull-requests

max-california:
## 🤖 Bot Response

1. **Blocking issues**

- **The workflow is still configured as `workflow_dispatch`, so “PR-only” intent does not match the actual event payload.**  
  Even if the business intent is “only for pull requests,” GitHub Actions will not populate `github.event.pull_request.*` for a `workflow_dispatch` run. In the current YAML, those fields are used for checkout, PR number, slot naming, and URLs. Unless another workflow dispatches this one and passes equivalent values as inputs *and this workflow reads those inputs instead*, the current implementation is relying on context that won’t exist.

  Concretely, these references are unsafe in a `workflow_dispatch` workflow:
  - `github.event.pull_request.head.sha`
  - `github.event.pull_request.number`

  **Why this matters:** the workflow may deploy the wrong commit, generate unstable/non-PR-specific preview names, or silently fall back to run-number behavior.

  **Fix options:**
  - If this workflow should run directly on PRs, change it to `on: pull_request` and keep using PR context.
  - If it should remain manually/indirectly dispatched, pass `prNumber` and `refSha` as inputs from the caller and replace the `github.event.pull_request.*` usages with those inputs.

- **`DEPLOY_HOST` still appears incorrectly wired to `vars.DEPLOY_USER`.**  
  If this wasn’t intentional, it will break deployment targeting regardless of PR-only scope.  
  **Fix:** map `DEPLOY_HOST` to a host-specific variable, e.g. `vars.DEPLOY_HOST`.

- **SSH host key checking is disabled.**  
  This is independent of trigger semantics and is still a real security issue for a deployment workflow.  
  **Fix:** provide a trusted `known_hosts` entry and enable verification.

- If this is intended to be a PR-only workflow but invoked indirectly, make that contract explicit in the workflow interface:
  - require `prNumber` and `refSha` inputs,
  - validate they are present early,
  - avoid fallback logic that silently turns a PR preview into a generic run preview.

- The callback step should still use the resolved metadata output (`steps.meta.outputs.callback_url`) rather than raw `inputs.callbackUrl`, otherwise the fallback behavior remains inconsistent.

3. **Tests**

- Please add or verify coverage for the actual invocation path:
  - if triggered by `pull_request`, test that PR SHA/number resolution works;
  - if triggered by another workflow via dispatch, test that required PR metadata is passed and consumed correctly.

4. **Overall assessment**

If the intention is truly “only on pull requests,” the current trigger/configuration still needs to reflect that. As written, the workflow’s event type and the context it consumes are inconsistent, so I’d still treat that as a pre-merge issue.

---
*Response by AI Git Bot*

PR #162, opened 2026-05-24
feat: add auto-login, expandable settings, fix e2e workflow issues

### LINKED ISSUE #147: Configure E2E PR-workflow for this project

tmseidel:
FEATURES:
- Add auto-login mode for e2e/preview environments
  - New AutoLoginConfig with automatic admin account creation
  - Pre-authenticates all requests, skips login form
  - Enabled in preview.yml for testing
  - Property: giteabot.security.auto-login.enabled (default: false)

- Make system-settings sections expandable/collapsable
  - Each section (System prompts, MCP, Tool, Workflow, Deployment) can be individually toggled via card header click
  - Smooth CSS transitions with chevron icon rotation
  - Collapse state persisted in localStorage
  - Add "Expand all" / "Collapse all" buttons
  - Accessibility: proper aria-expanded / aria-controls attributes

FIXES:
- Fix POSIX sh compatibility in preview.yml
  - Replace bash-specific constructs (local, (( )), substring expansion)
  - Use POSIX-compatible alternatives (cut, [ ], $(()))
  - Allows script to run under 'sh -e' on containers

- Fix E2E slash commands on GitHub (@bot rerun-tests / regenerate-tests)
  - Fallback to payload.issue.number for GitHub issue_comment events
  - Only accept when issue.pullRequest is non-null (actual PR, not plain issue)
  - Unblocks slash commands on GitHub; Gitea/GitLab already working

SECURITY:
- Auto-login disabled by default, clearly marked for non-production use
- Prominent WARN log on startup when enabled


max-california (commented):
## 🤖 AI Code Review

1. **Blocking issues**

- **Auto-login currently applies to all non-API UI routes, including setup/login flows, with no path guard.**
  `AutoLoginFilter` authenticates every unauthenticated request unconditionally. That means `/login` and `/setup` become effectively inaccessible as unauthenticated pages whenever the flag is enabled, and logout will immediately re-authenticate on the next request despite the class comment claiming `/logout` is skipped. This is at least inconsistent with the documented behavior, and may break expected flows/tests that rely on seeing the login page or setup page in preview/demo.
  - **Why it matters:** the implementation does not match its own contract, and the logout comment is misleading.
  - **Suggested fix:** implement `shouldNotFilter(...)` or explicit path checks to skip `/login`, `/logout`, `/setup/**`, and static assets if that is the intended behavior; otherwise update the docs/comments to reflect the real behavior.

- **`E2ETestWorkflow.resolvePrNumber(...)` is less strict than the orchestrator fallback and can misclassify plain issue comments as PRs.**
  In `PrWorkflowOrchestrator`, the fallback from `issue.number` is guarded by `issue.pullRequest != null`, which is the right GitHub distinction between issue comments on issues vs PRs. In `E2ETestWorkflow`, the local `resolvePrNumber(...)` uses `issue.number` without checking `issue.pullRequest`.
  - **Why it matters:** this reintroduces ambiguity the orchestrator just fixed. If this workflow is ever invoked with an issue-comment payload that is not actually a PR comment, it can run against a bogus PR number `issue.number`, then later fail in less obvious ways.
  - **Suggested fix:** keep the fallback logic identical in both places. Prefer extracting a shared helper to avoid drift.

- **Avoid persisting the security context manually inside the auto-login filter unless necessary.**
  `CONTEXT_REPOSITORY.saveContext(...)` on every request is more invasive than just populating the request-local context. If the goal is “every request is authenticated while flag is on,” session persistence is not strictly needed and can create odd interactions around logout/session semantics.
  - If you do want session persistence, consider documenting that logout is intentionally non-sticky in auto-login mode.
  - If not, removing the explicit save would simplify behavior.

- **The deterministic preview encryption key in `preview.yml` is okay for ephemeral previews, but it’s worth constraining visibility further.**
  The PR already warns against production use, which is good. Still, since this is a hardcoded known key plus admin/admin credentials, I’d recommend ensuring the preview job is never reusable outside preview/test contexts and that docs clearly call out the exposure model.
  - Not a blocker because this is explicitly preview-only, but worth double-checking operationally.

- **`PlaywrightTestSuiteRunner.runExistingTests(...)` saves cloned cases one-by-one.**
  This is probably fine for current suite sizes, but if reruns grow toward the configured caps, batching with `saveAll(...)` would be cleaner and cheaper.

- **`effectiveRetries(...)` formatting looks suspicious in the diff.**
  The diff shows:
  ```java
  private static int effectiveRetries(TestPlan plan, int workflowDefault) {        Integer fromPlan = plan.maxRetries();
  ```
  This may just be a rendering artifact, but please verify the actual file formatting/compilation around that method.

3. **Tests**

- **Missing tests for auto-login security behavior.**
  Add MVC/security tests covering:
  - auto-login disabled: normal login required
  - auto-login enabled: authenticated access to protected UI routes
  - behavior for `/login`, `/logout`, and `/setup` specifically
  - startup bootstrap creates admin user only when absent

- **Missing tests for GitHub issue-vs-PR distinction in the workflow layer.**
  There should be a test proving `E2ETestWorkflow` does **not** treat a plain GitHub issue comment payload as a PR run target.

- **Missing rerun-only runner tests.**
  Good to see slash-command tests updated, but I’d strongly recommend dedicated tests for `PlaywrightTestSuiteRunner` rerun mode:
  - restores files from previous suite
  - skips planner/author/runner agents
  - clones cases into new suite
  - returns skipped/error when previous suite has no cases or no files can be restored

- **Missing hydration tests for GitHub slash commands.**
  Add coverage that `E2eTestSlashCommandHandler` hydrates PR details from `getPullRequestDetails(...)` when handling GitHub `issue_comment` payloads lacking `pull_request.head`.

4. **Overall assessment**

Good direction overall: the GitHub slash-command fix and rerun-only path address real workflow gaps, and the preview deployment hardening is useful. But I’d fix the auto-login behavior mismatch and align PR-number resolution logic before merge.

---

PR #169, opened 2026-05-24
Feature/configure e2e pr workflow

### LINKED ISSUE #147: Configure E2E PR-workflow for this project

tmseidel:
Additional features for stabilizing the e2e workflow and make it more usable

### DIFF
--- a/doc/PR_WORKFLOWS_E2E.md
+++ b/doc/PR_WORKFLOWS_E2E.md
@@ -58,6 +58,7 @@ via `PrWorkflowContext.hints`.
    | `maxRetries`     | integer | `1`          | Per-test retry budget. A test that passes after retry is tagged `FLAKY`. Capped at 5. |
    | `maxTestCases`   | integer | `20`         | Hard cost guard. Capped at 100 server-side. |
    | `suiteLifecycle` | string  | `ephemeral`  | See [Suite lifecycle modes](#suite-lifecycle-modes) below. |
+   | `promotionThresholdPercent` | integer | `100` | Minimum percentage of executed cases that must pass for the suite to be **promoted** (offer-as-pr / commit-to-pr / promote-on-merge). `100` keeps the original "only fully green suites are promoted" behaviour; lower it (e.g. `80`) when you want partial successes to still trigger the configured promotion action. `ERROR` / `SKIPPED` suites are never promoted regardless of this value. See [Treating generated tests as a regression baseline](#treating-generated-tests-as-a-regression-baseline). |
 
 3. Save. The next PR-open / PR-synchronise webhook triggers the workflow.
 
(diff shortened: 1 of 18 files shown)


max-california (commented):
## 🤖 AI Code Review

1. **Blocking issues**

- **Whitelist enforcement appears incomplete for AI-spending entry points.**  
  The PR introduces `BotService.isUserAllowed(...)` and `BotWebhookService.isCallerAllowed(...)`, but in the shown diff it is only applied in `handleIssueComment(...)`. The docs/UI say the whitelist limits *PR opens/syncs, @-mentions in PR or issue comments, inline review comments, review submissions, agent issue assignments and follow-up comments*. However:
  - `reviewPullRequest(...)` has no whitelist check
  - `handleBotCommand(...)` has no whitelist check
  - `handlePrComment(...)` has no whitelist check on the non-agent path
  - `handleInlineComment(...)` has no whitelist check
  - `handleReviewSubmitted(...)` has no whitelist check
  - `handleIssueAssigned(...)` is mentioned in docs but not shown guarded here

  If the rest of the file truly lacks those checks, the feature does not enforce the stated security boundary and public-repo token-spend protection will be bypassable through several webhook types.  
  **Fix:** apply `isCallerAllowed(bot, payload)` consistently at every AI-triggering webhook entry point, before any expensive work is started.

- **`PROMOTE_ON_MERGE` eligibility now diverges from immediate-run promotion semantics.**  
  In `E2ETestWorkflow.run(...)`, promotion uses `TestSuiteOutcome` and explicitly never promotes `ERROR`/`SKIPPED`. In `E2eTestPrCloseHandler.isEligibleForPromotion(...)`, eligibility is inferred from `PrWorkflowRunStatus` plus per-case statuses. A `FAILED` workflow run can now promote if case pass-rate meets threshold, but this method does not appear to check the suite’s own terminal outcome/status, only test-case rows. That risks promoting suites from runs that failed for non-test reasons after/beside case execution.  
  The comment says `ERROR`/`SKIPPED` suites are never promoted, but I don’t see that enforced here unless `PrWorkflowRunStatus` maps 1:1 to those outcome types and no ambiguous `FAILED` states exist.  
  **Fix:** gate close-handler promotion on the persisted suite outcome/status (if available), or persist enough outcome metadata to distinguish assertion failures from infra/runtime errors and mirror `meetsPromotionThreshold(...)` exactly.

- **Avoid swallowing workflow-enabled lookup failures as “disabled”.**  
  `isWorkflowEnabled(...)` catches `RuntimeException` and returns `false`. That’s safe, but it can silently turn transient DB/config errors into behavior changes (e.g. posting “did not understand” instead of handling a valid review command). At least `warn` level logging may be more appropriate here, or differentiate “missing config” from actual lookup failure.

- **`isCallerAllowed(...)` does duplicate parsing work.**  
  It calls `botService.getAllowedUsernames(bot)` and then `botService.isUserAllowed(bot, caller)`, which reparses the whitelist. Not a big deal, but easy to simplify by checking membership against the already computed set.

- **Consider using `Locale.ROOT` for lowercasing usernames.**  
  `toLowerCase()` without an explicit locale can behave unexpectedly in some locales. For identifiers/usernames, `toLowerCase(Locale.ROOT)` is safer.

- **Unrecognized-command reply may be noisy for typoed mentions in non-review bots.**  
  The new fallback is useful, but verify this won’t create bot chatter for any mention that happens to include `@bot` in casual discussion. If that’s a concern, consider only replying when the comment syntactically looks like a command directed at the bot.

3. **Tests**

- Good additions around singular slash-command variants and promotion-threshold boundaries.

- **Missing tests for whitelist enforcement across all documented entry points.**  
  Given this is a security/cost-control feature, I’d strongly recommend explicit tests that blocked users cannot trigger:
  - PR open/sync workflow execution
  - PR mention command handling
  - inline review comment handling
  - review submission handling
  - issue assignment / issue comment paths
  - allowed users still can

- **Add tests for `E2eTestPrCloseHandler` non-promotion on infra-style outcomes.**  
  Especially if there is a persisted suite/run status distinction, add coverage proving `ERROR`/`SKIPPED` are not promoted even with threshold `0`.

- **Add unit tests for `BotService` whitelist parsing/matching.**  
  Cases: blank/null, commas + newlines mixed, trimming, case-insensitivity, duplicate entries, blank username with non-empty whitelist.

4. **Overall assessment**

Solid direction overall: the compile-time-safe workflow param keys and slash-command tolerance are good improvements, and the new promotion-threshold tests are useful. But I’d hold merge until the whitelist is enforced consistently across all documented AI-spending webhook paths, and until the `PROMOTE_ON_MERGE` logic is verified to match the documented “never promote ERROR/SKIPPED” semantics.

---

PR #177, opened 2026-05-24
Fix: Create unique branch name per test suite promotion (rerun collision fix)

tmseidel:


max-california (commented):
## 🤖 AI Code Review

1. **Blocking issues**

- **Fallback branch naming with `System.currentTimeMillis()` is still collision-prone.**  
  If `run.getId()` is null, the code falls back to `"t" + System.currentTimeMillis()`. Two promotions started within the same millisecond can still generate the same branch name, especially under parallel execution or retries in test environments. That means the “unique branch per promotion” guarantee is not fully met for unsaved/new runs.  
  **Why it matters:** this PR is explicitly fixing rerun collisions; leaving a time-based fallback weakens the fix in a realistic edge case.  
  **Suggested fix:** use a stronger unique suffix, e.g. `UUID.randomUUID()` (possibly shortened), or persist the run before promotion so an ID is always available.

- **Consider extracting branch-name generation into a dedicated helper and testing it directly.**  
  Right now the uniqueness logic is embedded in `promote()`. A small helper like `buildWorkBranch(mode, prNumber, run)` would make this behavior easier to reason about and easier to test, especially around null `run.id`.

- **Verify branch-name compatibility/length against supported providers.**  
  The new format is still reasonable, but if `prNumber`/suffix formats grow or provider-specific constraints exist, it may be worth centralizing/sanitizing branch naming. Not a problem in this diff by itself, just something to keep in mind.

3. **Tests**

- **Add coverage for `run.getId() == null`.**  
  That’s the most important missing case introduced by this change. At minimum, add a test asserting a non-empty branch suffix is used when the run has no ID. Ideally also avoid asserting exact timestamp content, since that’s brittle.

- **If you keep the timestamp fallback, add a test documenting its behavior.**  
  Even if not fully collision-safe, it should at least be explicit and covered.

4. **Overall assessment**

Good targeted fix and the updated tests cover the main rerun scenario well. The only thing I’d want addressed before merge is the null-`runId` fallback, since the current timestamp-based suffix can still collide and undermines the uniqueness guarantee.

---

PR #191, opened 2026-05-30
feat: implement unit-test-author PR workflow

### LINKED ISSUE #187: UI: Recommend native-tool usage for agent instead of legacy tooling
**Current behavior**
When editing an AI-Integration the usage of native tool calls are marked as experimental. When using frontier models from Anthropic, Gemini, OpenAI or DeepSeek the native tooling is working like expected and is more reliable than the legacy tooling.

**Expected behavior**
The native tool calling should be enabled by default. The "experimental" flag should be removed with the hint that if using frontier models it is the recommended route. Only if it don't work in agentic worklfows a fallback to the legacy json tooling should be considered.

### LINKED ISSUE #190: New PR Workflow:  unit-test-author
Category: TESTING · Persona: developer / tech lead · Deployment target? no

Generates / extends unit tests for files touched by the diff. Different from the shipped e2e-test workflow: no preview env, no browser — just pr-unit-test-write + the project's own mvn test / npm test / pytest / go test runner.

Triggers: PR opened/synchronized, @bot write-unit-tests <path>.
Outputs: New test files in the PR's branch (via the existing coding-agent infra), inline coverage delta in the PR comment

Implementation nodes:
* The implementation should follow the agentic pr workflows e2e (package `org/remus/giteabot/prworkflow/e2e` and agentic-review (package `org/remus/giteabot/prworkflow/agentreview`)
* A separate system prompt can be edited in the system-settings beside the other system-prompts
* workflow parameter similar to the e2e test that can be edited on the workflow-edit page:
  * Max retries per test
  * Max test cases per suite
  * Suite lifecycle
  * Promotion pass-rate threshold (%)
* The agent should create it's testcases in a separate package next to exisiting tests
* The agent should call afterwards the execution of the generated tests
* Update the documentation in the same way the other workflows are documented.

tmseidel:


max-california:
## 🤖 Agentic Code Review

## Summary

Nice feature addition overall, but I found **two blocking correctness/security issues** in the implementation, plus a couple of follow-up maintainability/documentation mismatches.

## Blocking issues

### 1. The write guard still allows production-code modifications for some frameworks
**Files:**  
- `src/main/java/org/remus/giteabot/prworkflow/unittest/UnitTestFramework.java:25,34,37`  
- `src/main/java/org/remus/giteabot/prworkflow/unittest/tools/UnitTestToolExecutor.java:119-131`

The docs promise “production code is never touched”, but the actual allowed prefixes contradict that:

- `NPM` allows writes under `src/`
- `CARGO` allows writes under `src/`
- `DOTNET` allows writes under `Tests/`, `tests/`, `test/` only, which is fine, but case sensitivity depends on repo/platform
- the executor checks only `startsWith(prefix)`, so anything under those prefixes is accepted as a “test”

That means the agent can write files like:

- `src/index.js` in an npm repo
- `src/lib.rs` in a cargo repo

Those are production paths, not test-only paths.

This is a real safety bug, not just a doc mismatch. If the workflow pushes directly to the PR branch, it can modify application code despite the stated guarantees.

**Suggested fix:** tighten `allowedTestPrefixes()` and/or add framework-specific filename rules in `UnitTestToolExecutor`, e.g.:

- npm: `test/`, `tests/`, `__tests__/`, maybe `src/**/__tests__/` but not bare `src/`
- cargo: allow `tests/` only unless you explicitly want inline unit tests, in which case the workflow’s “production code is never touched” claim must be removed
- if inline tests are intended for some ecosystems, that needs a separate, explicit policy and commit-time validation

Also, the comment in `UnitTestToolExecutor` mentions a workflow-level assertion that nothing outside test dirs changed, but I don’t see such an assertion before `commitAndPush()`.

---

### 2. Slash-command help and command handling are inconsistent, so the advertised command won’t work end-to-end
**Files:**  
- `README.md` / `doc/PR_WORKFLOWS_UNIT_TEST.md` advertise `@bot generate-tests`
- `src/main/java/org/remus/giteabot/prworkflow/unittest/UnitTestSlashCommandHandler.java:18-19,33`
- `src/main/java/org/remus/giteabot/admin/BotWebhookService.java:447-462`
- `src/main/java/org/remus/giteabot/admin/BotWebhookService.java` imports/fields only mention `E2eTestSlashCommandHandler`

The new workflow introduces `UnitTestSlashCommandHandler`, but `BotWebhookService` still only references the E2E slash-command handler, and the fallback “unknown command” help text only lists E2E commands.

So as it stands, one of two things is true:

- either the new handler is **not wired into command dispatch at all**, in which case `@bot generate-tests` never triggers the workflow
- or it is wired elsewhere, but the bot still replies with misleading “I did not understand that command” help and omits the new commands from operator guidance

Either way, the user-facing contract added in the docs is not implemented consistently.

**Suggested fix:** wire `UnitTestSlashCommandHandler` into the same dispatch chain as E2E, and update the unrecognized-command/help response to include:
- `@bot generate-tests`
- `@bot rerun-unit-tests`

Also note the class Javadoc says it handles `@bot rerun-tests`, but the regex only accepts `generate-tests?` and `rerun-unit-tests?`.

## Non-blocking findings

### 3. Framework auto-detection is shallow and will miss common mono-repo / nested-project layouts
**File:** `src/main/java/org/remus/giteabot/prworkflow/unittest/FrameworkDetector.java:33-69`

Detection only checks top-level marker files. That means repos with:
- `backend/pom.xml`
- `services/api/go.mod`
- `frontend/package.json`
- solution/project files in nested folders

will be reported as unsupported and skipped, even though the repo clearly contains a supported toolchain.

If this is intentional MVP behavior, document it clearly. Otherwise consider recursive detection with a priority strategy.

### 4. Result mapping treats failed generated tests as workflow success
**File:** `src/main/java/org/remus/giteabot/prworkflow/unittest/UnitTestService.java:190-198`

When the runner returns `FAILED`, the workflow returns `Result.success(...)`. That may be deliberate (“workflow succeeded in generating tests, but tests failed”), but it’s surprising for a testing workflow and can affect orchestration/metrics/UI semantics.

If intentional, I’d suggest documenting that distinction clearly; otherwise map `FAILED` to workflow failure.

## Overall

Good structure and documentation, but I’d block this until the **write-safety hole** and **slash-command integration inconsistency** are fixed. Those both affect the core contract of the feature.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

This looks much better than the earlier version: the biggest safety/documentation gaps around test-file writes and slash-command discoverability appear to have been addressed. I don’t see a clear blocking issue in the diff shown.

## Findings

### Minor: duplicated documentation changes in the PR description
The unified diff repeats some hunks/files multiple times (`doc/PR_WORKFLOWS.md`, `doc/agentic-workflows/README.md`, `ToolCatalog.java`, etc.). I assume that’s an artifact of how the diff was pasted here rather than the actual branch state, but if those duplicates are really present in the PR generation pipeline, it makes review harder and increases the chance of merge/conflict mistakes.

**Suggestion:** verify the actual PR diff is deduplicated and only contains each hunk once.

### Minor: terminology drift around `use_legacy_tool_calling`
**Files:**  
- `doc/AGENT.md`  
- `doc/TOOL_CALLING.md`  
- `src/main/java/org/remus/giteabot/admin/AiIntegration.java`

The docs and entity comments are now aligned on the new default (`use_legacy_tool_calling = false`), which is good. One thing to keep an eye on is operator wording consistency:

- docs say “Enable native tool calling”
- recovery steps say “turn it off” to get legacy fallback
- field name is still negative (`useLegacyToolCalling`)

That’s all logically correct, but it’s easy for future maintainers to regress because the persistence flag and UI label are inverses.

**Suggestion:** keep the transient accessor/API and admin-form naming covered by tests, so future refactors don’t silently flip the semantics again.

## Overall

No blocking concerns from what’s shown here. The added pre-commit guard plus `UnitTestPathGuard` language closes the most important safety hole, and the command routing/help text now matches the advertised workflow behavior.

---

PR #193, opened 2026-05-31
feat: replaced default connection factory of RestClient to the http5 …

### LINKED ISSUE #100: Error 400 with vLLM via OpenAI integration
**Current behavior**
When using vLLM via the OpenAI integration the following error occurs:

http 400 error:
`validation error: {'type': 'missing', 'loc': ('body',), 'msg': 'Field required', 'input': None}`

I solved it adding a nginx sidecar to deployment yaml using the recommendation of AI below (proxy_http_version and proxy_request_buffering fix). But it's just a workaround.

**Expected behavior**
The vLLM integration works out of the box

**Possible solutions 1**
The provider builds its `RestClient` here, with no custom request factory:

```47:56:src/main/java/org/remus/giteabot/ai/openai/OpenAiProviderMetadata.java
    public RestClient buildRestClient(AiIntegration integration, String decryptedApiKey) {
        if (decryptedApiKey == null || decryptedApiKey.isBlank()) {
            throw new IllegalStateException("OpenAI integration requires an API key");
        }
        return RestClient.builder()
                .baseUrl(integration.getApiUrl())
                .defaultHeader("Authorization", "Bearer " + decryptedApiKey)
                .defaultHeader("Content-Type", "application/json")
                .build();
    }
```

A change there pinning the JDK client to HTTP/1.1 usually fixes this class of issue. Conceptually:

```java
java.net.http.HttpClient jdk = java.net.http.HttpClient.newBuilder()
        .version(java.net.http.HttpClient.Version.HTTP_1_1)
        .build();

return RestClient.builder()
        .baseUrl(integration.getApiUrl())
        .requestFactory(new JdkClientHttpRequestFactory(jdk))
        .defaultHeader("Authorization", "Bearer " + decryptedApiKey)
        .defaultHeader("Content-Type", "application/json")
        .build();
```

Caveats:
- This affects **all** OpenAI integrations, including real OpenAI. Real OpenAI is happy with HTTP/1.1, so practically harmless, but it's a global behavior change.
- Better placed on a dedicated `VllmProviderMetadata` (Path B from my last message) so it only impacts vLLM connections.

If HTTP/1.1 alone doesn't help, swap `JdkClientHttpRequestFactory` for **Apache HttpClient 5** — it's the most battle-tested HTTP client for Java↔Python servers and is the easiest workaround for the chunked-encoding edge case. That requires adding `org.apache.httpcomponents.client5:httpclient5` to `pom.xml` and using `HttpComponentsClientHttpRequestFactory`.

**Possible solution 2**
Add the workaround by adding an nginx to the deployment, like described in https://github.com/tmseidel/ai-git-bot/issues/14#issuecomment-4386283297

tmseidel:
…connectionfactory from httpclient for OpenAiProviderMetadata.java to fix several 400er error that come up when using several openai based llm interfaces

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,6 +81,11 @@
             <groupId>org.thymeleaf.extras</groupId>
             <artifactId>thymeleaf-extras-springsecurity6</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents.client5</groupId>
+            <artifactId>httpclient5</artifactId>
+        </dependency>
+
 
         <!-- Model Context Protocol client SDK -->
         <dependency>
--- a/src/main/java/org/remus/giteabot/ai/openai/OpenAiProviderMetadata.java
+++ b/src/main/java/org/remus/giteabot/ai/openai/OpenAiProviderMetadata.java
@@ -1,8 +1,10 @@
 package org.remus.giteabot.ai.openai;
 
+import org.apache.hc.client5.http.impl.classic.HttpClients;
 import org.remus.giteabot.admin.AiIntegration;
 import org.remus.giteabot.ai.AiClient;
 import org.remus.giteabot.ai.AiProviderMetadata;
+import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
 import org.springframework.stereotype.Component;
 import org.springframework.web.client.RestClient;
 
@@ -48,8 +50,12 @@ public class OpenAiProviderMetadata implements AiProviderMetadata {
         if (decryptedApiKey == null || decryptedApiKey.isBlank()) {
             throw new IllegalStateException("OpenAI integration requires an API key");
         }
+        var httpClient = HttpClients.createDefault();
+        HttpComponentsClientHttpRequestFactory factory =
+                new HttpComponentsClientHttpRequestFactory(httpClient);
         return RestClient.builder()
                 .baseUrl(integration.getApiUrl())
+                .requestFactory(factory)
                 .defaultHeader("Authorization", "Bearer " + decryptedApiKey)
                 .defaultHeader("Content-Type", "application/json")
                 .build();


max-california (commented):
## 🤖 Agentic Code Review

## Summary

The intent makes sense, but I see one maintainability/performance issue that should be addressed before merging: this change creates a new Apache `HttpClient` per OpenAI integration/client build and never closes it. Because `AiClientFactory` caches the resulting `AiClient`, those connection managers can live for a long time and accumulate across config changes.

## Findings

### Blocking

- **Potential connection-pool/resource leak from per-client `HttpClient` creation**
  - **File:** `src/main/java/org/remus/giteabot/ai/openai/OpenAiProviderMetadata.java:53-58`
  - `HttpClients.createDefault()` creates a new Apache client with its own connection manager. This PR instantiates one every time `buildRestClient(...)` runs, but the returned `RestClient`/`OpenAiClient` path has no corresponding shutdown hook or close path.
  - In this codebase, `AiClientFactory` caches clients by integration ID + `updatedAt` (`src/main/java/org/remus/giteabot/admin/AiClientFactory.java:39-48`). When an integration changes, a new client is built and the old cached client is simply replaced; nothing closes the old Apache client.
  - Over time, especially in environments with multiple integrations or config churn, this can leave behind idle pools/threads/sockets.

  **Suggestion:** avoid creating a fresh transport inside `OpenAiProviderMetadata`. Prefer one of:
  1. inject/reuse a singleton `ClientHttpRequestFactory` / Apache `CloseableHttpClient` bean managed by Spring, or
  2. extend the factory/cache layer so the underlying Apache client is tracked and closed on eviction/rebuild, if per-integration customization is truly required.

### Non-blocking

- **Please document why OpenAI alone needs a custom request factory**
  - **Files:** `src/main/java/org/remus/giteabot/ai/openai/OpenAiProviderMetadata.java`, `pom.xml`
  - The description mentions fixing “several 400er error[s]”, but the code doesn’t explain what behavior in the default `RestClient` transport was problematic, nor why Apache HttpClient 5 fixes it.
  - Since the other providers still use the default request factory, a short comment with the concrete symptom would help future maintainers decide whether this should be generalized, reverted, or tuned further.

- **Missing timeout/pool configuration may make behavior less predictable**
  - **File:** `src/main/java/org/remus/giteabot/ai/openai/OpenAiProviderMetadata.java:53-55`
  - If you are switching transports specifically to improve reliability under concurrent OpenAI use, it would be worth configuring connect/read timeouts and, if relevant, pool settings explicitly rather than relying on `createDefault()`.
  - That’s not necessarily wrong today, but it weakens the stated “fix” because the transport change is otherwise fairly opaque.

## Overall

Good direction, but I’d hold this until the Apache client lifecycle is handled properly. The functional change is plausible; the resource-management story is the main concern.

---

PR #197, opened 2026-06-04
feat: allow any user to interact with bot on PRs when no whitelist is…

### LINKED ISSUE #71: Other authors can talk also to the bot in a PR
**Current behavior**
If I reference the bot with `@ai_bot` in a PR of another user, the bot does not react on this comment.

**Expected behavior**
* The bot should take this comment and react also on the comment with the known context.
* It must be marked that the answer is only a side-clarifiying response.

This feature shall be implemented on branch **develop**

tmseidel:
… configured

    Previously, handlePrComment restricted interaction to the PR author only,
    regardless of whitelist settings. This change introduces dedicated
    access-control logic (isPrCommenterAllowed) for PR comments:

    - No whitelist configured: any user mentioning the bot may interact.
    - Whitelist configured: only the PR author OR users in the whitelist may interact; all others are ignored.

    The global isCallerAllowed method remains unchanged since it is shared
    by other webhook handlers with different access requirements.

    Tests added:
    - noAgentSession_noWhitelist_nonAuthorComment_isAllowed
    - noAgentSession_whitelist_nonAuthorNotInWhitelist_isIgnored
    - noAgentSession_whitelist_prAuthorNotInWhitelist_isAllowed
    - noAgentSession_whitelist_nonAuthorInWhitelist_isAllowed

### DIFF
--- a/src/main/java/org/remus/giteabot/admin/BotWebhookService.java
+++ b/src/main/java/org/remus/giteabot/admin/BotWebhookService.java
@@ -179,10 +179,13 @@ public class BotWebhookService {
     /**
      * Handles a comment on a PR discussion thread.
      * <p>
+     * Access control: if the bot has no {@code userWhitelist}, any user mentioning the bot
+     * may interact.  If a whitelist is configured, only the PR author <em>or</em> users listed
+     * in the whitelist may interact — all other commenters are ignored.
+     * <p>
      * Routes to the agent when an agent session exists for the PR (i.e. the PR was created by the
-     * agent and can be continued).  The agent-session path intentionally skips the PR-author check
-     * because the coding agent is the PR author; human follow-up comments must still reach the agent.
-     * For manually created PRs (no active session), only the PR author may issue commands.
+     * agent and can be continued).  For manually created PRs (no active session), the comment is
+     * routed to the code-review handler.
      */
     @Async
     public void handlePrComment(Bot bot, WebhookPayload payload) {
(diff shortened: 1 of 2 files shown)


max-california (commented):
## 🤖 Agentic Code Review

## Summary

This change looks correct overall and the new tests cover the intended access-control matrix well. I didn’t find a clear blocking bug in the implementation.

## Findings

### Non-blocking

#### 1. PR-comment access rules now diverge from `handleBotCommand`; document that distinction clearly
`src/main/java/org/remus/giteabot/admin/BotWebhookService.java:146-156, 190-198`

`handleBotCommand()` still requires the PR author, while `handlePrComment()` now allows:
- everyone when no whitelist is configured
- PR author or whitelisted users when a whitelist is configured

That seems intentional from the PR description, but it creates two different policies for very similar “mention the bot on a PR” entry points. This is easy to regress later because the names are close and both ultimately route into review/test handlers.

Suggestion:
- add a short Javadoc note on `handleBotCommand()` explicitly saying it is intentionally stricter than `handlePrComment()`
- optionally add one focused test asserting this divergence, so future refactors don’t “simplify” them back together incorrectly

#### 2. Consider a direct unit test for `isPrCommenterAllowed()`
`src/main/java/org/remus/giteabot/admin/BotWebhookService.java:505-527`

Current tests verify behavior indirectly through `handlePrComment()`, which is good for end-to-end coverage. Since access control is now encapsulated in its own method, a few small direct tests would make the policy easier to maintain, especially around edge cases like:
- blank / null commenter username with whitelist configured
- `bot == null`
- case-insensitive whitelist matching

Not required for this PR, but it would strengthen maintainability.

## Verdict

Looks good to merge. The implementation matches the stated behavior, and the added tests exercise the important allow/deny scenarios.

---

PR #201, opened 2026-06-05
release/1.10.0

tmseidel:


max-california (commented):
## 🤖 Agentic Code Review

## Summary

Looks mostly good: the `runOnPrCreation` flag is wired through all four webhook handlers and covered by focused tests, and the migration/entity/UI pieces are present.

I found one likely behavioral regression in PR-comment handling, plus one maintainability gap around persistence/query coverage for the new bot field.

## Findings

### High

- **PR-comment authorization semantics have been broadened beyond the documented whitelist feature**
  - **Files:** `src/main/java/org/remus/giteabot/admin/BotWebhookService.java:191-233`, `src/main/java/org/remus/giteabot/admin/BotWebhookService.java:505-527`
  - Previously, non-agent PR comments were restricted to the PR author. This change removes that check entirely for the code-review path and now allows:
    - anyone, when `userWhitelist` is blank
    - PR author or any whitelisted user, when `userWhitelist` is set
  - That is a significant product behavior change, not just a whitelist refinement. It means on unwhitelisted bots, **any repo user who mentions the bot on any PR can now trigger slash commands / review fallback**, which increases token-spend exposure and changes the trust model.
  - The new docs in `doc/USER_GUIDE.md` describe slash-command routing, but the changelog entry for the whitelist says AI-spending interactions are restricted by whitelist; it does not call out that PR comments were also opened up for non-authors when no whitelist exists.
  - If this is intentional, I’d strongly suggest:
    1. documenting it explicitly as a breaking/behavioral change, and
    2. adding tests that prove this is desired across providers, not just unit tests on `BotWebhookService`.
  - If it is **not** intentional, restore the old PR-author gate for non-agent PR comments and apply the whitelist only as an additional restriction.

### Medium

- **New persisted field is only added to the entity/UI/handlers; repository/query coverage should be verified**
  - **Files:** `src/main/java/org/remus/giteabot/admin/Bot.java:80-81`, `src/main/java/org/remus/giteabot/admin/BotController.java`, bot repository code not touched in this PR
  - `runOnPrCreation` is added to the entity and form, which is fine for normal JPA-managed loads/saves. But this project uses custom repository methods like `findAllWithIntegrations()` / `findByIdWithIntegrations()` in `BotService`.
  - I didn’t see repository code in the diff. If any of those methods use projections/constructor expressions/native SQL instead of full entity loading, this field may silently fail to round-trip in the UI or webhook runtime.
  - Suggestion: verify the `BotRepository` methods are full entity fetches and add a persistence/controller test covering save → reload of `runOnPrCreation`.

### Low

- **Documentation overstates provider symmetry for “create/open” events**
  - **Files:** `doc/PR_WORKFLOWS.md`
  - The doc says “create/open event across all four Git providers,” but the actual event mapping differs:
    - GitHub/Gitea use `opened`
    - GitLab uses `open`
    - Bitbucket handles both `pullrequest:created` and `pullrequest:open`
  - The implementation is correct, but the wording is a bit too uniform and may confuse operators debugging webhook behavior.
  - Suggestion: tweak the wording to say “provider-specific PR/MR open/create events” rather than implying identical semantics.

## Overall

The `runOnPrCreation` feature itself looks correctly implemented and well tested. The main concern is the PR-comment access-control change in `BotWebhookService`, which appears broader than the release notes imply and may be an unintended regression.

---

PR #217, opened 2026-06-15
feature/context-window

### LINKED ISSUE #216: Harden context-window management
**Problem**
The context-window management subsystem — responsible for keeping the
agent's conversation history within the AI model's token limits — don't work at all

If the context-size is reached, especially on agents-based features, the llm responds with a 400 and the agent will cancel.

**Solution**
We need an algorithm to compact the size of the context window
The algorithm should be:

```
┌─────────────────────────────────────────────────────────────────┐
│                    CONTEXT WINDOW BUDGET                        │
│                                                                 │
│  Layer 1: TOOL RESULT TRUNCATION                                │
│  ───────────────────────────────                                │
│  Applied at ingestion time. Caps individual tool outputs        │
│  before they enter the conversation history.                    │
│                                                                 │
│  Layer 2: PROACTIVE COMPACTION                                  │
│  ─────────────────────────────                                  │
│  Triggered between rounds. Removes old conversation units       │
│  when the last AI call's prompt size approaches the limit.      │
│                                                                 │
│  Layer 3: AGGRESSIVE COMPACTION                                 │
│  ──────────────────────────────                                 │
│  Emergency fallback. Drops everything except a summary and      │
│  the most recent N logical units.                               │
│                                                                 │
│  Layer 4: REACTIVE ERROR HANDLING                               │
│  ──────────────────────────────                                 │
│  Last resort. Catches "prompt too long" errors from the AI      │
│  provider and triggers compaction mid-round.                    │
└─────────────────────────────────────────────────────────────────┘
```

tmseidel:


max-california (commented):
## 🤖 AI Code Review

### Diff chunk 1/2
1. Blocking issues

**A. Token-usage estimation uses wrong prompt size (proactive compaction broken for legacy mode)**
In `AgentLoop.run`, the code:
```java
int promptChars = currentMessage == null ? 0 : currentMessage.length();
tokenTracker.record(ctx.session(), turn, promptChars);
```
`promptChars` captures only the latest user message, not the entire prompt sent to the model (history + system prompt + user message). `TokenUsageTracker` uses this value to estimate input tokens when the AI provider does not return explicit token counts (e.g., legacy chat via `ChatTurn.text`). This leads to a severe underestimate of the real prompt size, so proactive compaction will almost never fire for providers that don’t report usage. For providers that do report usage the estimated fallback is irrelevant, but for those that don’t (any legacy path, llama.cpp, Ollama without usage, etc.) this breaks Layer 2 (proactive compaction) as designed.

**Fix:** Compute the approximate total prompt character size before the AI call, e.g. sum of character lengths from history messages, system prompt, and current message, and pass that total as `promptChars`, or accept the total directly in `record`. This must be corrected to regain reliable context-window pressure detection.

**B. Docker Compose default `APP_PUBLIC_URL` changed to a hardcoded IP**
`docker-compose.yml` changed the default from `http://localhost:8080` to `167.235.231.84:8080`. This looks like an accidental commit of a local/testing address and will mislead users or break local development. Revert to `http://localhost:8080` or use a placeholder like `${APP_PUBLIC_URL:-http://localhost:8080}`.

**C. ChatTurn record signature change – all callers must be updated**
`ChatTurn` now requires `inputTokens` and `outputTokens` in its constructor. The diff updates the observable client implementations (`AnthropicAiClient`, etc.), but if any other part of the codebase (not shown in the diff) constructs `ChatTurn` directly with the old three-argument constructor, the project will not compile. Verify that there are no remaining direct `new ChatTurn(…)` calls outside the changed files. All provider implementations and test code in the diff have been updated, but this is a high-risk change – ensure compilation across the entire project.

2. Non-blocking suggestions

- **Docker Compose hardcoded budget variables**  
  The new lines `AGENT_BUDGET_MAX_CONTENT_ROUNDS: 20` (etc.) are static, making them impossible to override with environment variables. Consider using the same pattern as other variables: `${AGENT_BUDGET_MAX_CONTENT_ROUNDS:-20}` to allow runtime configuration.

- **`callAiWithRetry` only retries once after aggressive compaction**  
  If the compacted history is still too large after aggressive compaction, the second attempt will throw without further mitigation. This is acceptable for a last-resort, but consider whether a more gradual fallback (multiple retries with increasingly aggressive compaction) is warranted for robustness. Not a blocker.

- **`HistoryCompactor` early-exit condition multiplies `keepLastN` by 3**  
  The condition `history.size() <= keepLastN * 3` is a heuristic and may miss compaction for small histories that exceed the budget. While harmless, it’s an opaque magic number – add a comment explaining the rationale.

3. Tests

- **Missing test for proactive compaction with legacy (no token usage) mode**  
  No test verifies that `TokenUsageTracker` triggers compaction after several rounds when the AI client returns zero token counts (legacy path). Write an `AgentLoop` test that uses a mock AI returning `ChatTurn.text("...")` and a large prompt to ensure `shouldCompactProactively` fires and the history is compacted. The current `TokenUsageTrackerTest` only tests with explicit tokens.

- **Test for incorrect `promptChars` calculation**  
  Once the calculation is fixed, add a test confirming that the total prompt size is used for estimation (or that explicit token counts take precedence). This protects the fix.

4. Overall assessment

The feature is well-designed and faithfully mirrors the documented context-window strategy. The core algorithms (tool-result truncation, history compaction with atomic tool-call grouping, token tracking) are sound. However, the two blocking issues – the estimation bug and the Docker compose default – must be resolved before merge. With those fixed, the changes should safely improve long-running agent robustness.

### Diff chunk 2/2
1. Blocking issues

- **Breaking `ChatTurn` constructor change may cause compilation failures or incorrect token reporting.**  
  The new `ChatTurn` record requires `inputTokens` and `outputTokens` in its canonical constructor. All call sites that previously used `new ChatTurn(text, calls, stopReason)` must be updated to supply the token counts. The diff shows test files updated (e.g., `StubAiClient`, `ChatTurnTest`), but the actual AI client implementations (`OpenAiClient`, `AnthropicAiClient`, `GoogleAiClient`, `OllamaClient`, `LlamaCppClient`) and other producers of `ChatTurn` are not visible in this diff. If any of those still use the old three-argument constructor, the build will fail. Moreover, unless the token counts are correctly extracted from the provider response and passed, the entire proactive compaction layer will use invalid values (e.g., 0 tokens) and never trigger, defeating the context window management.  

  **Fix:** Verify that every `new ChatTurn(…)` in the codebase now includes proper `inputTokens` and `outputTokens` derived from the AI provider’s response metadata. Confirm this with a full compilation and, ideally, integration tests that validate the token reporting for each supported provider.

- **Missing integration of reactive error handling (Layer 4) is not covered by the diff.**  
  The documentation mentions catching “prompt too long” errors and triggering compaction. The only visible change is making `AbstractAiClient#isPromptTooLongError` public in a test helper, but the actual usage in `AgentLoop` (catching `HttpClientErrorException`, invoking compaction, and retrying) is absent from this diff. If that logic is incomplete or missing, sessions hitting the context limit will simply fail without any recovery.  

  **Fix:** Confirm that the full reactive‑error path exists in `AgentLoop` or the calling service, and that it correctly calls `HistoryCompactor` (or the aggressive fallback) before re‑sending the request. Add at least one integration‑level test simulating a provider error and verifying that compaction retries succeed.

2. Non-blocking suggestions

- **Validate `contextWindowTokens` at the point of use.**  
  `AgentServiceFactory#getContextWindowTokens` falls back to 200,000 when the `AiIntegration` is absent or has a default value. Consider logging a warning when the configured value is zero or negative, as that effectively disables compaction without the operator’s intent. A small guard would improve observability.

- **Potential performance concern in logical‑unit grouping (O(n²) worst case).**  
  The described algorithm walks back through the history and, for each tool response, scans backward to find the matching assistant call. In very long conversations the worst‑case complexity could be noticeable. If the implementation matches the description, consider using an index map to make grouping O(n).

- **Consider making `TokenUsageTracker` thread‑safe if concurrent session access is possible.**  
  While session runs are typically serialised by locks, if the tracker is ever shared across threads (e.g., in tests or future use), the current mutable state on the session entity via the service may not be safe. Document the concurrency assumption or add a comment that the tracker assumes single‑writer semantics.

3. Tests

- **Add tests for `HistoryCompactor` logical‑unit grouping, especially edge cases.**  
  The diff introduces thorough tests for `TokenUsageTracker` and `ToolResultTruncator`, but no tests for the grouping logic itself (orphan tool calls, orphan tool responses, consecutive plain messages). These are critical to avoid corrupting the conversation flow.

- **Add integration tests for the full compaction loop.**  
  Simulate a multi‑round agent session where the context approaches the limit, trigger proactive compaction, verify that old units are summarised and dropped without breaking tool‑call continuity. Also test the aggressive fallback when normal compaction cannot shrink enough.

- **Ensure each AI client correctly extracts `inputTokens`/`outputTokens`.**  
  Write tests (or contract tests) that exercise each provider’s response parsing and confirm the token fields are populated in the resulting `ChatTurn`. Without this, the compaction trigger may remain silent for all but one provider.

4. Overall assessment

The feature design (documented multi‑layer compaction, token tracking, truncation) is sound, and the core units introduced here (`TokenUsageTracker`, `ToolResultTruncator`) are well‑tested. The major risk lies in the `ChatTurn` constructor change and the absence of visible integration for reactive error handling. Before merging, verify that the entire codebase compiles and that all AI clients supply the new token data. Once those gaps are addressed, the change will be a valuable improvement to long‑running agent reliability.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

Good direction overall: the new loop-level truncation/compaction layers are sensible and the documentation is helpful. However, I found **two blocking correctness issues** and **one likely deployment breakage** in the persistence/config changes.

## Blocking

### 1. New JPA columns were added without a matching DB migration
**Files:**  
- `src/main/java/org/remus/giteabot/admin/AiIntegration.java:67`  
- `src/main/java/org/remus/giteabot/agent/session/AgentSession.java:129`  
- `src/main/java/org/remus/giteabot/agent/session/AgentSession.java:136`

This PR adds three new persisted columns:

- `ai_integrations.context_window_tokens`
- `agent_sessions.total_input_tokens`
- `agent_sessions.total_output_tokens`

I checked the repo and could not find any accompanying Flyway/Liquibase/schema migration in this change. Since the project already references migration-based schema management in comments (`AgentSession` mentions Flyway V10), this will fail against existing databases at startup or on first access.

**Why this matters:** existing deployments will not have these columns, so Hibernate/JPA reads/writes will break.

**Suggested fix:** add an explicit schema migration for all three columns, including safe defaults for existing rows (`200000`, `0`, `0`).

---

### 2. `TokenUsageTracker` underestimates prompt size because it only counts `currentMessage`
**File:** `src/main/java/org/remus/giteabot/agent/loop/AgentLoop.java:100`  
Related: `src/main/java/org/remus/giteabot/agent/loop/TokenUsageTracker.java`

The design doc correctly says proactive compaction should use the **last AI call’s full prompt size**. But the implementation records:

```java
int promptChars = currentMessage == null ? 0 : currentMessage.length();
tokenTracker.record(ctx.session(), turn, promptChars);
```

When the provider does not return `inputTokens()`, the fallback estimate in `TokenUsageTracker` is based only on `promptChars`, which here is just the current user message — **not** `history + system prompt + current message + tool descriptors`.

That means proactive compaction will almost never trigger for providers that don’t report usage, precisely when this feature is needed most.

**Why this matters:** long histories can keep growing until the provider rejects the prompt, defeating Layer 2 and pushing everything into emergency retry behavior.

**Suggested fix:** estimate from the actual request payload size, e.g. include:
- serialized/history message content,
- system prompt,
- current message,
- possibly tool schema/descriptor text when in native-tool mode.

Even a rough full-prompt character sum is much better than just `currentMessage.length()`.

## High

### 3. `APP_PUBLIC_URL` default was changed to a bare IP/port without a scheme
**File:** `docker-compose.yml:15`

This changed from:

```yaml
APP_PUBLIC_URL: ${APP_PUBLIC_URL:-http://localhost:8080}
```

to:

```yaml
APP_PUBLIC_URL: ${APP_PUBLIC_URL:-167.235.231.84:8080}
```

This is risky for two reasons:

1. It removes the URL scheme (`http://`), so anything expecting a valid absolute URL may break.
2. It hardcodes what looks like an environment-specific public IP into the default docker-compose config.

**Why this matters:** local/dev deployments will now inherit an invalid or surprising default URL.

**Suggested fix:** keep the prior safe default (`http://localhost:8080`) unless there is a very strong reason otherwise, and if needed document how operators override it.

## Medium

### 4. Persisted session compaction can keep growing because it adds a new summary message every time
**File:** `src/main/java/org/remus/giteabot/agent/session/AgentSessionService.java:175`

`compactContextWindow()` removes old messages and then does:

```java
session.addMessage("user", summary);
```

On repeated follow-up runs, this will keep adding fresh summary messages to the persisted session. Since `toAiMessages()` replays those summary messages, the compacted history can accumulate multiple synthetic summaries over time.

Not necessarily fatal, but it weakens the goal of stabilizing history size across many follow-ups.

**Suggested improvement:** if the oldest surviving message is already a compaction-summary marker, replace/update it instead of appending another summary message.

## Low / Nit

### 5. `HistoryCompactor` summary role choice may be a little misleading
**File:** `src/main/java/org/remus/giteabot/agent/loop/HistoryCompactor.java:126`

The synthetic compaction message is inserted as role `"user"`. That works mechanically, but it is not really user input. A system/developer-style note would be semantically cleaner if your provider abstraction supports it; otherwise this is acceptable, just worth considering.

---

Overall: the loop changes themselves look promising, but I would not merge this without the DB migration and the prompt-size estimation fix.

---

tmseidel:
@max-california The required DB-Migration is V27__agent_session_token_tracking.sql

max-california:
🤖 Sorry, I did not understand that command.

This bot is not configured to run code reviews, so I can only respond to the slash commands listed below. Anything else will be ignored.

No interactive commands are configured for this bot.

PR #219, opened 2026-06-17
    fix: avoid ObjectNotFoundException when saving detached AgentSession after compaction

tmseidel:


max-california (commented):
## 🤖 Agentic Code Review

## Summary

This change looks directionally correct: reloading a managed `AgentSession` before mutating it is a sensible fix for the detached-entity / `ObjectNotFoundException` path described in the PR. The `compactContextWindow` change in particular matches the Hibernate failure mode.

I found one blocking correctness issue around caller-visible state, plus one test gap that makes the risky behavior easy to regress.

## Blocking

### 1. Callers still keep using the stale detached `session`, so some updates become invisible immediately after these service calls
The service now mutates a freshly loaded managed entity and returns it, but many callers ignore the return value and continue reading fields from the original detached `session`.

Concrete examples:

- `src/main/java/org/remus/giteabot/agent/IssueImplementationService.java:373`
  ```java
  sessionService.compactContextWindow(session);
  String branchName = session.getBranchName();
  ```
- `src/main/java/org/remus/giteabot/agent/IssueImplementationService.java:412`
  ```java
  sessionService.setBranchName(session, branchName);
  ...
  if (session.getPrNumber() == null) {
  ```
- `src/main/java/org/remus/giteabot/agent/IssueImplementationService.java:431`
  ```java
  sessionService.setPrNumber(session, prNumber);
  ...
  notificationService.postFollowUpSuccessComment(..., session.getPrNumber());
  ```

Before this PR, mutating the passed object meant the caller’s reference reflected the new values immediately. After this PR, that is no longer true unless the caller reassigns:

```java
session = sessionService.setPrNumber(session, prNumber);
```

This is especially risky for:
- `setBranchName`
- `setPrNumber`
- `setGeneratedIssueNumber`
- `setStatus`
- `recordPlan`
- `compactContextWindow`

The most obvious user-visible bug is in follow-up handling: after `setPrNumber(session, prNumber)`, `session.getPrNumber()` on the old instance may still be `null`, so downstream comment generation can behave incorrectly.

**Suggested fix:** update all call sites to use the returned `AgentSession` consistently, e.g.
```java
session = sessionService.compactContextWindow(session);
session = sessionService.setStatus(session, ...);
session = sessionService.setBranchName(session, branchName);
session = sessionService.setPrNumber(session, prNumber);
```

If you want to make this safer long-term, consider either:
- documenting that these methods return the new authoritative managed instance and callers must reassign, or
- changing the API to accept `sessionId` instead of `AgentSession` for write methods, which makes the replacement semantics explicit.

## Medium

### 2. The updated unit test no longer verifies persistence interaction
`src/test/java/org/remus/giteabot/agent/session/AgentSessionServicePlanPersistenceTest.java`

The renamed test still says “Saves”, but it now only asserts in-memory mutation of the mocked managed entity. It no longer verifies the repository interaction that is now essential to the fix:

```java
when(repository.getReferenceById(42L)).thenReturn(managed);
```

Given the new contract, the important assertions are:
- `getReferenceById` is called with the input session ID
- the returned instance is the one mutated and returned

Without that, a future refactor could accidentally go back to mutating the detached argument and this test would still pass.

**Suggested fix:** add `verify(repository).getReferenceById(42L);` in both tests, and consider renaming the first test to avoid “Saves” wording since there is no explicit `save()` anymore.

## Overall

Good fix idea, but I wouldn’t merge it until the call sites are updated to use the returned managed entity. That’s the main behavioral change introduced by replacing `save(detached)` with “load managed + mutate”.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

The updated approach fixes the earlier “stale caller object” problem for scalar fields, and the compaction path still avoids `merge()` on the detached entity. That said, there is still one blocking correctness issue: `addMessage` now returns the caller’s detached `session` without updating it, which breaks the same caller-visible contract this patch is trying to preserve elsewhere.

## Blocking

### 1. `addMessage` persists to the managed entity but returns an unmodified detached session
In `src/main/java/org/remus/giteabot/agent/session/AgentSessionService.java`:

```java
public AgentSession addMessage(AgentSession session, String role, String content) {
    AgentSession managed = repository.getReferenceById(session.getId());
    managed.addMessage(role, content);
    return session;
}
```

All the other setter-style methods in this patch now mutate both:
- the caller’s `session`
- the managed entity used for persistence

But `addMessage` only mutates `managed` and returns the original detached `session`, which still lacks the newly added message.

That creates an inconsistent API and can break any in-process logic that expects the returned/current session to reflect the appended message. It also makes this method behave differently from `recordPlan`, `setBranchName`, `setPrNumber`, etc.

**Suggested fix:** either:
- mirror the new dual-write pattern here too:
  ```java
  session.addMessage(role, content);
  AgentSession managed = repository.getReferenceById(session.getId());
  managed.addMessage(role, content);
  return session;
  ```
  or
- return `managed` consistently and update all callers to rebind.

Given the rest of this patch chose “mutate caller object too”, `addMessage` should probably follow the same contract.

## Medium

### 2. Test coverage still only exercises `recordPlan`
The added test updates are good, but they only cover `recordPlan`. The risky behavioral change in this PR affects multiple methods:
- `addMessage`
- `setBranchName`
- `setPrNumber`
- `setGeneratedIssueNumber`
- `setStatus`
- `recordTokenUsage`
- `compactContextWindow`

At minimum, `addMessage` deserves a regression test now, because it currently appears to violate the intended contract and would have caught the issue above.

## Overall

Closer, but I’d still block this revision until `addMessage` is made consistent with the rest of the “mutate detached object + persist via managed entity” strategy.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

The compaction fix itself still looks right, but this revision introduces a new blocking issue: `addMessage` now adds the same message twice whenever the caller passes a managed `AgentSession` (which is a common case inside a transactional flow). The rest of the scalar-field methods are less risky, but `addMessage` as written is not safe to merge.

## Blocking

### 1. `addMessage` can duplicate messages by mutating both `session` and `managed`
In `src/main/java/org/remus/giteabot/agent/session/AgentSessionService.java`:

```java
public AgentSession addMessage(AgentSession session, String role, String content) {
    session.addMessage(role, content);
    AgentSession managed = repository.getReferenceById(session.getId());
    managed.addMessage(role, content);
    return session;
}
```

This assumes `session` and `managed` are distinct objects. That is true for a detached entity, but not guaranteed here. If the caller is already inside the same persistence context and passes a managed `AgentSession`, `getReferenceById(session.getId())` will resolve to the same managed instance already associated with the session. In that case both calls target the same entity and append two `ConversationMessage`s.

This is a real risk because these service methods are themselves `@Transactional`, and callers may already be operating on entities loaded in the same transaction.

**Suggested fix:** avoid dual mutation for collection updates. Safer options:
- return the managed entity consistently and require callers to rebind, or
- detect identity/equality and only add once when `managed == session`.

Example minimal guard:

```java
AgentSession managed = repository.getReferenceById(session.getId());
if (managed == session) {
    managed.addMessage(role, content);
} else {
    session.addMessage(role, content);
    managed.addMessage(role, content);
}
return session;
```

But architecturally, returning the managed instance consistently is cleaner than trying to mirror mutable collections across two objects.

## Medium

### 2. The current tests still don’t cover the highest-risk path
`src/test/java/org/remus/giteabot/agent/session/AgentSessionServicePlanPersistenceTest.java` only covers `recordPlan`, which deals with scalar fields. The subtle failure mode in this PR is around collections and entity identity, especially `addMessage` and `compactContextWindow`.

At minimum, I’d add:
- a test for `addMessage` when `repository.getReferenceById(id)` returns the same instance as the input session
- a test for `addMessage` when it returns a different instance
- ideally, an integration-style test covering `compactContextWindow()` followed by `addMessage()`

## Overall

I’d block this revision on `addMessage`. The detached-entity merge fix is valid, but the current dual-write approach is unsafe for collection mutations because it can duplicate messages when the input entity is already managed.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

This is closer: the `managed == session` guard removes the obvious duplicate-message bug, and the core compaction fix still makes sense. I still see one blocking correctness issue, though: `compactContextWindow()` now returns the managed instance on the compaction path, but returns the original `session` on the no-op path. That inconsistent contract is likely to produce very subtle caller behavior.

## Blocking

### 1. `compactContextWindow()` now has inconsistent return semantics depending on whether compaction happens
In `src/main/java/org/remus/giteabot/agent/session/AgentSessionService.java`:

- no compaction needed:
  ```java
  return managed;
  ```
- threshold not reached:
  ```java
  return managed;
  ```
- compaction happens:
  ```java
  return managed;
  ```

Actually in this latest diff, all visible branches return `managed`, which is good inside the method itself — but it is still inconsistent with the rest of the service methods, which intentionally return the caller’s original `session` after dual-writing to both objects.

That means callers now have two different contracts in the same service:
- `recordPlan`, `setBranchName`, `setPrNumber`, `setGeneratedIssueNumber`, `setStatus`, `addMessage` → return caller object
- `compactContextWindow` → return managed object

This matters because callers that ignore the return value get different behavior depending on which method they called, and callers that *do* use the return value may silently switch object identity mid-flow only for compaction.

Given the effort in this patch to preserve “subsequent reads on the caller’s object”, `compactContextWindow()` is the outlier. After compaction, the caller’s detached `session` still contains the old messages, while the returned `managed` has the compacted history. That split-brain state is hard to reason about.

**Suggested fix:** pick one contract and apply it consistently across the service:
- either always return the managed instance and update callers to rebind, or
- for `compactContextWindow`, also synchronize the caller object to match the managed state before returning it.

I’d strongly prefer the first option, because mirroring collection state between detached and managed entities is fragile.

## Medium

### 2. Tests still don’t exercise the collection/identity-sensitive code paths
The updated `recordPlan` tests are fine for scalar fields, but the risky logic here is really in:
- `addMessage`
- `compactContextWindow`

Recommended additions:
- a unit test where `getReferenceById()` returns the same instance as `session` for `addMessage`
- a unit test where it returns a different instance
- a test that verifies `compactContextWindow()` behavior from a caller perspective, especially what object is returned and what happens to the original `session`

## Overall

The PR is close, but I’d still hold it until the service settles on a single, consistent return/update contract. Right now `compactContextWindow()` behaves fundamentally differently from the other write methods, which is likely to cause hard-to-debug downstream issues.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

This revision is much stronger. The main detached-entity problem is addressed in a cleaner way now: `compactContextWindow` operates by `sessionId`, callers are rebound where needed, and the loop no longer repeatedly merges a detached `AgentSession` with a stale child collection. The new batching in `AgentLoop` also reduces transaction churn.

I didn’t find a clear blocker from the diff provided. Most of the remaining risk is around timestamp-ordering assumptions and the changed persistence contract, which I’d treat as follow-up/test-hardening rather than merge blockers.

## Medium

### 1. `flushMessages()` relies on microsecond-distinct timestamps to preserve message order
`src/main/java/org/remus/giteabot/agent/session/AgentSessionService.java`

```java
Instant base = Instant.now();
for (int i = 0; i < messages.size(); i++) {
    PendingMessage msg = messages.get(i);
    managed.addMessage(msg.role(), msg.content(), base.plus(i, ChronoUnit.MICROS));
}
```

and the corresponding comment in `AgentSession`:

```java
// ... must carry distinct, call-ordered timestamps rather than relying on the @PrePersist default
```

This is a reasonable workaround, but it couples ordering correctness to DB timestamp precision and to `createdAt` being the sole replay sort key. If this runs against a database/dialect that truncates more aggressively than expected, messages from the same batch could still collapse to identical timestamps and replay in undefined set order.

**Suggestion:** if you keep this design, I’d add an integration test that persists a round containing multiple pending messages and verifies `toAiMessages()` returns them in the original order. Longer term, a monotonic sequence/order column would be more robust than overloading `createdAt` for intra-round ordering.

## Medium

### 2. The service contract changed substantially; some tests still need to pin the new behavior
A lot of logic now depends on “persist by `id`, rebind when compaction returns a managed entity, skip persistence for transient sessions, batch flush per round”. The diff already updates several tests, which is good, but the following contracts are important enough to deserve explicit coverage if not already present outside the truncated diff:

- `AgentLoop#flushRound()` is a no-op for `session.id == null` (review agent path)
- `flushMessages()` preserves message order across `user` / `assistant` / `tool`
- `compactContextWindow(sessionId)` can be followed by another loop run without surfacing deleted child references
- `recordPlan()` callers now use the returned managed entity or an out-of-band source (`outcome.payload()`), rather than reading stale detached state

None of these look obviously broken from the code shown; I’m just calling out where regressions would be costly.

## Low

### 3. `recordPlan()` now returns a managed entity, but callers are split between rebinding and bypassing session state
`IssueImplementationService` now correctly prefers `implementationResult.plan()` in the important places, which avoids stale detached reads. That’s good. Still, the service layer contract is now mixed:

- some writes are effectively “fire and forget” (`setStatus`, `setPrNumber`, etc.)
- `recordPlan()` returns a managed entity, but callers often use the plan payload instead
- `compactContextWindow()` explicitly requires rebinding

That’s workable, but it’s subtle.

**Suggestion:** document this in `AgentSessionService` class/method Javadocs, especially that methods persist via a freshly fetched managed entity and callers should not assume the passed `AgentSession` instance stays authoritative after a transaction boundary.

## Overall

Looks good overall. The architectural direction is much better than the earlier dual-write attempts, and the `sessionId`-based compaction plus batched message flushes fit the bug being fixed.

If you want one extra safety net before merge, I’d prioritize an integration test proving that:
1. a session is compacted,
2. a subsequent loop flushes new messages,
3. `toAiMessages()` replays the expected ordered conversation without `ObjectNotFoundException`.

---

PR #222, opened 2026-06-18
 feat(refactoring): Refactoring: reviewDiff moved from AbstractAiClient to CodeReviewService

tmseidel:


max-california (commented):
## 🤖 Agentic Code Review

## Summary

The refactor mostly makes sense, but I found **two blocking regressions** and **one suspicious config change** that should be reverted or fixed before merge.

## Blocking issues

### 1. Review errors are no longer audited
`CodeReviewService` now owns diff chunking/retry, but the error-auditing behavior did not move with it.

Before this PR:
- `AbstractAiClient.reviewDiff(...)` caught per-chunk failures and called `reportError(e)`.

After this PR:
- `CodeReviewService.reviewDiffWithChunking(...)` catches chunk failures, but **never records them**.
- `AuditingAiClient.submitReviewPrompt(...)` intentionally does **not** audit review errors (`AuditingAiClientTest` even asserts that).

This means failed review chunks are now silently omitted from AI usage/error telemetry.

**Relevant files**
- `src/main/java/org/remus/giteabot/review/CodeReviewService.java`
- `src/main/java/org/remus/giteabot/ai/AuditingAiClient.java`
- `src/test/java/org/remus/giteabot/ai/AuditingAiClientTest.java`

**Why it matters**
Operationally this is a regression: review failures will stop showing up in the audit/usage pipeline even though chat failures still do.

**Suggested fix**
Either:
- restore review error recording in `AuditingAiClient.submitReviewPrompt(...)`, or
- introduce an explicit review-error audit hook in `CodeReviewService` when a chunk fails.

Right now the responsibility moved, but the auditing did not.

---

### 2. `OllamaClient.isPromptTooLongError()` can now throw `NullPointerException`
In `OllamaClient` this PR removed the null check:

```java
```

If the provider returns a 400 without a response body, `body` is `null` and this method throws `NullPointerException` while handling an error path.

**Relevant file**
- `src/main/java/org/remus/giteabot/ai/ollama/OllamaClient.java:113-117`

**Why it matters**
This breaks retry classification for exactly the failure path this refactor depends on.

**Suggested fix**
Keep the previous null guard, or delegate to the default implementation when `body == null`.

## High risk / likely unintended

### 3. `docker-compose.yml` now defaults `APP_PUBLIC_URL` to a hard-coded public IP
This change is unrelated to the refactor and changes the default from:

```yaml
http://localhost:8080
```

to:

```yaml
http://167.235.231.84:8080/
```

**Relevant file**
- `docker-compose.yml`

**Why it matters**
This is surprising and risky:
- changes local/dev behavior,
- bakes an environment-specific public IP into default config,
- may break links/callbacks for anyone running locally,
- could accidentally expose incorrect URLs in generated content.

Unless there is a very specific documented reason, this should not be part of this PR.

## Minor notes

### 4. Logging in `CodeReviewService` lost the actual model name
This new line:

```java
log.info("Requesting code review from AI provider model={} (via chunking)", "see aiClient");
```

is effectively placeholder text and less useful than before. Not blocking, but worth improving if the model can be surfaced cleanly.

### 5. Generated Lombok setter Javadoc in `AbstractAiClient` is malformed
The `@Setter` replacement produced odd Javadoc:

```java
 *  Attaches an
 *  that receives token usage and error
 *  reports for this client. Set by the
 * .
```

Not a functional issue, but it makes the API docs worse than the explicit setter did.

---

If you fix the auditing regression and the Ollama null handling, the structural refactor itself looks reasonable.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

The refactor mostly makes sense: moving diff chunking out of `AbstractAiClient` and into `CodeReviewService` is a reasonable separation of concerns. However, I found **two blocking correctness regressions** introduced by the interface changes, plus one likely compile/runtime issue around null handling that needs attention.

## Blocking issues

### 1. `AiClient.isPromptTooLongError()` now NPEs on providers that return no response body
**File:** `src/main/java/org/remus/giteabot/ai/AiClient.java`  
**Lines:** around 99–111

The default implementation used to guard against `null`:

```java
if (body == null) {
    return false;
}
```

That guard was removed, but the method still immediately does:

```java
```

So any `HttpClientErrorException` with a null response body now throws `NullPointerException` during retry classification. Since `CodeReviewService.reviewSingleChunk()` calls `aiClient.isPromptTooLongError(e)` inside the error path, this can mask the real provider error and break review handling.

This same regression was also introduced in `OllamaClient.isPromptTooLongError()`.

**Suggested fix:** restore the null check in the default interface method and in provider overrides that dereference `body`.

---

### 2. `AuditingAiClient` no longer implements the expanded `AiClient` contract
**File:** `src/main/java/org/remus/giteabot/ai/AuditingAiClient.java`

`AiClient` now requires two additional methods:

- `void reportError(Throwable error);`
- `String getModel();`

`AuditingAiClient` in the diff does not implement either of them. Unless there is some other change outside the truncated diff, this is a compile break.

Even if it compiles via additional unseen edits, it should still explicitly delegate both methods to the wrapped client:

```java
@Override
public void reportError(Throwable error) {
    delegate.reportError(error);
}

@Override
public String getModel() {
    return delegate.getModel();
}
```

This is especially important because `CodeReviewService` now directly depends on both methods (`aiClient.getModel()` for logging and `aiClient.reportError(e)` for chunk-failure auditing).

---

## Medium severity

### 3. Review-path auditing assumptions are now inconsistent with the actual decorator behavior
**Files:**
- `src/main/java/org/remus/giteabot/review/CodeReviewService.java`
- `src/main/java/org/remus/giteabot/ai/AuditingAiClient.java`
- `src/test/java/org/remus/giteabot/ai/AuditingAiClientTest.java`

The new design assumes review errors are audited explicitly by `CodeReviewService` via `aiClient.reportError(e)`, and therefore `AuditingAiClient.submitReviewPrompt()` intentionally does **not** wrap exceptions.

That’s okay for chunked review calls coming from `CodeReviewService`, but it makes auditing dependent on every caller remembering to invoke `reportError()` manually. The old design kept review error auditing closer to the client path.

At minimum, this is brittle API design: `submitReviewPrompt()` looks like a normal provider call, but unlike `chat()` it is not centrally audited by the decorator.

**Suggestion:** either:
- document very clearly that callers of `submitReviewPrompt()` must audit failures themselves, or
- make `AuditingAiClient.submitReviewPrompt()` audited too, and ensure `CodeReviewService` doesn’t double-record.

Right now the split responsibility is easy to misuse.

---

## Low severity / nits

### 4. `handleReviewSubmitted()` null-check was weakened
**File:** `src/main/java/org/remus/giteabot/review/CodeReviewService.java`  
**Line:** around 347

This changed from:

```java
if (latestReview == null || latestReview.getId() == null) {
```

to:

```java
if (latestReview.getId() == null) {
```

That’s only safe if `orElse(null)` was also changed elsewhere; in the shown code it still appears to be:

```java
.orElse(null);
```

If so, this is another NPE path. If the full file really still uses `orElse(null)`, this is blocking too. If the surrounding code changed outside the visible snippet, ignore this comment.

---

## Overall

The architectural direction is good, but I would not merge as-is until the `AiClient` null-handling and `AuditingAiClient` contract issues are fixed.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

Looks good overall. I didn’t find any blocking correctness issues in the reviewed AI client refactor/additions. The new interface defaults preserve backward compatibility, the auditing wrapper forwards the new capability cleanly, and provider implementations appear consistent with the existing abstractions.

## Findings

### Nit / maintainability

- **`src/main/java/org/remus/giteabot/ai/AbstractAiClient.java:136-147`**  
  The debug log prints the entire `newUserMessage` content:
  ```java
  log.debug("New User Message ({} chars):\n{}", newUserMessage.length(), newUserMessage);
  ```
  In this codebase, user messages can contain PR diffs, comments, and tool results, which may be large or sensitive. Elsewhere in `CodeReviewService` you already truncate previews before logging. Consider aligning this log with that safer pattern, e.g. log only length plus a bounded preview.

- **`src/main/java/org/remus/giteabot/ai/AbstractAiClient.java:146`**  
  `newUserMessage.length()` assumes the argument is non-null. Current callers seem to pass non-null strings, and the interface docs say the message *may be empty* rather than null, so this is probably fine today. Still, if any future tool-calling flow uses `null` to mean “no new user turn”, this method will NPE before the provider call. A small null-safe guard here would make the base class more robust.

---

PR #226, opened 2026-06-20
feature/move-review-settings

### LINKED ISSUE #220: Move Code-Review specific properties
**Current behavior**
On the edit-page of an AI-Integration you can edit the following properties:
* Max Tokens
* Max Diff Chars/Chunk
* Max Diff Chunks
* Retry Truncated Chars
This is misleading because they are not bound to the AiIntegration. In addition it clutters the UI.

**Expected behavior**
Since these properties are settings relevant for executing the simple PR Workflow they should be handled as `org.remus.giteabot.prworkflow.WorkflowParamField` instead of properties of `org.remus.giteabot.admin.AiIntegration`
* On the Workflow-Edit page these 4 settings can be edited in the `review` section
* They are persisted as Workflow-Parameter
* The JPA Entity AiIntegration does not have these 4 properties
* A flyway migration is created (no data is transfered, just the schema)

**Implementation nodes**
* Remove the properties from `org.remus.giteabot.admin.AiIntegration`
* Add WorkflowParams to the `review` PrWorkflow
* Add the UI controls to the `review` section of the Workflow Edit-Page
* Use the old default values for the new parameter
* target-Branch: `develop`
* Currently you can edit the settings as part of an AiIntegration Url to the edit-page is ai-integrations/[id]/edit
* The mentioned settings are the following properties: org.remus.giteabot.admin.AiIntegration#maxTokens, org.remus.giteabot.admin.AiIntegration#maxDiffCharsPerChunk, org.remus.giteabot.admin.AiIntegration#maxDiffChunks and org.remus.giteabot.admin.AiIntegration#retryTruncatedChunkChars
* These settings are used in the
* The method org.remus.giteabot.review.CodeReviewService#reviewPullRequest which has connections to the properties must be unchanged
*These values are given to each implementation of org.remus.giteabot.ai.AiClient and are used in the implemting methods of org.remus.giteabot.ai.AiClient#reviewDiff(java.lang.String, java.lang.String, java.lang.String, java.lang.String, java.lang.String)

### LINKED ISSUE #221: Move Code-Review specific properties
Originates from #220

## Quality assessment

The issue intent is clear and technically plausible, but the current draft is incomplete and slightly contradictory. It mixes desired behavior with implementation constraints, does not define how the new workflow params are surfaced to existing review execution without changing CodeReviewService.reviewPullRequest, and lacks explicit acceptance criteria around migration, defaults, and backward compatibility for existing integrations/workflow configs.

## Summary
Move the code-review execution settings currently stored on `AiIntegration` into workflow-level configuration for the `review` PR workflow.

## Problem
The following settings are currently editable on the AI integration edit page (`ai-integrations/[id]/edit`):
- `maxDiffCharsPerChunk`
- `maxDiffChunks`
- `retryTruncatedChunkChars`

This is misleading because these settings do not conceptually belong to an AI integration itself. They are only relevant when executing the simple PR review workflow. Keeping them on `AiIntegration` also adds unnecessary clutter to the AI integration UI.

## Desired outcome
These 3 settings should become workflow parameters for the `review` workflow and no longer be persisted on `org.remus.giteabot.admin.AiIntegration`.

### Data model
- Remove these properties from `org.remus.giteabot.admin.AiIntegration`:
  - `maxDiffCharsPerChunk`
  - `maxDiffChunks`
  - `retryTruncatedChunkChars`
- Add corresponding `WorkflowParamField`s to the `review` workflow (`org.remus.giteabot.prworkflow.review.ReviewWorkflow` or its related review workflow config types).
- Persist the values as workflow parameters via the existing workflow configuration mechanism.

### UI
- Remove the 3 controls from the AI integration edit page.
- Add the 3 controls to the `review` section on the workflow edit page.
- Use the same default values as today (defaults can be set in application.properties):
  - `maxDiffCharsPerChunk = 120000`
  - `maxDiffChunks = 8`
  - `retryTruncatedChunkChars = 60000`

### Database migration
- Add Flyway migrations for supported databases to remove the obsolete schema columns from `ai_integrations`.
- Do not migrate existing values into workflow configuration data.
- Schema-only migration is sufficient.

## Important constraints
- The review execution path should continue to behave as before, except that these 4 values are now sourced from workflow parameters instead of `AiIntegration`.
- `org.remus.giteabot.review.CodeReviewService#reviewPullRequest(...)` must remain unchanged.
- The values must still be passed through to each `org.remus.giteabot.ai.AiClient` implementation and continue to influence `AiClient#reviewDiff(...)` behavior as they do today.

## Notes from current implementation
- Current source of the values: `org.remus.giteabot.admin.AiIntegration`
- Current consumer path includes:
  - `org.remus.giteabot.prworkflow.review.ReviewWorkflow`
  - `org.remus.giteabot.review.CodeReviewService#reviewPullRequest(...)`
  - `org.remus.giteabot.ai.AbstractAiClient` and provider-specific AI client implementations
- Current `AiIntegration` defaults:
  - `maxDiffCharsPerChunk = 120000`
  - `maxDiffChunks = 8`
  - `retryTruncatedChunkChars = 60000`

## Acceptance criteria
- [ ] The AI integration edit page no longer shows inputs for `maxTokens`, `maxDiffCharsPerChunk`, `maxDiffChunks`, or `retryTruncatedChunkChars`.
- [ ] The workflow edit page shows these 3 inputs in the `review` workflow section.
- [ ] The `review` workflow declares these 3 settings as workflow params using `WorkflowParamField`.
- [ ] New or unset workflow configurations use the same default values currently defined on `AiIntegration`.
- [ ] `AiIntegration` no longer contains these 4 JPA-mapped properties.
- [ ] Flyway migrations exist for both supported database dialects to remove the corresponding `ai_integrations` columns.
- [ ] No data migration is performed from `ai_integrations` into workflow configuration records.
- [ ] Review execution still uses these values when invoking AI review logic.
- [ ] `CodeReviewService#reviewPullRequest(...)` signature and behavior contract remain unchanged.

## Implementation hints
- Check how other workflows declare typed params with `WorkflowParamField` and persist them through workflow configuration.
- The review workflow likely needs to resolve the 4 values from workflow params before constructing or invoking the existing review path.
- Because `CodeReviewService#reviewPullRequest(...)` must stay unchanged, any adaptation likely needs to happen before that call, or inside object construction/factory wiring rather than by changing the service method signature.

## Out of scope
- Migrating existing per-integration values into workflow configuration
- Changing the behavior or signature of `CodeReviewService#reviewPullRequest(...)`

## Target branch
`develop`

## Assumptions

- The existing workflow configuration system already supports integer workflow params for the review workflow UI and persistence.
- Schema migrations are required for both H2 and PostgreSQL, consistent with the repository's existing Flyway structure.
- Resetting existing custom AI-integration values back to workflow defaults is acceptable because the original issue explicitly says no data should be transferred.

## Open questions

- Should existing bots/workflow configurations receive explicit persisted review param values during default configuration creation, or is relying on `WorkflowParamField.defaultValue` sufficient?

---

tmseidel:


max-california (commented):
## 🤖 Agentic Code Review

## Summary

This refactor mostly makes sense: moving review chunking from `AiIntegration` to workflow-level config is a better fit, and the `describeParams()` change improves the Details modal by showing defaults rather than only persisted keys. I didn’t find any obvious runtime breakage in the main review path.

I do see one **blocking migration/data-loss issue** and one **maintainability/test-gap concern** worth addressing before merging.

## Blocking

### 1. Existing per-integration review chunking settings are silently discarded with no migration path
**Files:**  
- `src/main/resources/db/migration/h2/V28__remove_code_review_settings_from_ai_integrations.sql`  
- `src/main/resources/db/migration/postgresql/V28__remove_code_review_settings_from_ai_integrations.sql`

The migration explicitly drops:

- `max_diff_chars_per_chunk`
- `max_diff_chunks`
- `retry_truncated_chunk_chars`

and the comment says:

> No data migration — operators reconfigure these in the workflow UI.

That’s a breaking behavioral change for existing installations. Today those values are already configured per AI integration and actively used by review execution; after this migration they are irretrievably lost and all bots fall back to `review.chunking.*` defaults unless an operator manually re-enters settings in workflow config.

This is more than a cosmetic admin change: it can change review behavior immediately after deploy for all existing bots, especially where operators had tuned chunking for smaller models.

**Why this matters**
- Existing production configuration is deleted during schema migration.
- There’s no automatic carry-forward to workflow selections.
- The PR removes the old UI fields at the same time, so operators have no advance warning in-product.

**Suggestion**
At minimum, avoid destructive loss on upgrade. Options:
1. **Best:** add a data migration that copies old values into the selected `review` workflow params for each workflow configuration/bot that uses review.
2. **If that’s not feasible in this PR:** postpone dropping the columns and keep fallback logic reading `AiIntegration` values until a follow-up migration/UI-assisted transition is complete.
3. Document this clearly in release notes if you intentionally accept the breaking change, but I’d still consider it too risky for an automatic Flyway migration.

## Non-blocking

### 2. No validation prevents invalid/unsafe integer workflow params for review chunking
**Files:**  
- `src/main/java/org/remus/giteabot/prworkflow/review/ReviewWorkflow.java`  
- `src/main/java/org/remus/giteabot/config/ReviewChunkingProperties.java`  
- `src/main/java/org/remus/giteabot/prworkflow/config/WorkflowParamsValidator.java`

`ReviewWorkflow.paramsSchema()` exposes the three chunking values as plain `INTEGER` params, and `intParam(...)` accepts any parsed integer. There’s no lower-bound validation on either:
- Spring properties (`ReviewChunkingProperties`)
- workflow params persisted via `WorkflowParamsValidator`

That means values like `0` or negative numbers can be saved and then passed into `CodeReviewService`. Depending on the chunking implementation there, this can lead to bad behavior or exceptions when splitting/truncating diffs.

The old HTML form for `AiIntegration` had `min="1"` on these fields; that guard disappears with the move to workflow params unless the dynamic workflow form enforces equivalent constraints.

**Suggestion**
- Add bean validation on `ReviewChunkingProperties` (`@Min(1)` at least).
- Add schema-level numeric constraints or explicit checks in `ReviewWorkflow.run()` before constructing `CodeReviewService`.
- If the workflow-params framework doesn’t support min/max yet, clamp/reject invalid values in `intParam(...)`.

## Nits / follow-up suggestions

### 3. `describeParams()` intentionally drops unknown persisted keys for registered workflows
**Files:**  
- `src/main/java/org/remus/giteabot/prworkflow/config/WorkflowParamsValidator.java`

This may be acceptable, but it’s a behavior change from the old `maskSecrets()` implementation, which started from `typed(raw, schema)` and therefore preserved unknown keys. The new `describeParams()` iterates only schema fields, so unknown persisted params disappear from the Details modal for registered workflows.

That’s probably fine for cleanliness, but if unknown keys are useful during downgrade/debug scenarios, this should be a deliberate choice and maybe called out in Javadoc.

### 4. Good change: `describeParams()` showing defaults is an improvement
**Files:**  
- `src/main/java/org/remus/giteabot/prworkflow/config/WorkflowSelectionService.java`  
- `src/main/java/org/remus/giteabot/prworkflow/config/WorkflowParamsValidator.java`  
- `src/main/java/org/remus/giteabot/prworkflow/config/WorkflowConfigurationController.java`

This looks like a solid UX improvement. The Details modal will now show the full schema with defaults instead of only persisted values, while still masking secrets. No issue here; just worth keeping.

## Overall

Good direction overall, but I’d block on the destructive migration as written. The move from AI-integration settings to workflow settings is reasonable, but upgrades shouldn’t silently erase existing tuned values.

---

PR #228, opened 2026-06-20
release/1.13.0

tmseidel:


max-california (commented):
## 🤖 Agentic Code Review

## Summary

This PR is mostly documentation/versioning, and the new docs line up well with the code that’s already in the tree: the new review workflow params exist, `V28` does drop the three `ai_integrations` columns, and the agent loop/context-window docs broadly match the current implementation.

I found one material documentation mismatch that could mislead operators tuning the agent, plus one smaller correctness gap in the migration guide wording.

## Findings

### Medium

- **`doc/AGENT.md`: documented defaults do not match the code for multiple budget settings**
  
  The updated table now says:
  - `AGENT_BUDGET_MAX_ROUNDS` = `20`
  - `AGENT_BUDGET_MAX_CONTENT_ROUNDS` = `10`
  - `AGENT_BUDGET_MAX_CONTEXT_TOOL_REQUESTS_PER_ROUND` = `10`
  - `AGENT_BUDGET_MAX_VALIDATION_RETRIES` = `10`
  - `AGENT_BUDGET_MAX_HISTORY_CHARS` = `180000`

  But `src/main/java/org/remus/giteabot/config/AgentConfigProperties.java` still defines:
  - `maxRounds = 10`
  - `maxContextRounds = 3` (property name is `max-context-rounds`; the env var label in the doc says `MAX_CONTENT_ROUNDS`, likely intended but easy to confuse)
  - `maxContextToolRequestsPerRound = 5`
  - `maxValidationRetries = 3`
  - `maxHistoryChars = 120_000`
  - `maxToolResultChars = 8_000` does match

  This is more than a stale comment issue: operators relying on the docs will configure or capacity-plan for behavior the application does not actually have. Either update the docs to reflect the current code defaults, or include the corresponding code/config change in this release.

### Low

- **`doc/MIGRATION_1.12_TO_1.13.md`: “your data is safe” is too strong given `V28` permanently drops non-default values**
  
  The guide says:
  > “Your data is safe — the columns are simply gone after migration.”

  But the same document correctly states that custom per-integration values are **not migrated** and are discarded when `V28` runs. For operators who had non-default chunking values, that’s data loss unless they inventory/re-enter those values first.

  Suggest rephrasing this section to make the risk explicit, e.g. “schema migration is safe for default behavior, but custom values in those columns are permanently removed unless recorded beforehand.”

## Notes

- I verified that the new PR Review docs match the implementation:
  - `ReviewWorkflow.paramsSchema()` exposes the three workflow params.
  - `ReviewChunkingProperties` provides the documented fallback defaults.
  - `V28` exists for both PostgreSQL and H2 and drops the documented columns.
- The new context-window-management doc is consistent with the presence of `ToolResultTruncator`, `HistoryCompactor`, `TokenUsageTracker`, and the batching/flush logic in `AgentLoop`.

Overall: good release documentation addition, but I’d fix the `AGENT.md` default table before shipping.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

Looks good overall: the docs and version bump line up with the codebase as checked out, especially around the new review-workflow chunking configuration and the added agent context-budget properties.

I found one documentation inconsistency that is likely to mislead operators upgrading from 1.12.x.

## Findings

### Medium

- **`doc/MIGRATION_1.12_TO_1.13.md` overstates rollback failure risk for 1.12.x**
  
  In `doc/MIGRATION_1.12_TO_1.13.md`, the rollback section says that after `V28`, the 1.12.x application “still maps” the removed `ai_integrations` columns via JPA and “may fail at startup or on first database access.”

  That does not match the current code shape around this change:
  - `src/main/java/org/remus/giteabot/admin/AiIntegration.java` no longer maps `max_diff_chars_per_chunk`, `max_diff_chunks`, or `retry_truncated_chunk_chars`.
  - Review chunking is now sourced from `review.chunking.*` and workflow params via `ReviewChunkingProperties`, `ReviewWorkflow`, and `CodeReviewServiceFactory`.
  - The `V28` migration only drops those columns.

  If this release branch is 1.13.0, then the guide should be explicit that the rollback warning applies only when reverting to an **older binary that still expected those columns**. As written, it reads like a statement verified against the current neighboring versions, but the repository no longer contains that older mapping.

  **Suggestion:** reword the rollback note to something like:
  > “If you roll back to a pre-1.13 binary that still expects these columns, it may fail until the columns are restored from backup.”

  That keeps the caution, but avoids implying a fact about code that is no longer present in this branch.

### Nit

- **`docker-compose.yml` defaults differ from `doc/AGENT.md`**
  
  `doc/AGENT.md` now documents defaults of:
  - `AGENT_BUDGET_MAX_CONTEXT_TOOL_REQUESTS_PER_ROUND = 10`

  But `docker-compose.yml` sets:
  - `AGENT_BUDGET_MAX_CONTEXT_TOOL_REQUESTS_PER_ROUND: 20`

  This may be intentional as a sample override, but if so it would be worth commenting clearly in the compose file to avoid confusion between “application default” and “example deployment override.” The same applies to `AGENT_BUDGET_MAX_CONTENT_ROUNDS`, which is `20` in compose while `application.properties` defaults it to `10`.

## Overall

Aside from the migration-guide wording issue, the release documentation appears consistent with the implementation:
- `application.properties` contains the documented new agent budget properties and defaults.
- `ReviewWorkflow`/`ReviewChunkingProperties` match the new PR Review docs.
- `V28` and the migration guide are aligned on the schema change itself.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

This looks good overall. I checked the surrounding code and the documentation changes are aligned with the actual implementation defaults and migrations I could find (`review.chunking.*`, Flyway `V28`, and the updated `agent.budget.*` defaults). I did not find a blocking correctness issue in the release docs themselves.

## Findings

### Medium

- **`doc/PR_WORKFLOWS_REVIEW.md`: “reviewed sequentially” is inaccurate for the `botCommand` action**
  
  In the new workflow doc, step 3 says “For the `review` and `botCommand` actions” and then describes diff fetching/splitting/chunk review behavior. From the code in `src/main/java/org/remus/giteabot/prworkflow/review/ReviewWorkflow.java`, both actions receive the same chunking parameters, but `botCommand` is dispatched to a different service path and does not obviously imply the same “fetches the diff and splits it into chunks” flow.
  
  This reads as if `botCommand` always performs diff review/chunk iteration like a normal PR review, which is broader than what the code guarantees.
  
  **Suggestion:** narrow the wording in `doc/PR_WORKFLOWS_REVIEW.md` so the chunking flow is described specifically for `review`, and separately say that conversational handlers reuse the same `CodeReviewService` configuration where relevant.

### Low

- **`doc/MIGRATION_1.12_TO_1.13.md`: rollback guidance is a bit too absolute**
  
  The rollback section says operators should “Restore the columns from a database backup before rolling back the binary.” That is safe advice, but it may overstate the only recovery path. Since the migration doc already shows the exact removed columns and types/defaults are known elsewhere, operators could also recreate the columns manually if needed.
  
  Not wrong, but worth softening if you want the guide to be operationally precise.
  
  **Suggestion:** change “Restore the columns from a database backup before rolling back the binary” to something like “restore or recreate the columns before rolling back the binary.”

### Nit

- **`doc/development-archive/agent-session-state-cleanup.md` is a large implementation plan in a release PR**
  
  Not a correctness problem, but this file reads like active design work rather than archived historical documentation. Since it’s under `development-archive/`, that may be intentional; still, it’s worth ensuring this is meant to ship as user-visible repository documentation and not just internal scratch material.

## Positives

- `doc/AGENT.md` matches the actual defaults in `src/main/resources/application.properties` for:
  - `agent.budget.max-rounds = 20`
  - `agent.budget.max-context-rounds = 10`
  - `agent.budget.max-context-tool-requests-per-round = 10`
  - `agent.budget.max-validation-retries = 10`
  - `agent.budget.max-history-chars = 180000`
  - `agent.budget.max-tool-result-chars = 8000`

- The new migration guide is consistent with:
  - `ReviewChunkingProperties`
  - `review.chunking.*` application properties
  - Flyway migration `V28`

- The new `PR_WORKFLOWS_REVIEW.md` fills a real documentation gap and is linked from the overview docs appropriately.

---

max-california (commented):
## 🤖 AI Code Review

### Diff chunk 1/3
1. **Blocking issues**

- **`contextWindowTokens` was added to `AiIntegration` without a shown DB migration.**  
  In this chunk, `AiIntegration` now has a new non-null JPA column:
  ```java
  @Column(nullable = false)
  private int contextWindowTokens = 200_000;
  ```
  but the listed migration docs only mention V28 dropping old review-chunking columns. If there is no matching Flyway migration in the remaining chunks, startup will fail against existing DBs or schema validation will drift.  
  **Fix:** add H2/PostgreSQL migrations for `context_window_tokens` with a default value and backfill existing rows.

- **`callAiWithRetry()` only catches `HttpClientErrorException`, which is likely too narrow for provider “prompt too long” failures.**  
  The code relies on `aiClient.isPromptTooLongError(e)` but only for `HttpClientErrorException`. Depending on client/provider, oversized prompts may surface as other `RestClientResponseException` subclasses, server-side exceptions, wrapped runtime exceptions, or provider-specific exceptions. The new reactive compaction path will silently not run in those cases.  
  **Fix:** catch a broader exception type aligned with how all `AiClient` implementations fail, or normalize prompt-too-long detection inside the `AiClient` abstraction so the loop doesn’t depend on Spring exception class details.

- **Prompt-size estimation for proactive compaction is materially wrong because it ignores tool-call payloads and tool metadata.**  
  `TokenUsageTracker.record(..., promptChars)` is passed only `currentMessage.length()`, and later the proactive recheck sums only:
  - `history` message `content`
  - `systemPrompt`
  - `currentMessage`
  
  That omits tool call arguments, `toolResult` lengths in the estimator path, and native tool descriptors sent with the request. This can substantially underestimate prompt size, especially in tool-heavy rounds, making compaction too late or ineffective.  
  **Fix:** base compaction on actual provider-reported input tokens when available; otherwise estimate from the full serialized request shape (history content + tool results + tool calls + tool descriptors + system prompt + current message).

- **`docker-compose.yml` defaults drift from docs/code defaults.**  
  Docs now say `AGENT_BUDGET_MAX_CONTENT_ROUNDS=10` and `...MAX_CONTEXT_TOOL_REQUESTS_PER_ROUND=10`, but compose hardcodes both to `20`. That will confuse operators and make behavior differ between local Docker and property defaults. Consider aligning compose with application defaults, or documenting that compose intentionally overrides them.

- **The new native-mode completion behavior in `CodingAgentStrategy` deserves an explicit safeguard/test around empty/no-op runs.**  
  Finishing successfully when there are uncommitted changes is sensible, but plain-text completion without tool calls now becomes “continue” forever until budget exhaustion when no changes exist. That may be intended, but it’s worth verifying models that answer conversationally in native mode don’t degrade into repeated “missing tool” nudges.

- **`HistoryCompactor` summary role being `"user"` is a bit odd semantically.**  
  It probably works, but a synthetic user message may bias subsequent model behavior more than a system/assistant-style summary would. If your message model supports it, consider making compaction summaries a dedicated assistant/narrator-style message.

3. **Tests**

Recommended/additional coverage for the changed behavior:

- **Migration test** for the new `AiIntegration.contextWindowTokens` column and existing-row compatibility.
- **AgentLoop reactive compaction test**:
  - first AI call throws “prompt too long”
  - history is aggressively compacted
  - second call is retried
  - failure is rethrown if retry still fails
- **Proactive compaction/truncation tests** covering:
  - tool message truncation in-memory
  - persisted tool-message truncation via `truncateToolMessages`
  - fallback to full compaction when truncation is insufficient
- **`CodingAgentStrategy` tests** for native mode:
  - plain-text completion + workspace changes => successful finish with plan payload
  - plain-text completion + no changes => continue with missing-tool feedback
  - legacy mode plain-text response still fails as before
- **`IssueImplementationService.handleIssueComment` regression test** confirming:
  - rebinding after `compactContextWindow`
  - PR creation uses the locally tracked `prNumber`
  - success comment uses `implementationResult.plan()` rather than stale session state

4. **Overall assessment**

Promising refactor overall, especially the move toward workflow orchestration and explicit context-window management, but I’d block on the missing schema migration for `contextWindowTokens` and the too-narrow exception handling in the new retry path.

### Diff chunk 2/3
1. **Blocking issues**

- **`OllamaClient.isPromptTooLongError()` now NPEs on null response bodies**
  - In this diff, the previous null check was removed:
    ```java
    ```
    If the provider returns a 400 with no body, this throws `NullPointerException` and masks the original error. That breaks both review retry behavior and agent compaction/retry handling.
  - **Fix:** restore the null guard, or delegate to `AiClient.super.isPromptTooLongError(e)` when `body == null`.

- **`CodeReviewService.handleReviewSubmitted()` can now throw NPE when there are no reviews**
  - This changed from:
    ```java
    if (latestReview == null || latestReview.getId() == null)
    ```
    to:
    ```java
    if (latestReview.getId() == null)
    ```
    But `orElse(null)` is still used immediately above, so `latestReview` can absolutely be `null`.
  - **Fix:** put the null check back.

- **`compactContextWindow()` adds the summary with a fresh timestamp, so it won’t be “the first message” in replay order**
  - The code comment says:
    > Add a summary message as the first message in the surviving history
  - But `managed.addMessage("user", summary)` uses a new timestamp “now”, while the kept messages are older. Since replay is sorted by `createdAt`, the summary will likely appear last, not first. That changes the intended semantics of compaction and may confuse the model by showing summary *after* the surviving context.
  - **Fix:** insert the summary with a timestamp earlier than the oldest retained message (or rebuild the retained message set with explicit timestamps).

- **Persisted compaction currently breaks tool-call/tool-result pairing**
  - `compactContextWindow()` removes the oldest raw messages by count, while `toAiMessages` later drops tool messages/blank assistant messages. That means persisted history may end up keeping orphaned assistant tool-call messages or half of a logical interaction. Your architecture doc explicitly calls out preserving logical units, so this implementation seems weaker than intended.
  - This may be acceptable for coarse persisted follow-up context, but it’s worth verifying. A better approach would compact on logical units, not raw messages.

- **Summary message role in `compactContextWindow()` may be suboptimal**
  - The compaction summary is added as `"user"`. That means the model sees an extra user instruction rather than system/session metadata. If downstream prompts treat user messages with higher priority, this may distort behavior.
  - Consider either:
    - using an assistant-style summary if that fits current replay behavior better, or
    - introducing a dedicated session-summary handling path if the message model supports it.

- **`ReviewWorkflow` now constructs a `CodeReviewService` for actions that don’t need diff chunking**
  - `doBotCommand`, `doInlineComment`, `doReviewSubmitted`, and `doPrClosed` all thread chunking params through even though only diff review uses them. Not wrong, just a sign the service now mixes concerns.
  - Consider separating review-diff behavior from conversational/session handlers later to reduce coupling.

- **`ReviewChunkingProperties` lacks validation**
  - Since these values can now come from app properties and workflow params, bad values like `0` or negative numbers could lead to broken chunking behavior (`substring(0, 0)`, no progress loops, etc.) depending on call paths.
  - Suggest adding bean validation (`@Min(1)`) and validating workflow param schema as positive integers as well.

3. **Tests**

- Add a regression test for **`OllamaClient.isPromptTooLongError()`** with a `HttpClientErrorException` that has a null/empty body.
- Add a regression test for **`CodeReviewService.handleReviewSubmitted()`** when no valid latest review exists.
- Add tests for **persisted session compaction ordering**, specifically that the summary appears before retained messages in replay order.
- Add tests for **review chunking via workflow params/fallbacks**:
  - workflow param overrides app defaults,
  - app defaults are used when no workflow config exists,
  - retry-on-prompt-too-long uses `retryTruncatedChunkChars`.
- Add tests for **`WorkflowParamsValidator.describeParams()`** to confirm:
  - defaults are shown for absent fields,
  - secrets remain masked,
  - non-schema raw fields behavior is intentional.

4. **Overall assessment**

Good direction overall: moving diff chunking out of `AiIntegration`, adding token-aware chat metadata, and reducing detached-entity merge problems all look valuable. However, the two regressions (`latestReview` null handling and the Ollama null-body NPE) plus the compaction ordering bug should be fixed before merge.

### Diff chunk 3/3
1. **Blocking issues**

- **Migration numbering / release compatibility looks risky**
  - In this chunk, `src/main/resources/db/migration/postgresql/V28__remove_code_review_settings_from_ai_integrations.sql` is added, but the diff also shows raw SQL adding:
    - `agent_sessions.total_output_tokens`
    - `ai_integrations.context_window_tokens`
  - I don’t see the corresponding new/versioned Flyway migration file in this chunk for those schema changes, and `V28` is already being used for a different purpose. If those new columns were introduced by modifying an existing migration or via an unversioned file, upgrades from existing installations will break or drift:
    - existing DBs won’t get the new columns reliably
    - changing an old applied migration is unsafe for Flyway-managed environments
  - **Fix:** ensure both H2 and PostgreSQL have new forward-only Flyway migrations for the added columns, with unique version numbers greater than existing ones. Do not modify previously released migration files.

- **Documented migration behavior is destructive for non-default review settings**
  - `V28` drops `max_diff_chars_per_chunk`, `max_diff_chunks`, and `retry_truncated_chunk_chars` with **no data migration**, while the migration guide says custom values must be manually re-entered later.
  - That is a backwards-compatibility/regression risk for upgraded production instances: existing customized integrations silently lose effective behavior on first boot.
  - If this is intentional, it’s still severe enough that I’d block unless there is either:
    - an automatic migration into workflow config, or
    - a startup warning / pre-migration check surfacing affected rows before deletion.
  - **Fix:** preferably migrate existing non-default values into workflow config automatically. At minimum, add runtime/admin warnings before/after migration and explicit release-note prominence.

- **Add validation around `contextWindowTokens`**
  - The UI uses `min="1"`, but server-side validation is what matters. If the entity/controller/service doesn’t enforce positivity, bad persisted values could disable compaction logic or create confusing runtime behavior.
  - **Suggestion:** add bean validation like `@Min(1)` on the field and a corresponding test.

- **`AuditingAiClientTest` expectation changed semantically**
  - The old test name/comment said review errors should not be double-audited; now the assertion expects the error to be recorded.
  - That may be correct after the refactor, but the current test name/comment mismatch makes the intended behavior unclear.
  - **Suggestion:** rename the test/comment to match the new contract, and verify there isn’t duplicate recording across `submitReviewPrompt` and underlying client error paths.

- **Brittleness in tests asserting exact `PendingMessage` batches**
  - Several updated tests verify exact `List<PendingMessage>` contents/order in `flushMessages(...)`. That’s useful, but also tightly couples tests to batching internals.
  - **Suggestion:** keep a few precise contract tests, but in higher-level service tests prefer asserting key persisted effects and ordering rather than the full exact batch shape.

3. **Tests**

- **Missing migration tests**
  - Recommend tests covering upgrade from a pre-1.13 schema for both H2 and PostgreSQL migrations, especially:
    - old DB with existing `ai_integrations` chunking columns
    - presence of new `context_window_tokens`
    - presence of `agent_sessions.total_output_tokens`
    - successful startup after migration

- **Missing review-workflow config precedence tests**
  - Since diff chunking moved from entity-level config to workflow config, add tests for:
    - workflow param override used when present
    - `review.chunking.*` properties used as fallback
    - legacy `AiIntegration` values no longer consulted

- **Missing persistence/validation tests for `contextWindowTokens`**
  - Add tests that new AI integrations default correctly, save/load correctly, and reject invalid values if server-side validation is added.

4. **Overall assessment**

Good progress on tests around agent session compaction/token tracking and the review refactor, but I would not merge yet due to the schema/migration concerns. The migration path needs to be made clearly safe and forward-only for existing deployments.

---

max-california (commented):
## 🤖 AI Code Review

1. **Blocking issues**

- None in the provided updated diff. The changes shown are documentation/versioning only, and I don’t see a correctness or security problem in the files included here.

- **`doc/CONTEXT_WINDOW_MANAGEMENT.md` appears inconsistent with current documented defaults.**  
  In `doc/AGENT.md`, `AGENT_BUDGET_MAX_HISTORY_CHARS` is documented as `180000`, but `doc/CONTEXT_WINDOW_MANAGEMENT.md` says:
  ```md
  | `max_history_chars` | 120,000 | Soft limit for total history |
  ```
  If 180k is the actual current default, this architecture doc is already stale and could mislead operators/tuners. I’d align the numbers or explicitly label the architecture doc values as illustrative.

- **Migration guide is clear, but the destructive nature still deserves strong release-note visibility.**  
  `doc/MIGRATION_1.12_TO_1.13.md` correctly states that custom `ai_integrations` diff-chunking values are discarded and must be re-entered. That’s good documentation, but this is still a high-impact operational change. If not already planned outside this diff, I’d make sure the release notes/changelog prominently call this out.

- **Development archive doc could confuse readers if indexed/discoverable as current design.**  
  `doc/development-archive/agent-session-state-cleanup.md` is useful, but it contains “Proposed” ADRs and implementation plans that may no longer match the merged code. Consider adding a stronger banner at the top like “historical design notes; may not reflect current implementation” if this folder is linked or searchable in published docs.

- **`PR_WORKFLOWS_REVIEW.md` says “for the `review` and `botCommand` actions” but the later list includes `inlineComment`.**  
  In the flow section:
  - step 3 says “For the `review` and `botCommand` actions…”
  - step 4 then says “For `botCommand` / `inlineComment`…”
  
  It reads a little inconsistently. Consider restructuring that section by action to avoid ambiguity about which handlers use diff chunking versus conversational handling.

3. **Tests**

- No new code paths are shown here, so no new test requirements are evident from this diff alone.
- If the underlying PR still includes the schema/runtime changes from earlier chunks, my previous recommendations about migration and compaction/retry regression tests still apply; this update itself doesn’t address or invalidate those concerns.

4. **Overall assessment**

This updated diff looks fine as documentation/version metadata. I don’t see new blocking problems in the files shown. If the previously identified code/migration issues are fixed elsewhere in the updated PR, this documentation update is a good improvement.

---

PR #232, opened 2026-06-25
fix: prevent empty reviews on output token limit exceeded

rmie:
# Prevent Empty Reviews on Output Token Limit Exceeded

## 📋 Goal
Ensure that the AI bot always produces helpful feedback rather than posting an empty review comment or failing silently if the output token limit is exceeded or the review response is empty.

---

### 1. AI Client Safeguards
Added checks to intercept empty or blank text responses from AI provider APIs when no function/tool calls are made. Instead of returning an empty string, they now log a warning and return a clear warning message:
* **Files Modified:**
  * [AnthropicAiClient.java](file:///D:/src/github/rmie/ai-git-bot/src/main/java/org/remus/giteabot/ai/anthropic/AnthropicAiClient.java)
  * [OllamaClient.java](file:///D:/src/github/rmie/ai-git-bot/src/main/java/org/remus/giteabot/ai/ollama/OllamaClient.java)
  * [LlamaCppClient.java](file:///D:/src/github/rmie/ai-git-bot/src/main/java/org/remus/giteabot/ai/llamacpp/LlamaCppClient.java)

### 2. CodeReviewService Guard
Standardized validation for both chat responses and pull request reviews in [CodeReviewService.java](file:///D:/src/github/rmie/ai-git-bot/src/main/java/org/remus/giteabot/review/CodeReviewService.java).
* If a response is null or blank, it fallbacks to a user-friendly warning message (`⚠️ *The review feedback exceeded the configured token limit and could not be fully generated.*`) instead of posting a blank comment.

### 3. Review Agent Strategy Robustness
Improved fallback outcomes in [ReviewAgentStrategy.java](file:///D:/src/github/rmie/ai-git-bot/src/main/java/org/remus/giteabot/prworkflow/agentreview/ReviewAgentStrategy.java) when the round budget is exhausted or a step finishes with blank text. It will now yield a Markdown warning comment rather than a loop failure.

### 4. Unit Testing
* Added tests in [CodeReviewServiceTest.java](file:///D:/src/github/rmie/ai-git-bot/src/test/java/org/remus/giteabot/review/CodeReviewServiceTest.java) covering empty reviews and response warning fallbacks across `reviewPullRequest`, `handleBotCommand`, and `handleInlineComment`.
* Added tests in [ReviewAgentStrategyLegacyTest.java](file:///D:/src/github/rmie/ai-git-bot/src/test/java/org/remus/giteabot/prworkflow/agentreview/ReviewAgentStrategyLegacyTest.java) to verify budget exhaustion and finish fallback warnings.

---

## 🧪 Verification
* All **963 unit and integration tests** built and passed successfully inside the `maven:3.9-eclipse-temurin-21-noble` Docker environment.


tmseidel:
@rmie Thanks for your contribution, Here are some remarks that must be addressed before we can merge this:

The goal is correct — empty AI responses should never be posted as PR comments. The fix is in the right places (service layer + agent strategy) but also adds redundant,
wrong-tier changes to the AI client layer that should be reverted. There's also a subtle ordering bug in reviewPullRequest.
CRITICAL: AI client changes should be reverted entirely
Wrong architectural tier. Protocol adapters (*AiClient) translate between Java and HTTP APIs. They should return whatever the upstream API returns, including empty strings. Producing user-facing markdown warnings like "Unable to generate review - empty response from AI" belongs in the service/strategy layer, which owns the user-facing contract. The client layer doesn't even know what markdown formatting the project uses.

MODERATE: Blank response stored in session before guard in reviewPullRequest
 CodeReviewService.java lines 101-102 and 116-117:
 ```java
 sessionService.addMessage(session, "assistant", review);  // stores blank if API returned ""
 // ... later ...
 if (review == null || review.strip().isEmpty()) {
     review = "⚠️ *...";  // corrects the variable but the session already has the blank
 }
 ```
  The session stores the blank response before the fallback is applied. This means:
 - The PR gets the warning message (correct)
 - The session stores the blank message (incorrect)
 - Future conversation turns will see an empty assistant message in history
 - If the PR is updated and a new review runs, the AI sees a blank assistant turn — confusing
 Compare with handleBotCommand and buildInlineCommentAndSend, where the guard is correctly placed before sessionService.addMessage.
 Fix: Move the empty-check above the addMessage calls, or re-store the corrected message:
```java
 if (review == null || review.strip().isEmpty()) {
     review = "⚠️ The review feedback was empty or could not be generated.";
 }
 sessionService.addMessage(session, "user", ...);
 sessionService.addMessage(session, "assistant", review);
 ```
 The test at line 524 should also be updated — eq("") should become contains("was empty").
 
 Summary of required changes PART 1:
 | File                       | Action                                                               |
 |----------------------------|----------------------------------------------------------------------|
 | AnthropicAiClient.java     | Revert all changes                                                   |
 | OllamaClient.java          | Revert all changes                                                   |
 | LlamaCppClient.java        | Revert all changes                                                   |
 | CodeReviewService.java     | Move the empty-check in reviewPullRequest above the addMessage calls |
 | CodeReviewServiceTest.java | Update eq("") on line 524 to match the corrected session storage     |  
 
 Regarding the AgentLoop:
 You changed the semantic of the loop-run in the following way:
 
 
| onBudgetExhausted returns      | outcome.success() | review set to        | Result                                |                                                        
|--------------------------------|-------------------|----------------------|---------------------------------------|                                                        
| Original: LoopOutcome.fail()   | false             | null → returns false | Silent failure — no PR comment at all |                                                        
| New: LoopOutcome.success(text) | true              | warning string       | Warning posted as PR comment          |

Changing fail to success is wrong. Budget exhaustion IS a failure — the loop couldn't finish its job. Disguising it as success with a warning string       
in the payload is semantically incorrect.

The real problem is that LoopOutcome.fail() has no payload — it throws away the error context:                                                                               
                                                                                                                                                                          
```java                                                                                                                                                                         
public static LoopOutcome fail(String selectedBranch) {                                                                                                                      
 return new LoopOutcome(false, selectedBranch, null);  // ← payload is always null                                                                                        
}                                                                                                                                                                            
```                                                                                                                                                                         

The AgentReviewService consumer can't extract an error message from a failed outcome, so it just logs and returns silently. Even if the strategy wanted to say why it
failed (token limit, budget exhaustion, etc.), there's no channel for that information. 

The fix for that is the following:
- LoopOutcome.fail(branch, errorMsg) — overload with payload
- AgentReviewService extracts the message from a failed outcome and posts it
- ReviewAgentStrategy.onBudgetExhausted stays as fail() but carries the warning text

@rmie Still one addition: Please clarify how AI was used to create the Pull-Request. - We are currently investigating how to handle AI-generated Pull-requests and this PR is relevant for this topic.

tmseidel (changes requested):
See my last comment on this review

rmie:
> @rmie Still one addition: Please clarify how AI was used to create the Pull-Request. - We are currently investigating how to handle AI-generated Pull-requests and this PR is relevant for this topic.

Hi @tmseidel,

Sure, happy to clarify! 

I asked gemini about a summery based on its brains, could't have written it better.

Let me know if you need any further details on the process!

---

This Pull Request was created using a collaborative **AI-assisted pair-programming workflow** with an agentic AI coding assistant. Here is how the AI was utilized:

1. **Problem Analysis & Code Location:** After identifying the issue where empty review comments were posted due to token limits or empty API responses, the AI scanned the codebase to locate the relevant services (`CodeReviewService`, `ReviewAgentStrategy`, etc.).
2. **Implementation Drafts:** The AI proposed the logic to gracefully catch empty/null responses and fall back to user-visible warning messages instead of failing silently or posting empty review boxes.
3. **Test Generation:** The AI assisted in writing unit tests (in `ReviewAgentStrategyLegacyTest.java` and `CodeReviewServiceTest.java`) to verify the new fallback behavior.
4. **Human Review & Verification:** I directed the workflow, reviewed all generated code and diffs, ran the build/test suite locally, and finalized the commits. 

In short, the AI acted as a co-pilot to locate the relevant code sections and write the initial implementation/tests, while the human developer retained full control over review, verification, and final integration.

---

tmseidel:
@rmie Ok, fine. Please see my comments. Happy to see the issues I mentioned to be fixed so that we can merge this PR

src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java
```diff
@@ -570,7 +570,7 @@ private List<String> findMatches(Path workspaceDir, Path basePath,
                     .toList();
 
             for (Path file : files) {
-                String relativeFilePath = workspaceDir.relativize(file).toString();
+                String relativeFilePath = workspaceDir.relativize(file).toString().replace('\\', '/');
```
tmseidel: Please revert this change. If you have problems with relative paths, please open a ticket where the behavior is described and why we need to fix it. Since the software is supported only to run on linux, there shouldn't be any relative paths containing back-slashes

tmseidel (changes requested):
Please revert the change in ToolExecutionService - If we have a problem here, a separate issue and PR is required, the scope is important. All other changes look good to me 🚀

src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java
rmie: sorry for that, claude went wild while trying to fix test issues on my windows machine and I didn't catch it.

tmseidel: @rmie there is still a new empty-line in ToolSelectionService. Please revert ALL changes ☺️

max-california (commented):
## 🤖 AI Code Review

1. **Blocking issues**
- None identified in the diff provided.

- In `CodeReviewService.reviewPullRequest`, the new null/blank fallback is applied *after* an existing debug log that still does `review.length()` / `review.substring(...)` in the new-PR path. If `aiClient.submitReviewPrompt(...)` can return `null` there, this will now still throw before the fallback is applied. The update-path logging already handles null safely, so it would be better to normalize `review` before any logging, or make that first log null-safe as well.
- The warning strings are duplicated across multiple classes. Consider centralizing them as constants to avoid drift and make future wording changes easier.
- `AgentReviewService.reviewPullRequest` now posts the review comment even when `outcome.success()` is false, as long as payload text exists, and then returns `false`. That may be intentional for fallback warnings, but it changes the meaning of the boolean from “comment posted successfully” to “loop ended successfully.” If callers use the return value to decide whether the PR was reviewed, this could cause confusing behavior. Worth verifying the contract and documenting or adjusting it.

3. **Tests**
- Add a test for `CodeReviewService.reviewPullRequest` where the AI client returns `null` rather than `""`, especially for the initial review path. That would catch the likely null-safety issue in the current logging.
- Add a test covering `AgentReviewService.reviewPullRequest` for the case where the loop fails with a non-blank payload, to lock in the intended behavior of “post fallback comment but return failure.”
- If the AI clients were changed as described in the PR text, those client-level null/blank response safeguards should ideally have direct unit tests too; they’re not visible in this diff.

4. **Overall assessment**
- The change is directionally good and improves resilience against empty AI output. I don’t see a merge-blocking problem in the shown diff, but I would strongly recommend fixing or verifying the potential null-handling issue in `CodeReviewService` before merging.

---

max-california:
⚠️ **AI Agent (Review)**: Failed to prepare workspace: Failed to clone repository: Cloning into 'agent-workspace-15695567255036107174'...
warning: Could not find remote branch bugfix/empty-response-ontoken-limit to clone.
fatal: Remote branch bugfix/empty-response-ontoken-limit not found in upstream origin

tmseidel (approved):

src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java
rmie: **Quick post-mortem on the slipped newline**

Clearly, I should have made the revert myself, but instead used an AI agent for it, so I was curious to find out why it slipped. 

Gemini's reply boils down to a conflict between:

- Please revert the change in ToolExecutionService
- the attached diff
- All other changes look good to me

"All" caused the ambiguity in this case. But even a junior developer would have resolved this into "except changes in ToolExecutionService", and reverted the whole file, instead of "All changes in the diff". Gemini edited the file using the diff context and fumbled on the newline.

I'm not sure where to go from here. Do we have to write everything with CARE in the future just in case it gets prompted to an agent?

*Note:* This is an honest attempt to better understand the interaction between humans and AI agents. I'm not trying to shift the blame back to you.

PR #233, opened 2026-06-27
feature/code-base-truncation

### LINKED ISSUE #230: Code base truncation for LLMs with smaller context-window
Truncating codebases for AI context requires careful handling. Code relies on strict logic, so missing elements like closing brackets, variable scopes, or dependent functions can break the AI's understanding.

To safely compress a codebase, focus on structural hierarchy and relevance filtering rather than simple text cutting.

**Top Strategies for Code Base Truncation**
* Use an Abstract Syntax Tree (AST): Parse files into an AST to extract only function signatures, class definitions, and docstrings. This gives the AI a lightweight "map" of the codebase without the bloated implementation details.
* Filter by Directory Structure: Use tools like tree or generate a structural map (components/, utils/, models/) so the AI can navigate the project before asking for specific file contents. (this is already done: AI-Git-Bot provides `rg`, `find`,`tree`)
* Implement Dependency Graphing: Track file imports (import, require, include). Only feed the AI the target file and its direct dependencies, stripping out unrelated modules entirely.Token-Aware Line Skipping: Strip out large comment blocks, verbose logging statements, test suites, and third-party node modules or libraries (node_modules/, vendor/).

**Idea of Implementation**
The industry standard approach for multi-language AI tools is to use Tree-sitter or ctags. Tree-sitter is the tool that powers GitHub's syntax highlighting and code navigation. This is how a universal signature extractor tool in Java using Universal Ctags (or Tree-sitter CLI) via system process calls, which natively supports all your requested languages can be implemented:

* Add our docker image with the native tool `sudo apt install universal-ctags`
* Rough implementation:
```java
import java.io.BufferedReader;
import java.io.File;
import java.io.InputStreamReader;
import java.util.ArrayList;
import java.util.List;

public class MultiLanguageMarkdownTool {

    /**
     * AI Tool Definition: Extrahiert Signaturen und gibt sie als Markdown-String zurück.
     * 
     * @param filePath Der Pfad zur Quellcodedatei.
     * @return Ein formatierter Markdown-String für den AI-Kontext.
     */
    public static String extractSignaturesAsMarkdown(String filePath) {
        File file = new File(filePath);
        if (!file.exists()) {
            return "### Error\nFile not found at `" + filePath + "`";
        }

        StringBuilder markdown = new StringBuilder();
        String fileName = file.getName();
        String langMarker = getMarkdownLanguageMarker(fileName);

        // Header für das LLM
        markdown.append("### Code Structure: `").append(fileName).append("`\n");
        markdown.append("```").append(langMarker).append("\n");

        try {
            // Prozessaufruf für Universal Ctags
            ProcessBuilder pb = new ProcessBuilder(
                "ctags", 
                "--output-format=json", 
                "--fields=+neKz", 
                file.getAbsolutePath()
            );
            
            Process process = pb.start();
            BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
            String line;
            boolean hasContent = false;

            while ((line = reader.readLine()) != null) {
                String formattedLine = parseCtagsLineToMarkdown(line);
                if (formattedLine != null) {
                    markdown.append(formattedLine).append("\n");
                    hasContent = true;
                }
            }

            process.waitFor();
            if (!hasContent) {
                markdown.append("// No classes or methods detected in this file.\n");
            }

        } catch (Exception e) {
            return "### Error\nFailed to parse file: " + e.getMessage();
        }

        markdown.append("```\n");
        return markdown.toString();
    }

    /**
     * Übersetzt das Ctags-JSON in ein lesbares Pseudo-Code-Format für das LLM
     */
    private static String parseCtagsLineToMarkdown(String jsonLine) {
        String name = extractJsonKey(jsonLine, "name");
        String kind = extractJsonKey(jsonLine, "kind");
        String signature = extractJsonKey(jsonLine, "signature");

        if (name == null || kind == null) return null;

        // Saubere Strukturierung für LLM-Aufmerksamkeit
        switch (kind.toLowerCase()) {
            case "c": case "class":     return "class " + name + " {";
            case "i": case "interface": return "interface " + name + " {";
            case "namespace":           return "namespace " + name + " {";
            case "f": case "function":  return "  function " + name + (signature != null ? signature : "()");
            case "m": case "method":    return "  method " + name + (signature != null ? signature : "()");
            case "constructor":         return "  constructor " + name + (signature != null ? signature : "()");
            case "macro":               return "  macro " + name;
            default:                    return null; // Ignoriert Variablen und lokale Scopes für Kontext-Ersparnis
        }
    }

    private static String getMarkdownLanguageMarker(String fileName) {
        String lower = fileName.toLowerCase();
        if (lower.endsWith(".py")) return "python";
        if (lower.endsWith(".ts") || lower.endsWith(".tsx")) return "typescript";
        if (lower.endsWith(".js") || lower.endsWith(".jsx")) return "javascript";
        if (lower.endsWith(".java")) return "java";
        if (lower.endsWith(".cs")) return "csharp";
        if (lower.endsWith(".c") || lower.endsWith(".h")) return "c";
        if (lower.endsWith(".cpp") || lower.endsWith(".hpp")) return "cpp";
        return "text";
    }

    private static String extractJsonKey(String jsonLine, String key) {
        String search = "\"" + key + "\":\"";
        int start = jsonLine.indexOf(search);
        if (start == -1) return null;
        start += search.length();
        int end = jsonLine.indexOf("\"", start);
        if (end == -1) return null;
        return jsonLine.substring(start, end);
    }

    public static void main(String[] args) {
        // Beispielaufruf
        String testFile = "UserService.cs"; 
        System.out.println(extractSignaturesAsMarkdown(testFile));
    }
}
```
The result would look like
```java
class OrderProcessor {
  constructor OrderProcessor(String processorId)
  method processOrder(String orderId, double amount)
  method internalCleanup()
}
```

* Another good way to save context and give a better toolhandling for the llm is to list dependencies, includes and namespaces:
* Rough implementation
```java
import java.io.BufferedReader;
import java.io.File;
import java.io.InputStreamReader;
import java.util.ArrayList;
import java.util.LinkedHashMap;
import java.util.List;
import java.util.Map;

public class UniversalDependencyTool {

    /**
     * AI Tool Definition: Extrahiert Importe, Includes und Namespaces für alle angeforderten Sprachen.
     * 
     * @param filePath Der relative oder absolute Pfad zur Code-Datei.
     * @return Eine strukturierte Map (wird als JSON an die KI zurückgegeben).
     */
    public static Map<String, Object> getDependencies(String filePath) {
        Map<String, Object> result = new LinkedHashMap<>();
        File file = new File(filePath);

        if (!file.exists()) {
            result.put("error", "File not found at " + filePath);
            return result;
        }

        result.put("file", file.getName());
        
        List<String> imports = new ArrayList<>();
        List<String> namespaces = new ArrayList<>();

        try {
            // Wir filtern Ctags so, dass es NUR Imports (i) und Namespaces/Packages (n) sucht.
            // Das spart massiv Zeit und reduziert den Output vorab.
            ProcessBuilder pb = new ProcessBuilder(
                "ctags", 
                "--output-format=json", 
                "--kinds-all=-*",      // Schalte alle Standard-Typen ab
                "--kinds-all=+i+n",    // Schalte NUR Imports (i) und Namespaces (n) an
                "--fields=+k",         // Füge das 'kind'-Feld zum JSON hinzu
                file.getAbsolutePath()
            );
            
            Process process = pb.start();
            BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
            String line;

            while ((line = reader.readLine()) != null) {
                String name = extractJsonKey(line, "name");
                String kind = extractJsonKey(line, "kind");

                if (name == null || kind == null) continue;

                if (kind.equalsIgnoreCase("import") || kind.equalsIgnoreCase("include")) {
                    imports.add(name);
                } else if (kind.equalsIgnoreCase("namespace") || kind.equalsIgnoreCase("package")) {
                    namespaces.add(name);
                }
            }
            process.waitFor();

        } catch (Exception e) {
            result.put("error", "Failed to parse dependencies: " + e.getMessage());
            return result;
        }

        result.put("declared_namespace_or_package", namespaces.isEmpty() ? "none" : namespaces.get(0));
        result.put("dependencies", imports);

        return result;
    }

    private static String extractJsonKey(String jsonLine, String key) {
        String search = "\"" + key + "\":\"";
        int start = jsonLine.indexOf(search);
        if (start == -1) return null;
        start += search.length();
        int end = jsonLine.indexOf("\"", start);
        if (end == -1) return null;
        return jsonLine.substring(start, end);
    }

    public static void main(String[] args) {
        // Beispiel-Testaufrufe für verschiedene Sprachen
        String[] testFiles = {
            "scripts/analytics.py", 
            "src/components/Button.tsx", 
            "main.cpp", 
            "Repository/UserRepository.cs"
        };

        for (String path : testFiles) {
            System.out.println("--- Executing toolcall for: " + path + " ---");
            System.out.println(getDependencies(path));
        }
    }
}
```

The result would look like:
```json
{
  "file": "Button.tsx",
  "declared_namespace_or_package": "none",
  "dependencies": [
    "react",
    "../hooks/useAuth",
    "@mui/material/Button"
  ]
}
```

tmseidel:


max-california (commented):
## 🤖 Agentic Code Review

## Summary

Nice feature overall: the tool registration, routing, Docker dependency, tests, and migration docs are all aligned. I found two correctness issues that look blocking before merge, both in the new ctags execution/parsing path.

## Blocking

### 1. `ctags-deps` is asking ctags for the wrong kinds, so it likely won’t return imports/packages at all
**File:** `src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java`  
**Lines:** around `1149-1154`

The command is built as:

```java
"ctags", "--output-format=json",
"--kinds-all=-*", "--kinds-all=+i+n",
"--fields=+k",
...
```

`--kinds-all` uses ctags’ one-letter kind letters, not the human-readable kind names that your formatter later expects (`import`, `include`, `namespace`, `package`). The implementation plan above also assumes `i` = import and `n` = namespace, but ctags kind letters are language-specific, so `+i+n` is not a portable way to say “imports and namespaces/packages” across languages. In many parsers this will simply select the wrong tag kinds or none at all.

As a result, `formatCtagsDependencies()` can easily receive empty output even for files with obvious imports/packages.

**Suggestion:** verify the actual ctags output format you want to support and either:
- avoid aggressive `--kinds-all` filtering and post-filter JSON by `kind`, or
- use language-specific options if you truly need parser-level filtering.

Right now the formatter logic and the ctags invocation are not reliably compatible.

---

### 2. `ctags-signatures` drops Python methods entirely because it doesn’t handle `member`
**Files:**  
- `src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java`  
- `src/test/java/org/remus/giteabot/agent/validation/ToolExecutionServiceCtagsTest.java`

Your own test fixture for Python uses:

```json
{"kind":"member", "name":"__init__", ...}
{"kind":"member", "name":"process_order", ...}
```

But `renderCtagsTag()` only handles:

- class / interface / namespace / module
- function
- constructor
- method
- macro

Anything else returns `null`. So with that fixture, the Python output will only contain the class line and omit all methods. The current test doesn’t catch that because it only asserts the language marker and filename.

That makes the tool materially less useful for Python and any other language where ctags emits `member` rather than `method`.

**Suggestion:** either:
- treat scoped `member` entries as methods when appropriate, or
- broaden the supported kinds based on actual ctags output for target languages.

Also strengthen the test to assert the rendered Python members, not just the fence header.

## Non-blocking

### 3. The Markdown header formatting is inconsistent between empty and non-empty signature output
**File:** `src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java`  
**Lines:** around `1012-1030`

For empty output you return:

```java
"### `" + new File(filePath).getName() + "`\n```text\n..."
```

But for non-empty output you return:

```java
sb.append("### ").append(new File(filePath).getName()).append("\n");
```

So one path wraps the filename in backticks and the other doesn’t. Minor, but easy to make consistent.

---

### 4. Migration guide says the old app will show these rows as `ToolKind.UNKNOWN`; worth double-checking against current catalog behavior
**File:** `doc/MIGRATION_1.13_TO_1.14.md`

The rollback section says 1.13.x will show the rows as `ToolKind.UNKNOWN` and reject them. That may be functionally true, but given the current architecture the more important effect is that these tool names are simply not present in the old static catalog / advertised descriptors. I’d reword this to avoid over-specifying internals unless you’ve verified the old release behavior exactly.

## Overall

Good change set, but I’d fix the `ctags-deps` invocation and the Python/member handling before merging. Those are both central to the feature and will otherwise make the new tools look flaky or incomplete in real use.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

This is a useful addition overall: the tool catalog, execution wiring, Docker dependency, migrations, and tests are all moving in the right direction.

I found **two blocking correctness issues** in the new `ctags` implementation, both in `ToolExecutionService`, plus a smaller maintainability concern around the migration docs/seed comments.

---

## Blocking issues

### 1. `ctags-deps` currently runs full-symbol extraction, so it will miss dependencies on many languages and can return lots of irrelevant tags
**File:** `src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java`

The implementation for `executeCtagsDepsTool()` no longer matches the design described in the doc and is likely incorrect in practice:

```java
String[] command = {"ctags", "--output-format=json",
        "--fields=+k",
        filePath.toAbsolutePath().toString()};
```

This asks ctags for *all* tags, then `formatCtagsDependencies()` filters only `kind == import/include/namespace/package`.

Two problems with that:

- Many ctags parsers do **not** emit import/include/package-ish entries by default unless kinds are explicitly enabled.
- Even when they do, this command may also emit many unrelated tags, which is unnecessary overhead and makes the behavior parser/version-dependent.

Your own implementation plan had the safer version:

```sh
--kinds-all=-* --kinds-all=+i+n
```

Without that restriction, `ctags-deps` can easily return `"dependencies":[]` for files that do have imports/includes, depending on language/parser defaults.

**Suggested fix:** restrict the ctags invocation to dependency-related kinds only, and verify the exact flags against the ctags version shipped in the image. At minimum, align the command with the intended behavior instead of relying on parser defaults.

---

### 2. `formatCtagsSignatures()` emits malformed pseudo-structure for scoped methods/classes because it never closes opened blocks and does not actually group by scope
**File:** `src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java`

The formatter claims to produce a structural map, but it currently just streams individual tags:

```java
case "class" -> tag.name() + " {";
case "interface" -> "interface " + tag.name() + " {";
...
case "method" -> "  method " + tag.name() + ...
```

That means output for a typical class becomes something like:

```java
OrderProcessor {
  constructor OrderProcessor()
  method process()
  method cleanup()
```

There is **never a closing `}`**, and there is no real parent/child grouping logic beyond indenting whenever `scope != null`.

That is more than cosmetic:

- nested types/scopes become ambiguous,
- multiple top-level classes/interfaces/modules become syntactically misleading,
- the “constructor if name equals scope” heuristic is applied without any actual enclosing structure,
- the output format described in docs/tests suggests a structured map, but the implementation is just a flat list with opening braces.

If this tool is meant to save context and improve agent reasoning, malformed structure is a correctness issue.

**Suggested fix:** either:
- implement real grouping by `scope` / `scopeKind` and emit balanced blocks, or
- simplify the output format to a clearly flat list (no `{`) so it does not pretend to be hierarchical.

---

## Non-blocking concerns

### 3. `Math.clamp(...)` may be a compatibility risk depending on the actual Java level used to compile the project
**File:** `src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java`

```java
limit = Math.clamp(limit, 1, MAX_CTAGS_SIGNATURES_LIMIT);
```

If this project is truly compiling on a JDK that includes `Math.clamp`, fine. If not, this will fail at compile time. Since I can’t run the build here, I’d recommend double-checking the effective toolchain/Maven config before merging.

If there’s any doubt, a simple `Math.min(max, Math.max(min, value))` is safer.

---

### 4. Migration/comments now conflict with the older V12 migration commentary
**Files:**
- `src/main/resources/db/migration/*/V12__bots_tool_configuration_fk.sql`
- `doc/MIGRATION_1.13_TO_1.14.md`

The new migration behavior is sensible, but the older V12 comments explicitly say later built-in tools must be enabled manually and that the application does not auto-extend Default. V29 now *does* extend Default via migration.

That’s not a runtime bug, but it creates historical/documentation inconsistency for future maintainers.

**Suggestion:** update the migration guide or add a note near V29/docs clarifying that V12’s comment was true at the time, but V29 is an explicit one-off extension of the default config.

---

## What looks good

- Tool registration in `ToolCatalog` is consistent with existing patterns.
- Path resolution / workspace confinement is reused correctly.
- Separate tests for parsing/formatting logic are a good addition.
- Docker image change is minimal and appropriate.
- The Flyway migration is idempotent and scoped to default configurations only, which matches the documented rollout.

If the two `ToolExecutionService` issues above are fixed, this looks close to ready.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

Nice feature overall: the tool registration, DB seed/migration, Docker dependency, and basic tests all line up. The main risk is in the new ctags output parsing/formatting logic inside `ToolExecutionService`, which currently makes some assumptions that are likely to produce misleading or low-value output for real files rather than a compact structural map.

## Findings

### Medium

#### 1. `ctags-signatures` does not actually group methods under their enclosing type as the docs promise
**Where:** `src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java`, `formatCtagsSignatures(...)` / `renderCtagsTag(...)`

The migration guide and development doc repeatedly describe this tool as producing a structural map and even show nested output like:

```java
class OrderProcessor {
  constructor ...
  method ...
}
```

But the implementation just renders each tag as an independent line in ctags order:

- `class OrderProcessor`
- `  method process()`
- `  method cleanup()`

There are no braces, no actual hierarchy, and no validation that the scoped members are emitted near their container. On languages / ctags outputs where ordering differs, the result can be confusing or misleading.

This is not just a docs mismatch: the tool’s value depends on conveying structure compactly. As implemented, it is basically a filtered flat tag list.

**Suggestion:** either:
- implement real grouping by `scope` / `scopeKind`, or
- tone down the docs/descriptions everywhere to say it returns a flat signature list, not a structural map.

#### 2. The hand-rolled JSON field extractor is brittle for ctags JSON
**Where:** `src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java`, `extractCtagsField(...)`

The parser scans strings manually and assumes the target field is always a quoted string value. That works for the happy path in the tests, but it is fragile against legitimate JSON variations:

- escaped quotes and backslashes are only partially handled
- unicode escapes are not decoded
- if ctags ever emits a non-string representation for a field you care about, the extractor silently returns `null`
- the method can match a field name anywhere in the line without validating JSON structure

Given the rest of the codebase already uses Jackson (`ToolCatalog` imports `tools.jackson`), a hand-written JSON parser here is a maintainability risk and harder to trust than using an actual JSON parser on each line.

**Suggestion:** parse each JSON line with Jackson into a small node/object and read `name`, `kind`, `signature`, `scope`, `scopeKind` from that. The extra dependency cost is effectively zero because the project already has it.

### Low

#### 3. `ctags-deps` may return duplicate dependencies with no deduplication
**Where:** `src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java`, `formatCtagsDependencies(...)`

Dependencies are collected into a `List<String>` and emitted verbatim. If ctags reports the same import/include more than once, the tool will return duplicates. That makes the output noisier and less useful for an LLM consuming it.

**Suggestion:** use a `LinkedHashSet<String>` to preserve order while deduplicating.

#### 4. Tests don’t cover the success path of the subprocess-backed tools
**Where:** `src/test/java/org/remus/giteabot/agent/validation/ToolExecutionServiceCtagsTest.java`

Most execution tests only cover argument validation / missing file / traversal, and one test explicitly expects failure because ctags is absent on the host. That’s understandable for portability, but it means the integration between:

- command invocation
- ctags flags
- parser expectations

isn’t actually verified anywhere.

Because the feature depends heavily on external tool behavior, this leaves room for surprises in production.

**Suggestion:** if CI can guarantee `ctags` availability, add a guarded integration test profile or environment-based test that exercises a real file end-to-end. If not, at least isolate command construction behind a smaller seam so it can be tested more directly.

## Overall

No blocking issues found in the wiring/migrations, but I’d strongly recommend tightening the ctags parsing/output behavior before relying on this as a “code-base structure extraction” tool. The current implementation is functional, but flatter and more brittle than the docs imply.

---

PR #235, opened 2026-06-28
bugfix/agentic-prworkflow-follow-ups

### LINKED ISSUE #231: Agentic PrReview Workflow does not allow further questions
**Current behavior**
If using the agentic Review PrWorkflow it's not possible to have a further conversation with the bot about its review. The simple PrReview Workflow does allow inline-comments and other commets mentioning with `@bot`

**Expected behavior**
Mentioning the bot with `@bot` and commenting a follow-up should be answered by the bot

tmseidel:


max-california (commented):
## 🤖 Agentic Code Review

## Summary

This fixes the main missing path for PR-comment `@bot ...` follow-ups on the agentic review workflow, and the new handler/test coverage is directionally good. However, it does **not** fully address the bug described in the PR plan: inline review comments and submitted reviews still only route to `ReviewWorkflow`, so bots configured with only `agentic-review` will continue to ignore those follow-up surfaces.

## Blocking

### 1. Inline comments and submitted reviews are still gated on `ReviewWorkflow.KEY` only
`src/main/java/org/remus/giteabot/admin/BotWebhookService.java:257` and `:284`

The PR description explicitly calls out three missing interaction surfaces: `@bot` mentions, inline comments, and review submissions. This patch only adds the new `AgentReviewSlashCommandHandler` into `handleBotCommand()` / `handlePrComment()`. It leaves:

- `handleInlineComment()` guarded by `isWorkflowEnabled(bot, ReviewWorkflow.KEY)`
- `handleReviewSubmitted()` guarded by `isWorkflowEnabled(bot, ReviewWorkflow.KEY)`

So a bot with only `agentic-review` enabled still won't respond to inline-comment or review-submission follow-ups. That means the bug remains partially unfixed relative to the stated issue.

**Suggestion:** either:
- add equivalent routing for agentic-review in those two methods, or
- narrow the PR scope/title/docs to say this fixes only top-level PR comment mentions.

## Major

### 2. Clarification runs ignore the workflow’s configured `maxToolRounds`
`src/main/java/org/remus/giteabot/prworkflow/agentreview/AgentReviewWorkflow.java:122-127`  
`src/main/java/org/remus/giteabot/prworkflow/agentreview/AgentReviewService.java:193-196`

Regular agentic review resolves `maxToolRounds` from workflow config, but clarification always calls:

```java
runReviewLoop(..., AgentReviewWorkflow.DEFAULT_MAX_TOOL_ROUNDS)
```

So operators who tuned the workflow budget won't get the same limit for follow-up runs. That’s inconsistent and can increase token/tool cost unexpectedly.

**Suggestion:** thread the resolved `maxToolRounds` into `answerClarification(...)`, just like `reviewPullRequest(...)`.

### 3. `@bot clarify` with no question is treated as “handled” but can never produce a response
`src/main/java/org/remus/giteabot/prworkflow/agentreview/AgentReviewSlashCommandHandler.java:71-85`  
`src/main/java/org/remus/giteabot/prworkflow/agentreview/AgentReviewWorkflow.java:97-100`  
`src/test/java/org/remus/giteabot/prworkflow/agentreview/AgentReviewSlashCommandHandlerTest.java:89-102`

The handler accepts `@bot clarify` with an empty captured question and dispatches it with `""`. But the workflow then ignores blank clarification hints:

```java
if (clarification != null && !clarification.isBlank()) {
    return doClarification(...)
}
```

That means:
- the slash-command handler returns `true`
- other handlers / fallback won't run
- the agentic workflow executes its **normal full review path** instead of replying to the user’s clarification request

This is a correctness bug and likely not what the user intended.

**Suggestion:** reject blank clarify commands in the handler and post a small usage/help reply, or at minimum return `false` so the normal fallback path can proceed intentionally.

## Minor

### 4. The fallback catch-all is broad enough to consume arbitrary mentions for agentic-review bots
`src/main/java/org/remus/giteabot/prworkflow/agentreview/AgentReviewSlashCommandHandler.java:43-45, 76-85`

Any `@bot <text>` becomes an agentic clarification whenever this handler is reached. Since it now sits before the old `ReviewWorkflow` fallback in `BotWebhookService`, this changes behavior for bots with both `review` and `agentic-review` enabled: freeform mentions that previously went through the non-agentic review conversation path will now always dispatch the agentic workflow.

That may be intended, but it’s a significant behavior change and I don’t see tests covering the mixed-enabled-workflows case.

**Suggestion:** add an explicit precedence test for bots with both workflows enabled, and decide whether freeform mentions should prefer `agentic-review` or the existing `review` workflow.

### 5. Model-list updates look unrelated to the bugfix
- `src/main/java/org/remus/giteabot/ai/anthropic/AnthropicProviderMetadata.java`
- `src/main/java/org/remus/giteabot/ai/google/GoogleAiProviderMetadata.java`

These changes may be fine, but they’re unrelated to the follow-up-routing bug and make review/risk assessment harder. If possible, split them into a separate PR.

## What looks good

- The new handler follows the existing E2E/unit-test slash-command pattern closely.
- Hydrating PR details before dispatch is a good match for issue-comment payloads.
- The added best-effort error comments in `AgentReviewService` are a useful UX improvement.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

Nice step forward overall: this closes the main routing gap for `agentic-review` mentions and adds the missing conversational path. The new slash-command handler follows the existing E2E/unit-test pattern well.

I do see **one blocking behavior regression** plus a couple of **maintainability / correctness risks** worth fixing before merge.

## Blocking

### 1. Inline comments / submitted reviews now silently stop working when both review workflows are enabled
**Files:** `src/main/java/org/remus/giteabot/admin/BotWebhookService.java:259-274`, `:295-310`

`handleInlineComment()` and `handleReviewSubmitted()` now do:

```java
if (agenticEnabled) {
    ...
    prWorkflowOrchestrator.run(bot, payload, AgentReviewWorkflow.KEY, hints);
    return;
}
```

That means if a bot has **both** `review` and `agentic-review` enabled, all inline-comment and review-submitted events are routed to the new agentic clarification path and the existing `ReviewWorkflow` path is never reached.

This is a behavior change from the previous implementation, where inline comments / submitted reviews were handled by the simple review workflow. It effectively makes the simple review conversation path unreachable for mixed configurations.

Why this matters:
- the two workflows have different UX and semantics;
- operators enabling both likely expect both workflows to remain usable, not for `agentic-review` to take precedence on these event types;
- this is inconsistent with `handlePrComment()` / `handleBotCommand()`, where explicit handlers compete first and only then fall back.

**Suggestion:** make precedence explicit instead of unconditional:
- either route inline/review-submitted events to `ReviewWorkflow` unless there is an explicit agentic signal,
- or only route to agentic when **review is not enabled**,
- or add separate mention/command detection for agentic follow-ups on those event types.

At minimum, document and test the chosen precedence.

## Non-blocking findings

### 2. New agentic behavior for inline/review events is untested
**Files:** `src/test/java/org/remus/giteabot/admin/BotWebhookServiceTest.java`

The new routing in `handleInlineComment()` and `handleReviewSubmitted()` is the highest-risk part of this change, but I don’t see corresponding tests added here.

Given the precedence issue above, I’d strongly recommend tests for:
- only `agentic-review` enabled → routes to `AgentReviewWorkflow`;
- only `review` enabled → routes to `ReviewWorkflow`;
- both enabled → verify intended precedence explicitly;
- blank inline/review body → verify whether it should dispatch or skip.

### 3. `AgentReviewSlashCommandHandler` duplicates hydration/reaction/error-posting logic already present in other handlers
**File:** `src/main/java/org/remus/giteabot/prworkflow/agentreview/AgentReviewSlashCommandHandler.java`

This mirrors `E2eTestSlashCommandHandler` / `UnitTestSlashCommandHandler` almost line-for-line. It’s fine short-term, but now there are three copies of:
- enabled-check pattern
- eyes reaction
- PR hydration
- best-effort error comment

That duplication makes provider-specific fixes likely to drift. Not a blocker for this PR, but I’d consider extracting a small shared helper/base component for PR slash-command handlers.

### 4. Provider model list changes are unrelated to the bugfix and increase rollout risk
**Files:**  
- `src/main/java/org/remus/giteabot/ai/anthropic/AnthropicProviderMetadata.java`  
- `src/main/java/org/remus/giteabot/ai/google/GoogleAiProviderMetadata.java`

These model-name updates are unrelated to the mention-followup bugfix. If any of these identifiers are invalid or not yet supported in production, they can break provider setup and obscure the PR workflow fix.

Recommend splitting these into a separate PR unless they are required for this feature.

## Notes

- `AgentReviewService.answerClarification()` looks consistent with the existing review loop and the added PR comment error reporting is a nice usability improvement.
- The prompt/tooling guidance updates also look sensible.

If you fix the inline/review precedence regression and add tests around that routing, this should be in good shape.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

The main follow-up support is headed in the right direction: `@bot clarify ...` now has a dedicated handler, the agentic workflow can run in a clarification mode, and `BotWebhookService` no longer hard-gates all conversational review traffic on `review` only.

I found one blocking behavior regression and one likely-correctness issue around review-submitted events. The rest looks reasonable.

## Blocking

### 1. Bots with both `review` and `agentic-review` enabled now route **all** inline/review-submitted follow-ups to agentic-review
**Files:** `src/main/java/org/remus/giteabot/admin/BotWebhookService.java:259-274`, `295-310`

The new logic does:

```java
if (agenticEnabled) {
    ...
    prWorkflowOrchestrator.run(bot, payload, AgentReviewWorkflow.KEY, hints);
    return;
}
```

That means if a bot has **both** workflows enabled, inline comments and submitted reviews will never reach `ReviewWorkflow` anymore. Previously, those events were handled by the conversational review flow (`ACTION_INLINE_COMMENT` / `ACTION_REVIEW_SUBMITTED`), which has session memory and existing semantics. This changes behavior globally, not just for agentic-only bots.

Why this matters:
- It’s a silent precedence change for existing dual-enabled bots.
- It discards the simple review workflow’s conversation/session behavior for these event types.
- It makes agentic-review “win” for events where the PR may actually have been reviewed by the non-agentic workflow.

Suggested fix:
- Only route inline/review-submitted events to `agentic-review` when `review` is **not** enabled, or when there is an explicit signal that the thread is tied to an agentic review.
- Otherwise preserve existing `ReviewWorkflow` handling.

For example, something like:
```java
if (agenticEnabled && !reviewEnabled) {
    ...
    return;
}
```
and fall through to `ReviewWorkflow` when both are enabled.

## Medium

### 2. `handleReviewSubmitted` may trigger agentic clarification on empty/non-conversational review submissions
**File:** `src/main/java/org/remus/giteabot/admin/BotWebhookService.java:302-307`

`extractReviewBody()` returns `null` for blank review content, but the caller turns that into `""` and still dispatches:

```java
var hints = Map.of(HINT_AGENTIC_REVIEW_CLARIFICATION, question != null ? question : "");
prWorkflowOrchestrator.run(... AgentReviewWorkflow.KEY, hints);
```

Then `AgentReviewWorkflow` skips clarification if the hint is blank, so the workflow run becomes a no-op. That’s not catastrophic, but it means review-submitted webhooks with no textual body can still create unnecessary workflow runs.

Worse, review-submitted events are often used for state changes/summary submission rather than an actual question. Treating every submitted review body as a clarification request may be too broad.

Suggested fix:
- Guard before dispatch:
```java
if (question == null || question.isBlank()) {
    return;
}
```
- Consider also checking review state/type if available, or only routing submitted reviews that actually mention the bot / contain non-empty conversational text.

## Low / Nit

### 3. The PR also changes suggested AI model names unrelated to the bugfix
**Files:**  
- `src/main/java/org/remus/giteabot/ai/anthropic/AnthropicProviderMetadata.java`  
- `src/main/java/org/remus/giteabot/ai/google/GoogleAiProviderMetadata.java`

These model-list changes look unrelated to the stated follow-up-support fix. If intentional, they should be split into a separate PR so behavior changes are easier to review and roll back independently. If not intentional, they should be dropped here.

## What looks good

- `AgentReviewSlashCommandHandler` follows the existing E2E/unit-test handler pattern closely.
- Hydrating PR details before dispatch is consistent with the existing command handlers.
- `AgentReviewWorkflow`’s hint-driven branching is clean and keeps the workflow ownership in the agentic package.
- Posting a best-effort failure comment from `AgentReviewService` is a useful UX improvement.

---

PR #239, opened 2026-06-28
bugfix/ctags-dep-wrong-call

tmseidel:


max-california (commented):
## 🤖 Agentic Code Review

## Summary

This looks like a real correctness fix: switching `ctags-deps` from kind-based filtering to `--extras=+r` plus `extras=reference` matching aligns with how universal-ctags actually emits import/include references in several languages, and the updated tests reflect that.

I didn’t find any blocking issue in the changed logic itself. The main remaining gap is test coverage for non-Java dependency extraction and for failure behavior.

## Findings

### Non-blocking

#### 1. Add a regression test for actual `include`/`import` kinds, not only `kind=package`
The new parser now treats any tag with `extras=reference` as a dependency, regardless of `kind`, which is fine and more robust than the old logic. However, the updated tests only cover samples where imports are emitted as:

- `kind="package", extras="reference"`

That validates the Java/TS behavior you observed, but it no longer exercises the previously intended cases where ctags may emit more explicit kinds like `include` or `import`.

**Why this matters:** the code comment says this is a cross-language fix, but the tests currently only pin one ctags-output shape. A small test with something like:

- `kind="include", extras="reference"`
- `kind="import", extras="reference"`

would make the regression protection stronger.

**Suggested location:** `ToolExecutionServiceCtagsTest.formatCtagsDependencies_*`

---

#### 2. The new warning log may not log the actual error message on failures
In `ToolExecutionService.java:1182-1184`, the code logs:

```java
log.warn("Error executing ctags-deps tool: {}", raw.output());
```

But `executeCommand()` returns failures in two different ways:

- process ran and failed: message is in `output`
- process could not start / timeout / interruption: message is in `error`

For example, when `ctags` is missing, `executeCommand()` returns:

- `output = ""`
- `error = "Failed to execute tool: ..."`

So this warning may become unhelpful in the most common failure case.

**Suggestion:** log both fields, or prefer `raw.error()` when present.

For example:

```java
log.warn("Error executing ctags-deps tool: {}", 
        raw.error().isBlank() ? raw.output() : raw.error());
```

Not blocking, but it would improve diagnosability.

---

#### 3. The relaxed `executeCtagsSignaturesTool_respectsLimitArg_clampedToMax` test now asserts very little
The modified test no longer checks success/failure semantics at all; it only checks `result != null`, which is effectively guaranteed by the method shape unless an exception escapes.

I understand the intent—make the test pass regardless of whether `ctags` is installed—but as written it doesn’t really validate the “clamped to max” behavior.

**Suggestion:** if you want this to stay environment-independent, assert something more specific about the contract that doesn’t depend on ctags being present, e.g. “does not throw” plus maybe split parsing/clamping into a helper that can be unit-tested directly. As-is, this test is very weak.

## Overall

The production change looks good and fixes the right problem. I’d be comfortable approving this after tightening the tests a bit, but I don’t see a blocker in the shipped behavior.

---

PR #240, opened 2026-06-30
fix: clone forked PRs by falling back to refs/pull/N/head

### LINKED ISSUE #238: Error when running Pull-Request workflows from forked repositories
**Current behavior**
When executing a PRWorkfow on a branch coming form a forked repository, the checkout fails and the PRWorkflow fails.

**Expected behavior**
Branches from forked repositories should work the same way like branches from the repository

**Implementation nodes**
This behavior seems to be limited to github, in gitea the checkout is possible
Error on github:

>  AI Agent (Review): Failed to prepare workspace: Failed to clone repository: Cloning into 'agent-workspace-15695567255036107174'...
warning: Could not find remote branch bugfix/empty-response-ontoken-limit to clone.
fatal: Remote branch bugfix/empty-response-ontoken-limit not found in upstream origin

tmseidel:


max-california (commented):
## 🤖 Agentic Code Review

## Summary

The core fix looks sensible: PR-based workflows now get a fallback path that can clone forked PRs via `refs/pull/<n>/head` when `git clone --branch <headBranch>` fails. The local-path handling in `buildCloneUrl()` also fixes an obvious test/dev regression.

I found one **blocking correctness issue** around push behavior after the fallback checkout, plus one **test coverage gap** that leaves the risky path unverified.

## Blocking

### 1. PR fallback leaves the workspace detached, which can break later `commitAndPush(...)`
**Files:** `src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java:91-103`, `132-180`  
**Impacts:** `UnitTestService`, and any future caller that both uses PR fallback and then pushes to the PR branch

In the fallback path, the code does:

```java
git fetch origin refs/pull/<n>/head
git checkout FETCH_HEAD
```

That leaves the repo in a **detached HEAD** state. Later, `commitAndPush(...)` assumes the workspace is already on the target branch when `createNewBranch == false`:

```java
// otherwise assumed to already be on the target branch
git push origin branchName
```

For `UnitTestService`, that assumption matters because `COMMIT_TO_PR` pushes back onto `headBranch` with `createNewBranch == false`. If the workspace came from the new fallback path, commits will be created on detached HEAD and then pushed using `git push origin <branchName>`. That push syntax pushes the **local branch named `<branchName>`**, not the current detached commit. In a forked-PR scenario where the initial branch clone failed, that local branch likely does not exist, so push can fail even though cloning succeeded.

**Suggested fix:** after fetching the PR ref, create/switch to a real local branch, e.g.

```bash
git checkout -B <branch> FETCH_HEAD
```

or otherwise adjust `commitAndPush(...)` to detect detached HEAD and push `HEAD:<branchName>` when needed. The former is probably cleaner because it preserves existing assumptions for downstream code.

## Non-blocking

### 2. The new test only validates read-only fallback, not the push-sensitive path
**File:** `src/test/java/org/remus/giteabot/agent/validation/WorkspaceServiceTest.java:32-65`

The added test proves that fallback can read the PR contents, which is good. But the highest-risk behavior change is what happens **after** fallback in workflows that commit back to the PR branch.

Given the detached-HEAD issue above, I’d strongly recommend adding a test that:

1. triggers the fallback clone,
2. makes a change,
3. calls `commitAndPush(..., createNewBranch=false)`,
4. verifies the remote branch/ref is updated as intended.

That would have caught the current problem and will protect future refactoring.

## Notes

- `buildCloneUrl()` passthrough for absolute/file URLs is a good change and matches the new test setup.
- Minor style nit only: `repositoryClient.getToken(),prNumber` in `AgentReviewService` is missing a space after the comma. Not important.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

Looks like a good fix overall: the fallback to `refs/pull/<n>/head` is the right direction for forked PRs, and the new integration test covers the main happy path. I found one blocking correctness issue around pushing after the fallback clone, plus a smaller maintainability concern.

## Blocking

### 1. Fallback checkouts can break later `git push origin <headBranch>` for forked PRs
**Files:** `src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java:77-103`, `src/main/java/org/remus/giteabot/prworkflow/unittest/UnitTestService.java:171-173`

After the fallback path, the workspace is created by:

1. cloning the base repository default branch,
2. fetching `refs/pull/<prNumber>/head`,
3. checking out a local branch named `headBranch` from `FETCH_HEAD`.

That gives you a local branch with the PR head contents, but it does **not** establish that `origin/<headBranch>` exists or is pushable on the cloned remote.

That matters because `UnitTestService` later does:

```java
workspaceService.commitAndPush(workspace, headBranch, ..., false);
```

and `commitAndPush()` pushes with:

```java
git push origin <branchName>
```

For a fork PR on GitHub/Gitea, `origin` is the base repository, while `headBranch` typically exists only on the fork. In that case the new fallback makes clone succeed, but the later push still targets the wrong remote/repo and will fail.

So this change fixes read-only PR workflows like agent review, but not write-back workflows like unit-test commit-to-PR for forked PRs.

**Suggestion:** either
- explicitly scope this fallback to read-only workflows only, or
- extend the workspace setup / repository client contract so fork PR workflows can push to the actual head repository remote, not always `origin`.

At minimum, add a test covering `UnitTestService` on a fork-style PR to make the expected behavior explicit.

## Non-blocking

### 2. `prepareWorkspace` now mixes two different responsibilities without documenting caller expectations
**Files:** `src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java:35-115`

The method used to mean “clone branch X”. It now sometimes means “clone default branch, fetch PR head ref, synthesize local branch X”. That’s fine, but downstream callers now have different semantics depending on whether `prNumber` is set:

- some callers need only a readable checkout,
- some callers assume the checked-out branch is also pushable to `origin`.

That distinction is currently implicit, which is what enabled the push issue above.

**Suggestion:** document this explicitly in the method Javadoc and/or split this into two entry points, e.g.:
- `prepareBranchWorkspace(...)`
- `preparePullRequestWorkspace(...)`

That would make the push/read-only behavior easier to reason about and reduce future regressions.

## Notes

- The `buildCloneUrl()` passthrough for `file://` and absolute paths seems reasonable and is covered by the new local-repo test scenario.
- I didn’t spot any obvious security regression in the fallback itself.

---

max-california (commented):
## 🤖 Agentic Code Review

## Summary

The fallback approach in `WorkspaceService` is a sensible fix for forked PRs in review flows, and the new integration-style test is useful. However, the change is only wired into `AgentReviewService`; other PR workflows still clone by `headBranch` with `prNumber = null`, so they will continue to fail on forked PRs. I’d treat that as a correctness gap for this PR.

## Findings

### High

#### `UnitTestService` still cannot clone forked PRs
`src/main/java/org/remus/giteabot/prworkflow/unittest/UnitTestService.java:97`

This workflow still calls:

```java
workspaceService.prepareWorkspace(
    owner, repo, headBranch,
    repositoryClient.getCloneUrl(), repositoryClient.getToken(), null);
```

For forked PRs, `headBranch` usually does not exist in the base repository, which is exactly the scenario this patch is fixing in `AgentReviewService`. As written, unit-test generation on forked PRs will still fail during workspace prep.

**Suggested fix:** pass the PR number here as well, same as in `AgentReviewService`, so `prepareWorkspace` can use the `refs/pull/<n>/head` fallback.

---

### Medium

#### The new fallback is PR-platform-specific but is now exposed through a generic workspace API
`src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java:56-103`

`WorkspaceService` is shared across GitHub, Gitea, GitLab, and Bitbucket clients, but the fallback hardcodes:

```java
refs/pull/<prNumber>/head
```

That ref is appropriate for GitHub/Gitea, but not generally for GitLab or Bitbucket. Today this is mitigated because only `AgentReviewService` passes a non-null `prNumber`, but the API now suggests a provider-agnostic capability that is not actually provider-agnostic.

This is not necessarily blocking if PR review is only enabled on GitHub/Gitea right now, but it would be safer to make the provider dependency explicit.

**Suggested fix options:**
- move “PR ref resolution” behind `RepositoryApiClient`, e.g. a method returning a fetchable ref for a PR when supported, or
- gate the fallback on provider capability instead of `prNumber != null` alone, returning the original branch-clone error where unsupported.

---

### Low

#### Missing coverage for the remaining PR workflow path
There’s a good test for `WorkspaceService` fallback itself, but no service-level test showing a forked PR path works from `AgentReviewService` and still fails/succeeds as intended elsewhere.

Given this bug is behavioral and workflow-specific, one targeted test around `AgentReviewService` or `UnitTestService` would help prevent regressions in call-site wiring.

**Suggested addition:** a test that verifies PR workflows pass `prNumber` into `prepareWorkspace` when operating on PR heads.

## Overall

Good direction and the core fallback logic looks reasonable, but I’d want the `UnitTestService` call site fixed before merging, otherwise forked PR handling remains inconsistent across PR workflows.

---

PR #242, opened 2026-07-02
fix: avoid cascade-delete on workflow configuration update

### LINKED ISSUE #241: Editing a workflow doesn't preselect the selected workflow
**Current behavior**
If you edit a workflow via `/system-settings`, then go via `system-settings/workflow-configurations/[id]/edit` to `system-settings/workflow-configurations/[id]/workflows` no workflow is selected, although this existing workflow has org.remus.giteabot.prworkflow.PrWorkflow assigned. If you save this form, no underlying org.remus.giteabot.prworkflow.PrWorkflow is assigned afterwards, the Bot that has this workflows assigned does nothing on Review.

**Expected behavior**
The perstisted org.remus.giteabot.prworkflow.PrWorkflow is preselected when entering the form at `system-settings/workflow-configurations/[id]/workflows`.

tmseidel:
Saving the detached entity from form binding directly in the update
path would cascade-delete all persisted selectedWorkflows, because the
detached entity's list is empty. The fix now fetches the managed
entity via findById, updates its name, and returns the managed
entity instead — preserving the workflow associations.

### DIFF
--- a/src/main/java/org/remus/giteabot/prworkflow/config/WorkflowConfigurationService.java
+++ b/src/main/java/org/remus/giteabot/prworkflow/config/WorkflowConfigurationService.java
@@ -52,28 +52,32 @@ public class WorkflowConfigurationService {
         if (configuration.getName() == null || configuration.getName().isBlank()) {
             throw new IllegalArgumentException("Name is required");
         }
-        configuration.setName(configuration.getName().trim());
+        String trimmedName = configuration.getName().trim();
 
         if (configuration.getId() != null) {
+            // Existing: update only name on the managed entity. The detached
+            // entity from form binding has an empty selectedWorkflows list —
+            // saving it directly would cascade-delete all persisted selections.
             WorkflowConfiguration existing = configurationRepository.findById(configuration.getId())
                     .orElseThrow(() -> new IllegalArgumentException("Workflow configuration not found"));
             if (existing.isDefaultEntry()) {
-                if (!existing.getName().equals(configuration.getName())) {
+                if (!existing.getName().equals(trimmedName)) {
                     throw new IllegalArgumentException("The default workflow configuration cannot be renamed");
                 }
-                configuration.setDefaultEntry(true);
             }
-        } else {
-            // New configurations are never default; the default is bootstrapped once.
-            configuration.setDefaultEntry(false);
+            if (configurationRepository.existsByNameAndIdNot(trimmedName, configuration.getId())) {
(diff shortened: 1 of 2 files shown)


max-california (commented):
## 🤖 Agentic Code Review

## Summary

Looks correct to me. The change addresses a real JPA/orphan-removal hazard in the update path by switching from saving the detached form-bound entity to mutating the managed entity loaded from the repository. In the context of `WorkflowConfiguration.selectedWorkflows` being `cascade = ALL, orphanRemoval = true`, this is the right fix and should preserve existing workflow selections during rename/update.

I did not find any blocking correctness or security issues in the patch itself.

## Findings

### No blocking issues found

The service change is consistent with the entity mapping and controller flow:

- `WorkflowConfiguration.selectedWorkflows` is `@OneToMany(..., cascade = CascadeType.ALL, orphanRemoval = true)`, so persisting a detached instance with an empty collection can indeed delete children.
- `WorkflowConfigurationController.save(...)` binds a `WorkflowConfiguration` directly from the form, so the service must defensively avoid trusting omitted associations on update.
- Returning the managed `existing` entity from a transactional service method is sufficient for JPA dirty checking to persist the renamed value without explicitly calling `save(...)`.

### Minor suggestions

#### 1. Add a regression test for preserving `selectedWorkflows`
The current test adjustment only removes the mocked `save(...)` call for one update scenario. It verifies the new control flow indirectly, but it does not explicitly prove the bug is fixed.

Suggested test coverage in `WorkflowConfigurationServiceTest`:
- create a persisted `WorkflowConfiguration` with one or more `selectedWorkflows`
- create an update DTO/entity with same `id` and new `name`, but no `selectedWorkflows`
- call `save(...)`
- assert the returned object is the persisted instance and still contains the original selections

That would make the intent of this PR much clearer and protect against future refactors reintroducing the cascade-delete bug.

#### 2. Consider aligning the analogous `BotToolConfigurationService`
`src/main/java/org/remus/giteabot/systemsettings/BotToolConfigurationService.java` already has a similar “preserve managed entity and collection” pattern, but it still calls `configurationRepository.save(target)` on update.

That is not wrong, but for consistency with this fix you may want to consider the same style there:
- load managed entity
- mutate it
- return managed entity

This is a maintainability/readability suggestion, not a defect in this PR.

## Overall

Good fix. The behavior described in the PR matches the surrounding code, and the change is appropriately scoped. The only thing I’d still want is an explicit regression test demonstrating that existing workflow selections survive an update from a detached form object.

---

PR #259, opened 2026-07-10
feat(testing): added e2e tests and a runner for local deployments to avoid regressions on a e2e-level

tmseidel:


max-california (changes requested):
## 🤖 Agentic Code Review

## Summary

This PR adds a local Playwright runner and extends the E2E suite, which is useful. However, there is one clear blocking regression in application configuration and one test-suite maintainability issue that is likely to cause flaky/non-isolated runs.

## Blocking issues

### 1. `docker-compose.yml` changes the default public URL to a hard-coded external IP
In `docker-compose.yml`:

```yml
APP_PUBLIC_URL: ${APP_PUBLIC_URL:-http://167.235.231.84:8080}
```

Previously this default was `http://localhost:8080`. Hard-coding an external server IP as the fallback for local Docker runs is a behavior change unrelated to adding E2E tests, and it is likely incorrect for most developers. It can break generated links, callbacks, redirects, or any feature depending on the app’s canonical public URL when running locally.

This should either stay as `http://localhost:8080` or be documented and justified separately in a dedicated change.

## Non-blocking findings

### 2. New E2E test mutates shared persisted state without suite-level isolation
`tests/e2e/pr-254/review-workflow-excluded-file-patterns-persists.spec.ts` edits a real persisted workflow configuration (`/system-settings/workflow-configurations/1/workflows`) and restores it in `finally`, which is good as best effort. But the suite is explicitly configured to run against a shared backend with `workers: 1`, and multiple existing tests also persist state. If this test fails after saving but before cleanup, later tests may observe modified state.

Not necessarily a blocker given the current runner model, but it would be more robust to:
- save via a more explicit success condition than `networkidle`,
- assert the save completed before proceeding,
- and, if possible, seed/reset test data instead of relying on best-effort restoration.

### 3. `.gitignore` only ignores `tests/e2e/node_modules/`
The new Playwright config writes to `../../target/...`, which seems already ignored per the comment, so that part is fine. But local Playwright runs often also generate browser/report artifacts under the test directory depending on invocation/config changes. Not required here, just worth keeping an eye on if contributors start committing local artifacts.

## What looks good

- `tests/e2e/run.sh` is straightforward and developer-friendly.
- `playwright.config.ts` uses `workers: 1` / `fullyParallel: false`, which matches the shared-state nature of these tests.
- Updating the existing PR-174 specs from 6 to 8 sections appears internally consistent.

---

max-california (changes requested):
## 🤖 Agentic Code Review

Summary: the added local Playwright runner is useful and the updated PR-174 assertions match the current templates, but I found one blocking issue in the new persistence test and one maintainability risk in the runner setup.

## Blocking

### `tests/e2e/pr-254/review-workflow-excluded-file-patterns-persists.spec.ts`: hard-coded configuration ID makes the test environment-specific
The new test navigates to:

- `'/system-settings/workflow-configurations/1/workflows'`

and assumes the relevant workflow configuration always has ID `1`.

That is brittle for a local runner whose stated goal is to work against arbitrary local deployments. Any non-fresh database, imported dataset, or environment where the default configuration was recreated will cause a false failure unrelated to the feature under test.

This is especially noticeable because the PR also removes `preview-app-loads.spec.ts`, which was the only deployment-agnostic smoke test under `pr-254`; after this change, the only remaining test in that folder depends on a specific database row existing at a fixed ID.

**Suggestion:** navigate via a stable UI path or discover the configuration dynamically before editing it, e.g. start from `/system-settings`, follow the workflow configuration link for the default/first entry, then assert and mutate from there. If the test really requires a known seeded record, document that precondition explicitly and make the lookup by name rather than by numeric ID.

## Non-blocking

### `tests/e2e/run.sh`: installs dependencies with mutable `npm install`
For a checked-in `package-lock.json`, using `npm install` in the runner makes local execution less reproducible and can silently rewrite the lockfile on developers’ machines. Since this script is intended as a repeatable local harness, `npm ci` would be a better default when `package-lock.json` is present.

Also, the script checks only for `node_modules/@playwright/test`; if dependencies are partially present or out of sync with the lockfile, the install step may be skipped incorrectly.

**Suggestion:** prefer `npm ci` unconditionally when `package-lock.json` exists, or at least document why `npm install` is required here.

### `.gitignore`: comment says target output is ignored, but only `tests/e2e/node_modules/` was added here
`playwright.config.ts` comments say artifacts land under the project’s root `target/` and are “already git-ignored”. That may be true elsewhere, but this PR’s ignore update only adds `tests/e2e/node_modules/`. Worth double-checking that `target/playwright-report/` and `target/test-results/` are in fact ignored at repository root, otherwise local runs will produce untracked files despite the comment.

---

max-california (dismissed):
## 🤖 Agentic Code Review

Summary: the new local Playwright runner is a useful addition, and the updated accordion-count assertions match the current templates. I did find one blocking issue in the new PR-254 test: it mutates shared workflow configuration and restores the wrong value when the target field was initially unset.

## Blocking

### `review-workflow-excluded-file-patterns-persists.spec.ts` can corrupt the original workflow config when the field starts empty
In `tests/e2e/pr-254/review-workflow-excluded-file-patterns-persists.spec.ts`, cleanup restores `originalValue` like this:

```ts
await input.fill('');
if (originalValue.length > 0) {
  await input.fill(originalValue);
}
await saveButton.click();
```

However, in the template the field value is computed as persisted value **or** default value:

- `src/main/resources/templates/system-settings/workflow-configurations/workflows.html:86-143`

That means an empty visible value does **not** necessarily mean “persisted empty string”; it may mean “no persisted override, currently showing the schema default”, and conversely a non-empty visible value may also just be the default. Your test always saves during cleanup, so when the field was originally unset and `originalValue === ''`, the cleanup writes an explicit empty string override into configuration. That changes app state instead of restoring it, and can affect subsequent runs and real data on a local deployment.

Actionable fix:
- capture whether the field had a persisted override before modifying it, not just the rendered input value; or
- use a dedicated test configuration/entity created for the test; or
- avoid saving during cleanup when the original state was “unset/default inherited”, because the UI currently has no way to “remove persisted override” once you submit an explicit value.

## Non-blocking

### Deleting `preview-app-loads.spec.ts` reduces baseline smoke coverage
`tests/e2e/pr-254/preview-app-loads.spec.ts` was removed and replaced with a workflow-setting persistence test. The new test is more feature-specific, but the deleted one provided a cheap “deployment is up and booted” smoke check. If this removal is intentional, consider adding an equivalent lightweight root-load smoke spec elsewhere so local/preview runs still fail fast on obvious app-start issues.

### Minor maintainability nit: `.gitignore` comment formatting
`.gitignore` adds:

```gitignore
#e2e test output
tests/e2e/node_modules/
```

Small nit: add a space after `#` for consistency with the rest of the file.

---

PR #260, opened 2026-07-10
fix(bots): show all 8 system-prompts in edit-page preview

### LINKED ISSUE #258: System-Prompts review misses 2 system-prompts
**Current behavior**
On the bots editpage `bots/[id]/edit` is a button to preview all 8 System-Prompts of the selected System-Prompt configuration. The preview shows only 6 of them

**Expected behavior**
Show all System-Prompts of the selected configuration (`Review-Agent System-Prompt` and `Unit-Test Author System-Prompt` are missing)

tmseidel:


max-california (approved):
## 🤖 Agentic Code Review

I have enough context from the diffs and the surrounding template code. Here is my review:

---

## Review Summary

The PR adds two missing accordion sections to the system-prompt preview modal in the bot edit page — for **Review-Agent** and **Unit-Test Author** prompts — bringing the total from 6 to 8. The HTML markup follows the existing pattern exactly, the JavaScript DOM references and render calls are wired up correctly, and the e2e test is updated accordingly. The change is straightforward and self-consistent.

---

### ✅ Correctness

- **Markup**: Each new `.accordion-item` block pairs an `accordion-button` with a matching `data-bs-target` / `id` on the collapse div, and the inner preview `<div>` has a unique `id` that matches the JavaScript lookup. All `aria-*` attributes are present.
- **JavaScript**: Two new `getElementById` references (`reviewAgentPromptPreview`, `unitTestAuthorPromptPreview`) match the `id`s in the HTML, and both are rendered via `renderMarkdown(selectedPrompt.<field>)` in the fetch `.then()` callback — consistent with the existing six prompts.
- **Test**: Count updated from 6→8, and the two new header texts are asserted. The test keeps the same structure and escape-guard at the end.

No bugs or mismatches are apparent in the diff.

---

### ⚠️ Backend dependency (call-out, not a blocker in this diff)

The template now reads `selectedPrompt.reviewAgentSystemPrompt` and `selectedPrompt.unitTestAuthorSystemPrompt` from the JSON response at `/system-settings/system-prompts/{id}/preview`. **If the backend SystemPrompt entity / DTO does not yet expose these fields**, the preview cells will silently show empty or `undefined`. I cannot verify the Java/Kotlin side from the workspace, so please confirm those fields exist in the response. If they do, everything is fine.

---

### 🔍 Minor observations (non-blocking)

- **Test header assertions are inconsistent**: The test checks some headers by full name (`"Review-Agent System-Prompt"`) and others partially (`"E2E Planner"`, `"E2E Author"` — missing `" System-Prompt"`). This isn't introduced by the PR (it was already uneven at 6 items) but is worth cleaning up eventually for clarity.

- **First section remains open**: The "Review System-Prompt" section is `collapse show` / `aria-expanded="true"` while all others (including the two new ones) are `collapsed`. This is correct — only one section should be open by default.

---

### Verdict

The change is correct, consistent, and well-tested. No blocking issues.

---

PR #261, opened 2026-07-10
 fix(agentreview): only respond to review submissions that mention the bot

tmseidel:


max-california (approved):
## 🤖 Agentic Code Review

I have enough context. Here is my review.

---

## Review Summary

This is a focused, well-tested change that gates the agentic-review workflow on whether the bot is `@`-mentioned in the review body. It prevents the bot from reacting to every review submission (e.g. approvals from other human reviewers), which is the correct behaviour.

**Verdict: APPROVE**

---

## Findings

### 1. ✅ Behavioural logic is correct

The `mentionsBot` check is inserted at the right point — inside the `if (agenticEnabled)` block that already had its own `return`, so the non-agentic review path (when `reviewEnabled` is also true) was already unreachable before this change. No regression in the dual-workflow case.

The `question` variable is now guaranteed non-null/non-blank after the `mentionsBot` guard, so the old ternary (`question != null ? question : ""`) in the hints map is correctly simplified to just `question`. Clean.

### 2. ✅ Tests are thorough and well-named

Three test behaviours are covered:

| Test | Condition | Expected |
|---|---|---|
| `reviewSubmitted_agenticReviewEnabled_dispatchesClarification` | body contains `@claude_bot` | workflow dispatched with full body as hint |
| `reviewSubmitted_noReviewBody_agenticReviewIgnored` | `null` body | workflow **not** dispatched |
| `reviewSubmitted_bodyWithoutBotMention_agenticReviewIgnored` | body without mention | workflow **not** dispatched |

The renamed test `reviewSubmitted_noReviewBody_agenticReviewIgnored` correctly reflects the changed behaviour (previously it dispatched with an empty hint; now it's ignored entirely).

### 3. ℹ️ Minor: case-sensitive mention matching (nit, not blocking)

`mentionsBot` uses `String.contains()`, which is case-sensitive. If a user types `@Claude_Bot` but the stored username is `claude_bot`, the bot won't respond. This is **consistent** with the pre-existing `isReviewAgainRequest` (line 706), which also uses case-sensitive `contains` for bot aliases — so no new problem is introduced. Still, a future improvement could be to case-normalise both sides (e.g. `text.toLowerCase().contains(alias.toLowerCase())`).

### 4. ℹ️ Minor: substring false-positive (nit, not blocking)

A review body containing `@claude_bot_extra` would match the bot `@claude_bot` via `String.contains`. In practice, Gitea usernames don't typically overlap like this, so it's extremely low risk. Not worth addressing.

---

No blocking issues. The change is clean, test-covered, and addresses the stated goal precisely.

---

PR #271, opened 2026-07-13
feat(pr-workflow): add i18n Coverage workflow Add a new opt-in PR workflow (key: i18n-coverage, category: DOCS) that detects when a pull request's i18n changes (added keys, changed values, deleted keys) are not consistently reflected across locale files, then drafts the missing translations per locale with an LLM agent and removes stale keys.

### LINKED ISSUE #269: New PRWorkflow: i18n-coverage
Originates from #268

## Quality assessment

The scope is now mostly clear and implementation-ready. The main behavior, triggering, configurability, slash-command semantics, and suite lifecycle expectations are defined. Remaining uncertainty is limited to non-critical implementation details such as exact comment format and per-file matching rules.

## Summary
Add a new agentic PR workflow: `i18n-coverage`.

This workflow detects i18n key/value coverage mismatches across locale files in a PR and drafts missing translations with the LLM, scoped per locale. It should support both `messages_*.properties`-style files and `i18n/*.json`-style files, based on configurable include patterns.

## Goal
Help maintain translation coverage when a PR changes user-facing strings by detecting when i18n changes are not consistently reflected across locale files.

"Changes" include:
- added keys
- changed values
- deleted keys

## Scope
This is a new `PRWorkflow`.

Framework support for creating a new PR workflow already exists, so no DB migration is required.

The workflow run is agentic, similar to the existing Agentic Review workflow.

## Triggering
The workflow should be triggered when:
- a bot is assigned to the PR, and
- that bot has this workflow assigned

This should follow the same triggering procedure used by the default review workflows.

Target branch: `develop`

## Configuration
The workflow should be creatable and configurable the same way as existing PR workflows.

Add the following workflow properties:

1. **Included file patterns for i18n files**
   - A configurable list of file patterns used to identify i18n files in the repository.
   - This must support patterns such as `messages_*.properties` and `i18n/*.json`.

2. **Baseline locale**
   - A configurable locale set on the workflow edit page.
   - The baseline locale is the reference locale whose keys/values/deletions are compared against the other locale files.

3. **Suite lifecycle**
   - Same lifecycle behavior as the e2e-test workflow.
   - Use `doc/agentic-workflows/SUITE_PROMOTION_USER_STORY.md` as the implementation reference.
   - When missing translations are detected, the workflow should follow the configured suite promotion behavior.

### Detection
For i18n files matching the configured include patterns, the workflow should detect PR changes affecting translation coverage relative to the configured baseline locale.

The workflow must consider all of the following as relevant changes:
- newly added translation keys
- changed translation values
- deleted translation keys

The workflow should identify when corresponding locale files are missing the equivalent updates.

### Agentic generation
When coverage issues are found, the workflow should use the LLM to draft the missing translations, scoped per locale.

Expected output:
- a short final summary comment after the workflow finishes
- generated translation drafts for the affected locales

### Slash command
Add a slash handler similar to `org.remus.giteabot.prworkflow.e2e.E2eTestSlashCommandHandler`.

Command format:

```text
@bot regenerate-i18n Please use for the french translation the word "flâner"
```

Behavior:
- re-run the workflow for the current PR
- use the trailing free-text instruction as additional guidance for the LLM-generated translations
- do not change the detection logic

## Implementation notes
- Implement as a new `PRWorkflow`.
- Reuse existing framework patterns for workflow creation and setup.
- Reuse the agentic execution approach used by the Agentic Review workflow where appropriate.
- Reuse slash-command handling patterns from `org.remus.giteabot.prworkflow.e2e.E2eTestSlashCommandHandler`.
- No DB migration is needed.

## Acceptance criteria
- A new PR workflow named `i18n-coverage` can be created and configured through the existing workflow setup flow.
- The workflow exposes a configurable list of include patterns for i18n files.
- The workflow exposes a configurable baseline locale on the workflow edit page.
- The workflow is triggered when an assigned bot with this workflow enabled is assigned to a PR, following the same mechanism as existing default review workflows.
- The workflow runs agentically.
- The workflow uses the e2e-test-style suite lifecycle / promotion behavior.
- The workflow inspects changed i18n files matching the configured patterns.
- The workflow treats added keys, changed values, and deleted keys as relevant changes.
- The workflow detects when non-baseline locale files are missing corresponding updates.
- The workflow drafts missing translations via the LLM per affected locale.
- After completion, the workflow posts a short summary comment.
- A slash command `@bot regenerate-i18n ...` is available.
- The slash command re-runs the workflow for the PR and passes the trailing text as extra translation guidance.
- No database migration is introduced.

## Suggested test coverage
- Workflow can be created and persisted without schema changes.
- Workflow configuration accepts include patterns and a baseline locale.
- Triggering matches existing assigned-bot PR workflow behavior.
- Detection works for:
  - added keys
  - changed values
  - deleted keys
- Detection works for both properties-based and JSON-based i18n file sets.
- Agentic execution produces a final short summary comment.
- Slash command parsing correctly extracts optional free-text guidance.
- Slash command re-runs the workflow for the current PR.

## Assumptions

- The workflow should report and draft translation updates through the standard suite promotion / agentic workflow mechanisms rather than introducing a new output channel.
- The baseline locale acts as the reference set of keys/values/deletions for determining whether other locale files need corresponding updates.
- Exact file-to-file locale matching can follow the existing repository/project conventions as long as it is driven by the configured include patterns and baseline locale.

## Open questions

- How should the workflow behave if the baseline locale file is not changed in the PR but non-baseline locale files are changed?
- For deleted keys, should the workflow only report missing deletions, or should it also draft explicit deletion guidance in the output?
- Should the workflow compare all locale files in the repository that match the configured patterns, or only files touched by the PR plus their locale counterparts?

---

tmseidel:


max-california (changes requested):
## 🤖 Agentic Code Review

Summary: the workflow structure is solid and the safety model is thoughtful, but I found one blocking correctness issue in the locale parser that will mis-detect bundle families and baselines for common filenames.

## Blocking

### `I18nFileParser` misclassifies non-locale suffixes as locales, which breaks family grouping and baseline selection
`src/main/java/org/remus/giteabot/prworkflow/i18n/I18nFileParser.java:54-60, 84-105`

The `LOCALE_SUFFIX` pattern is too permissive:

```java
Pattern.compile("^(?<base>.*?)(?<sep>[._-])(?<locale>[a-z]{2}(?:[_-][A-Za-z]{2,})*)$")
```

Because the trailing `locale` part accepts any 2+ letters after `-` / `_`, filenames like:

- `messages_pt-BR-extra.properties`
- `messages_release-notes.properties`
- `foo_en_backup.json`

can be parsed as if they contain a locale suffix, even though the trailing token is not a valid locale. The parser then strips part of the stem into `locale`, changing the computed `familyId`.

That matters because `I18nCoverageDetector` groups files by `familyId` and picks a baseline per family. Mis-grouping means the workflow can silently:

- compare the wrong files together,
- miss real coverage gaps,
- or treat unrelated files as a locale family.

The current tests only cover happy-path names like `messages_en.properties` and `fr-FR.json`; they don’t exercise this ambiguity.

**Suggested fix:** tighten locale detection to accepted locale shapes only, e.g. language + optional region/variant with explicit constraints, or better yet parse known bundle conventions separately for `.properties` and bare-locale JSON names. At minimum, add tests covering stems with non-locale trailing segments to ensure they remain part of the base name rather than being treated as locale tokens.

## Non-blocking

### Consider sorting `members` before baseline selection for deterministic fallback
`src/main/java/org/remus/giteabot/prworkflow/i18n/I18nCoverageDetector.java:123-179`

`pickBaseline()` falls back to lexicographically first path only at the very end, but the exact-match / language-only / implicit-default checks iterate `members` in discovery order. Today `findInScopeFiles()` sorts paths, so this is probably deterministic in practice, but sorting `members` once before selection would make that contract explicit and less fragile if file collection changes later.

### Properties parsing is intentionally simplified, but docs overstate format preservation
`src/main/java/org/remus/giteabot/prworkflow/i18n/I18nFileParser.java:17-24`
`src/main/java/org/remus/giteabot/prworkflow/i18n/I18nCoveragePromptLibrary.java:47-55`

The parser flattens `.properties` into key/value pairs and ignores comments, continuation lines, separator spacing, unicode escapes, etc. That’s okay if the agent rewrites whole files from visible content, but the prompt currently says to preserve “escaping and structure” for Java properties. That promise is stronger than what the implementation can reliably support. I’d soften the wording or explicitly scope support to simple `key=value` / `key:value` files.

---

max-california (changes requested):
## 🤖 Agentic Code Review

## Summary

Nice addition overall: the workflow is thoughtfully integrated with the existing PR-workflow architecture, has good tests, and includes layered write-scope guards. I found two correctness issues that look merge-blocking, both around how locale files are parsed/rewritten.

## Blocking findings

### 1. `.properties` parsing is not compatible with Java properties syntax, so the workflow can mis-detect gaps and permanently corrupt valid translation files
**Files:** `src/main/java/org/remus/giteabot/prworkflow/i18n/I18nFileParser.java` lines ~131-161, and by extension the whole workflow

`parseProperties()` is a very simplified parser:

- it ignores lines without `=` or `:`
- it strips leading whitespace before parsing
- it does not support the standard whitespace separator (`key value`)
- it does not support line continuations with trailing `\`
- it does not unescape keys/values

That is not just an approximation for detection: the agent is instructed to rewrite the **complete file**, and the prompt tells it to preserve existing keys based on the parsed view of the current file. So valid Java `.properties` files can be misread, causing keys to appear “missing”/“stale”, and then rewritten into a broken or lossy form.

Concrete examples that are valid in `.properties` but mishandled here:

- `greeting Hello` → ignored entirely
- multiline values with trailing `\` → split into bogus lines / dropped content
- escaped separators or spaces in keys → parsed incorrectly
- unicode escapes / escaped punctuation → compared as raw text, not semantic values

This is especially risky because the default patterns explicitly include `messages_*.properties`.

**Suggested fix:** use `java.util.Properties`-compatible parsing semantics, or constrain support to a clearly narrower format and reject unsupported files instead of silently parsing them incorrectly. At minimum, add tests for whitespace-separated entries and line continuations before shipping this workflow.

---

### 2. JSON translation files will be flattened for detection, but the write tool only accepts whole-file text, so nested JSON cannot actually be preserved as promised
**Files:** `src/main/java/org/remus/giteabot/prworkflow/i18n/I18nFileParser.java` lines ~163-195, `src/main/java/org/remus/giteabot/prworkflow/i18n/I18nCoveragePromptLibrary.java` lines ~47-55

The parser flattens nested JSON into dot-path keys (`a.b.c`). That is fine for comparison, but the workflow never stores the original JSON structure in a machine-usable form. The agent is then asked to write back the **complete file** while also preserving “nested vs. flat JSON”.

In practice, for nested JSON files the model only sees raw file text plus a flattened missing/stale-key list. There is no deterministic transformation layer that can apply those flattened edits back onto the original object tree. This makes the “preserve structure” promise unreliable and can easily lead to:

- nested JSON being rewritten as flat dot-key JSON
- arrays/objects being reshaped unexpectedly
- ordering/formatting churn unrelated to the requested coverage fix

Given that `i18n-write` blindly overwrites the file, this is a real data-shape corruption risk, not just cosmetic diff churn.

**Suggested fix:** either:
- implement structured JSON patching/re-emission in Java, preserving object structure and only letting the model draft translated values, or
- limit JSON support to already-flat objects and explicitly reject nested JSON files during detection.

## Non-blocking note

- The symlink check in `I18nToolExecutor.resolveInsideWorkspace()` only checks the final path, not symlinked parent directories. That pattern is inherited from `ReadmeSyncToolExecutor`, so it’s not new in this PR, but if you ever harden these workspace write tools further, walking parent segments with `toRealPath()` would close that gap.

---

max-california (dismissed):
## 🤖 Agentic Code Review

Summary: the workflow is thoughtfully integrated and has decent guardrails/tests, but I found one blocking correctness issue in `.properties` parsing that can produce false “missing/stale key” detection for valid i18n files. I also have one non-blocking concern about JSON support/documentation drift.

## Blocking

### 1. `.properties` continuation lines are parsed incorrectly, which can invent coverage gaps
`src/main/java/org/remus/giteabot/prworkflow/i18n/I18nFileParser.java`

The parser claims to follow `java.util.Properties#load` semantics, but continuation handling currently appends the next physical line verbatim:

- `assembleLogicalLine(...)` appends the continued line without stripping its leading whitespace.
- `countPhysicalLinesConsumed(...)`/`assembleLogicalLine(...)` therefore turn:

```properties
key=value\
  continued
```

into `value  continued`, whereas `Properties.load()` discards the leading whitespace on continuation lines.

This matters because the workflow compares parsed key sets across locales. If a baseline or locale file contains continued entries, the current parser can produce values that differ from real Java resource-bundle semantics, and more importantly can shift separator parsing in edge cases enough to mis-read entries. Since this workflow is supposed to operate on standard `.properties` bundles, it should not diverge from the platform parser here.

I don’t see a test covering the required whitespace trimming after a continuation. The existing `lineContinuation()` test actually asserts the current non-standard behavior.

**Suggested fix:** make logical-line assembly match `Properties.load()` semantics by skipping leading space/tab/form-feed on continuation lines before appending them, and add regression tests for standard continuation behavior.

## Non-blocking

### 2. Nested JSON is documented as supported, but the implementation silently skips it
The public docs / workflow description say JSON locale files are supported, but `I18nCoverageDetector.detect(...)` skips any JSON file where `I18nFileParser.isNestedJson(...)` is true.

That behavior may be intentional, but it creates a product mismatch:
- docs/UI copy say JSON is supported broadly,
- prompt/migration text says “preserve … nested vs. flat JSON”,
- implementation only safely handles flat root objects.

At minimum, the user-facing docs and parameter help should state “flat JSON objects only” to avoid surprising no-op runs on common nested i18n layouts.

Overall, aside from the `.properties` parsing bug, the design looks solid: branch resolution is careful, scope guards are in place at write time and pre-commit, and the tests cover many happy-path and regression scenarios.

---

PR #277, opened 2026-07-15
fix(review): propagate bot response write failures

### LINKED ISSUE #274: fix(review): propagate bot response write failures
## Observed behavior

`CodeReviewService#handleBotCommand` catches every exception, logs it, and returns normally. If the AI response was generated but `RepositoryApiClient#postPullRequestComment` failed, `ReviewWorkflow` can still return `SUCCESS` even though no response reached the pull request.

## Expected behavior

The command handler should keep its detailed log and propagate the failure. The existing workflow orchestrator should then mark the run failed, and the existing bot-level error handling should record the error.

## Provider scope

The failure is in the shared review layer and affects every provider through `RepositoryApiClient`. Provider retry semantics and API clients are intentionally unchanged.

## Relevant code path

`CodeReviewService#handleBotCommand` swallows the exception; `ReviewWorkflow#doBotCommand` therefore returns `WorkflowResult.success("Bot command handled")`.

## Acceptance criteria

- A PR-comment write exception propagates from `handleBotCommand`.
- `ReviewWorkflow` does not return success for that failure.
- The existing orchestrator failure path remains responsible for persisted `FAILED` status.
- Successful command behavior remains unchanged.
- No automatic comment retry is added.
- Focused tests and the full test suite pass.

SingularityKChen:
## What

Propagate failures from conversational PR-response handling so a failed comment write cannot be reported as a successful review workflow.

## Why

`CodeReviewService#handleBotCommand` currently logs and swallows every exception. When the provider rejects `postPullRequestComment`, the workflow can still finish successfully even though the user received no response.

## How

- keep the existing detailed error log;
- rethrow runtime failures and wrap an unexpected checked failure;
- rely on the existing workflow orchestrator to persist `FAILED` and on existing bot-level handling to record the error;
- add no automatic POST retry.

## Provider scope

The change is provider-neutral and applies to every supported provider through `RepositoryApiClient`. No provider client or webhook route changes.

## Tests

- comment-write exception propagates from `handleBotCommand`;
- `ReviewWorkflow` exposes the failure instead of returning success;
- the orchestrator exception-to-`FAILED` regression remains green;
- `mvn --batch-mode clean verify`.

Fixes #274

## AI involvement

AI assistance was used to:
- inspect the relevant repository and cross-provider code paths;
- compare webhook behavior across the supported Git providers;
- draft the tests and implementation;
- identify edge cases and prepare the pull request description.

The contributor:
- provided the original failure context and expected behavior;
- selected and approved the final scope and design;
- retained responsibility for the submitted changes.

The changes were validated with focused tests and the full test suite before
submission.

### DIFF
--- a/src/main/java/org/remus/giteabot/review/CodeReviewService.java
+++ b/src/main/java/org/remus/giteabot/review/CodeReviewService.java
@@ -215,6 +215,13 @@ public class CodeReviewService {
         } catch (Exception e) {
             log.error("Failed to handle bot command for comment #{} on PR #{} in {}/{}: {}",
                     commentId, prNumber, owner, repo, e.getMessage(), e);
+            if (e instanceof RuntimeException runtimeException) {
+                throw runtimeException;
+            }
+            throw new IllegalStateException(
+                    "Failed to handle bot command for comment #" + commentId
+                            + " on PR #" + prNumber,
+                    e);
         }
     }
 
--- a/src/test/java/org/remus/giteabot/prworkflow/review/ReviewWorkflowTest.java
+++ b/src/test/java/org/remus/giteabot/prworkflow/review/ReviewWorkflowTest.java
@@ -27,6 +27,7 @@ import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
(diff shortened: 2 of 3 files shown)


max-california (approved):
## 🤖 Agentic Code Review

Summary: The change looks correct and aligned with the existing workflow/orchestrator design. I didn’t find any blocking issues in the implementation or the added tests.

## What looks good

- `CodeReviewService.handleBotCommand(...)` now preserves the existing detailed error log but no longer falsely reports success when `postPullRequestComment(...)` fails.
- The new behavior matches how `PrWorkflowOrchestrator` is already designed to work: runtime exceptions are persisted as `FAILED` and then rethrown for higher-level bot error handling.
- The tests cover both layers that matter:
  - service-level propagation (`CodeReviewServiceTest`)
  - workflow-level non-swallowing (`ReviewWorkflowTest`)

## Notes

- The new `IllegalStateException` wrapping path for checked exceptions is sensible, although in practice most failures here are likely runtime exceptions already.
- I also verified that `handleBotCommand(...)` still intentionally swallows only the reaction failure, which seems appropriate since that is non-critical acknowledgement behavior.

## Minor suggestion

- `handleBotCommand(...)` stores the user/assistant messages in the session before writing the PR comment. That means a failed comment write will still advance session state. This PR description explicitly accepts orchestrator-level failure handling and no retry, so this is not a blocker, but it may be worth revisiting separately if you want retries or replayability without duplicated/misaligned conversation history.

---

tmseidel (approved):

PR #282, opened 2026-07-18
release/1.17.0

tmseidel:


max-california (approved):
## 🤖 Agentic Code Review

Summary: this release looks solid overall. The new `runOnPrUpdate` flag is wired through persistence/UI and the webhook routing changes for GitHub/Gitea are covered by tests. The `CodeReviewService` change also fixes a real correctness gap by giving follow-up bot commands the latest PR diff context.

I did not find any blocking issues.

## Findings

### Minor / non-blocking

- **Behavior change in `handleBotCommand` on diff-fetch failure is intentional but worth noting**
  - `src/main/java/org/remus/giteabot/review/CodeReviewService.java:178-181,223-232`
  - Follow-up bot commands now stop if the latest diff cannot be fetched, whereas previously existing sessions could continue without re-fetching the diff.
  - This seems consistent with the new goal of grounding responses in the current PR state, and the added tests cover it. Just calling it out because it is a user-visible change in failure mode.

- **Reopened PRs now reuse the same gate as opened PRs**
  - `src/main/java/org/remus/giteabot/github/GitHubWebhookHandler.java:79-82`
  - `src/main/java/org/remus/giteabot/gitea/GiteaWebhookHandler.java:298-302`
  - This looks correct and the tests are good. No issue here; just worth mentioning in release notes since it expands when reviews are triggered.

## Overall

- `runOnPrUpdate` migration/entity/template changes are coherent.
- GitHub/Gitea synchronized/reopened routing is correctly tested.
- Error propagation from bot-command handling into the workflow layer is now better defined and tested.
- I don’t see correctness, security, or maintainability problems that should block merge.

---

PR #292, opened 2026-07-19
feature/audit-logs

### LINKED ISSUE #287: Audit-Trail
Originates from #286

## Quality assessment

The issue describes a valid need, and repository context shows there is already partial audit-related persistence for AI usage plus append-only PR workflow steps. The missing piece is a coherent, system-level audit trail for workflow and moderation events. Critical questions remain around event scope and whether raw LLM content may be stored, so the issue is not yet fully implementation-ready without recording those as open decisions.

## Summary
Add a tamper-evident audit trail for PR workflow activity and related review actions.

## Problem
The system currently persists some operational history, but it does not provide a single, structured, append-only audit trail for PR workflow execution and related decision points.

Existing persistence appears to cover only parts of the story:
- `pr_workflow_runs` stores run-level status.
- `pr_workflow_steps` stores append-only workflow progress/error steps.
- `ai_usage_log` stores AI token usage per session.

This is not sufficient for answering questions such as:
- What happened during a workflow run, in exact order?
- Which bot and project were involved?
- Which AI interaction or system action led to a decision?
- Was a gate overridden or a finding suppressed, and by whom/what?
- Has the stored history been modified after the fact?

## Goal
Introduce a structured, append-only audit trail for PR workflows and related review actions, where each audit record is chained with a hash so later tampering is detectable.

## Scope
Create audit events for PR workflow execution and related domain actions, including the examples called out in the original request:
- `ReviewCompleted`
- `GateOverridden`
- `FindingSuppressed`
- `PullRequestMerged`

In addition, each PR workflow step should be represented in the audit trail so the full workflow can be reconstructed from audit events.

### 1. Append-only audit storage
- Add a dedicated persistence model for audit events.
- Audit records must be insert-only.
- Application code must not update or delete existing audit rows as part of normal operation.
- The audit trail must be queryable by repository/project, pull request, workflow run, and bot.

### 2. Tamper-evident hash chain
- Each audit event must contain a hash that makes post-hoc modification detectable.
- The design should support chaining events in order, for example by including the previous event hash in the next event's hash input.
- Hash generation must be deterministic for the stored event payload.
- Event ordering used for hashing must be unambiguous.

### 3. Event coverage
At minimum, the audit trail must support events for:
- PR workflow run started
- PR workflow step appended/completed/failed
- PR workflow run completed/cancelled/failed
- Review completed
- Gate overridden
- Finding suppressed
- Pull request merged

If some example event types do not yet exist in the current product flow, the implementation should:
- either introduce audit emission where the action already exists,
- or document those event types as deferred if the underlying feature is not yet implemented.

### 4. Event payload
Each audit event should store enough structured metadata to explain what happened.

At minimum, include where available:
- event id
- event type
- event timestamp
- hash
- previous hash
- bot identifier
- repository/project identifier
- pull request identifier/number
- workflow run id, if applicable
- workflow key, if applicable
- step name/order/status, if applicable
- actor/source that triggered the event (`system`, bot, user, webhook, etc.)
- correlation/session id, if applicable
- reason/cause metadata
- references to related AI interactions, if applicable
- event payload as structured JSON or equivalent

### 5. AI decision traceability
Where an audit event is caused by LLM output or an AI-backed decision, the event should include enough information to correlate the event with the relevant AI interaction.

Possible ways to satisfy this include:
- storing the existing AI audit session id,
- storing a foreign key/reference to AI usage/error records,
- storing prompt/response hashes,
- or storing a sanitized snapshot/reference of the relevant LLM output.

The implementation must not duplicate large payloads unnecessarily if correlation by reference is sufficient.

### 6. Failure safety
- Audit logging should not silently corrupt the main workflow state.
- If an audit write fails, behavior must be explicit and documented.
- Prefer preserving core workflow execution while surfacing audit failures clearly in logs/metrics, unless compliance requirements dictate fail-closed behavior.

## Non-goals
- Building a full end-user audit UI is not required unless it is needed for basic verification/testing.
- Replacing existing `pr_workflow_runs`, `pr_workflow_steps`, or `ai_usage_log` tables is not required in this issue.
- Backfilling historical events is not required unless separately specified.

## Suggested implementation direction
Use the existing workflow/run/session concepts rather than replacing them.

Likely approach:
- Keep `pr_workflow_runs` and `pr_workflow_steps` as operational tables.
- Add a new audit event table for durable append-only event records.
- Emit audit events from PR workflow orchestration and any existing moderation/review action paths.
- Reuse the existing AI audit/session correlation where possible.
- Represent event-specific details in structured payload fields rather than creating a separate table per event type.

## Acceptance Criteria
- A new PR workflow run creates audit events for run start, each emitted workflow step, and terminal outcome.
- Stored audit events are append-only in application behavior.
- Each event stores its own hash and the previous event hash, or another equivalent chain mechanism that makes tampering detectable.
- Given a workflow run, developers can reconstruct the ordered event history from the audit table.
- Audit events can be filtered by bot, repository/project, pull request, and workflow run.
- At least the following domain events are supported where the underlying product action exists: `ReviewCompleted`, `GateOverridden`, `FindingSuppressed`, `PullRequestMerged`.
- For events caused by AI output, the stored audit event can be correlated to the relevant AI interaction/session.
- Automated tests verify:
  - event creation for workflow lifecycle events,
  - append-only behavior at the application layer,
  - deterministic hash generation,
  - correct hash chaining/order,
  - correlation to workflow run/step context,
  - correlation to AI audit/session context where applicable.

## Testing Notes
Add tests for:
- persistence of ordered audit events for a normal workflow run,
- persistence when a workflow step fails,
- persistence when a run is cancelled/replaced by a newer run,
- deterministic hashing of identical event payloads,
- broken-chain detection logic, if verification logic is included,
- event emission for moderation/review actions that already exist in the product.

## Open Questions
- Scope for v1: should this issue cover only PR workflow lifecycle events, or also all moderation/review actions across the product?
- Can full LLM prompts/responses be stored in the audit trail, or should only references, hashes, or sanitized payloads be persisted due to privacy/cost/size concerns?
- What exact actor model is needed for override/suppression events: human user id, bot id, system process, or all of the above?
- Is hash verification only a backend/testing concern for now, or must there be an admin/API surface to validate a chain?
- Are `GateOverridden` and `FindingSuppressed` already implemented concepts in the codebase, or should their audit support be deferred until those actions exist?
- Should the hash chain be global, per pull request, or per workflow run?

## Assumptions

- Existing `pr_workflow_steps` are append-only operational records but do not yet satisfy the requested full audit-trail requirement.
- Existing `ai_usage_log` and AI audit session ids can likely be reused for correlation instead of storing full duplicate AI payloads.
- Some example events from the original issue may refer to product actions that are not yet fully implemented in the current codebase.

## Open questions

- Should v1 include only PR workflow lifecycle events, or also broader review/moderation events outside workflow execution?
- May full LLM prompts/responses be persisted, or should the audit trail store only references, hashes, or sanitized excerpts?
- Should the tamper-evident hash chain be scoped globally, per repository/PR, or per workflow run?

---

tmseidel:


max-california (changes requested):
## 🤖 Agentic Code Review

## Summary

Nice addition overall: the new audit domain, persistence, retention job, and orchestration hooks are coherent and mostly well covered by tests. I did find one blocking correctness issue around the hash-chain behavior, plus one significant observability gap in tool-call auditing.

## Findings

### Blocking

- **Hash chaining is effectively disabled for most emitted events because `runId` is often not set, but the implementation only chains by `runId`.**  
  In `PrAuditEventService.resolvePreviousHash(...)`, the previous hash is looked up only by `runId`. But several newly added tests and call sites emit events without a run id:
  - `src/test/java/org/remus/giteabot/audit/PrAuditEventServiceTest.java` (`shouldChainHashesAcrossEvents`)
  - `src/test/java/org/remus/giteabot/audit/PrAuditEventToolCallTest.java` (`shouldChainToolCallsInRun`)
  - `src/main/java/org/remus/giteabot/audit/PrAuditEventService.java` also permits `runId == null` in `emitToolCall(...)`
  
  When `runId` is null, `resolvePreviousHash` returns null, so every event becomes a new chain genesis. That contradicts the “tamper-evident audit trail” intent and means the tests named as chaining tests do not actually assert chaining.

  **Suggested fix:** either:
  1. make `runId` mandatory for all events that should belong to a chain, or  
  2. add a fallback chain key, e.g. `(botId, repoOwner, repoName, prNumber)` when `runId` is null, and test `previousHash` explicitly.

### Non-blocking

- **Failed tool calls are not audited at all.**  
  In `src/main/java/org/remus/giteabot/agent/loop/AgentLoop.java:171-190`, audit emission only happens for `ContinueWithToolResults`, and every emitted record is hard-coded with `success = true`. If a tool execution fails and the strategy surfaces that as an exception, or packages an error result differently, the audit log will miss the failure entirely or mislabel it as success.

  Given the feature’s purpose, failed tool invocations are often the most important ones to retain.

  **Suggested fix:** carry per-tool success/failure from strategy/tool execution into `ToolCallRecord`, and emit audit rows for both successful and failed executions.

- **The recorded tool-call duration is per-step, not per-tool, which can be misleading when multiple tools are executed in one round.**  
  In `AgentLoop`, `stepDurationMs` is measured once around the whole `strategy.step(...)` call and then applied to every tool call in that round. If the assistant asks for multiple tools, each event will report the full round duration.

  That may be acceptable if documented as “round duration”, but the field is named `durationMs`, which strongly implies per-event runtime.

  **Suggested fix:** either rename/document the metric in payload/schema, or collect actual per-tool timing in the execution layer.

- **Tests for chaining don’t verify the chain.**  
  `PrAuditEventServiceTest.shouldChainHashesAcrossEvents` and `PrAuditEventToolCallTest.shouldChainToolCallsInRun` only assert row count. They would currently pass even though `previousHash` is null for both rows.

  **Suggested fix:** assert that the second event’s `previousHash` equals the first event’s `hash`.

## Verdict

Good feature, but the chain-key issue undermines the core tamper-evidence claim and should be fixed before merge.

---

max-california (dismissed):
## 🤖 Agentic Code Review

Summary: the audit logging feature is a solid addition overall, but I found one blocking correctness issue in how tool-call audit events are emitted. The current implementation can silently drop tool-call records for failed tool executions, which makes the audit trail incomplete exactly in the cases where it is most valuable.

## Blocking

### 1. Failed tool calls are never audited
In `src/main/java/org/remus/giteabot/agent/loop/AgentLoop.java` around the new audit hook:

```java
if (ctx.auditToolCallConsumer() != null
        && decision instanceof StepDecision.ContinueWithToolResults(...)) {
    ...
    ctx.auditToolCallConsumer().accept(new AgentRunContext.ToolCallRecord(
        ...
        true,
        ...
```

This only emits audit records when `strategy.step(...)` returns `ContinueWithToolResults`, and it hardcodes `success=true`.

That means any tool failure path that causes `strategy.step(...)` to throw or to produce a non-`ContinueWithToolResults` decision leaves no audit event behind for the attempted tool call. Even within `ContinueWithToolResults`, the persisted audit record cannot distinguish successful and failed tool executions because `success` is always `true`.

Why this matters:
- The PR introduces audit logs specifically for traceability/tamper evidence.
- Failed tool executions are often the most important events to retain.
- The schema and service already model `success`, but the loop never records `false`.

Suggested fix:
- Emit audit events from the place that actually knows per-tool execution outcome, rather than inferring success from the outer decision type.
- At minimum, thread success/failure through `ToolCallResult` or another per-call structure and use that instead of hardcoding `true`.
- Add a test covering a failed tool execution and asserting a `TOOL_CALL_EXECUTED` record is still written with `success=false`.

## Non-blocking

### 2. Hash chaining is race-prone under concurrent inserts
`src/main/java/org/remus/giteabot/audit/PrAuditEventService.java`

`record()` does a read of the latest event and then inserts a new one with `previousHash` based on that read. With concurrent writes for the same run/PR, two transactions can resolve the same predecessor and create sibling events sharing one `previousHash`, which breaks the intended strict linear chain.

This may be acceptable if “best-effort tamper evidence” is enough, but the docs/comments describe a stronger property. Consider documenting the weaker guarantee or serializing inserts per chain key if you need a true chain.

### 3. `toJson()` silently returning `null` weakens auditability
Also in `PrAuditEventService`:

```java
catch (Exception e) {
    return null;
}
```

If serialization fails, the event is still recorded but loses its payload without any signal. Logging the failure would make diagnostics much easier.

---

---

PR #300, opened 2026-07-26
deploy(docker): removed the hardcoded spring profile to let it run with default-settings and internal h2 database

### LINKED ISSUE #299: Simplify creating a test-instance of the app
Currently you need to checkout the repo and start the app with docker-compose:
```bash
git clone https://github.com/tmseidel/ai-git-bot.git && cd ai-git-bot && docker compose up -d
```

This shouldn't be neccessary, for test-purposes the internal h2 database can be used with the default settings, therefore a simple docker run should be sufficient:

```bash
docker run -p 8080:8080 tmseidel/ai-git-bot:latest
```

tmseidel:


max-california (changes requested):
## 🤖 Agentic Code Review

Summary: the Dockerfile change itself is technically safe, but it changes the image’s default runtime behavior from PostgreSQL-backed `application-docker.properties` to the base `application.properties` profile. That aligns with the PR title, but it also means the image no longer matches the production-oriented defaults implied elsewhere in the repo unless operators explicitly set env vars. I found one maintainability/documentation gap to address.

## MEDIUM

1. **Default image behavior now diverges from the documented Compose/deployment path**
   - **File:** `Dockerfile:162`
   - Removing `--spring.profiles.active=docker` makes the container start with `application.properties`, which defaults to:
     - H2 file DB (`jdbc:h2:file:./data/giteabot`)
     - `prompts.dir=prompts`
     - no `app.public-url` default
   - Meanwhile `application-docker.properties` and `docker-compose.yml` still reflect the Docker/Postgres deployment model (`SPRING_PROFILES_ACTIVE: docker`, Postgres defaults, `/app/prompts`, `APP_PUBLIC_URL`, etc.).
   - This is not necessarily wrong, but it creates two competing “normal” container modes: the image’s default is now local/H2, while the repo’s main Compose path is still Docker/Postgres. That can confuse operators and makes the dedicated `application-docker.properties` easier to forget or drift.
   - **Suggestion:** either:
     - update the Docker/README guidance to clearly state that the image now defaults to H2 unless `SPRING_PROFILES_ACTIVE=docker` or equivalent datasource env vars are supplied, or
     - if the intent is to fully retire the docker profile, follow up by consolidating/removing `application-docker.properties` and the now-redundant `SPRING_PROFILES_ACTIVE: docker` in Compose.

## LOW

1. **Check prompt path assumptions for the non-docker default**
   - **Files:** `Dockerfile:150`, `src/main/resources/application.properties:4`
   - The runtime image copies prompts into `/app/prompts`, and the container’s working directory is `/app`, so `prompts.dir=prompts` should resolve correctly today. Still, this change makes the image rely on the working directory for prompt discovery, whereas the docker profile pinned `/app/prompts` explicitly.
   - **Suggestion:** consider setting `PROMPTS_DIR=/app/prompts` in the image environment to make prompt resolution independent of working-directory changes.

---

max-california (approved):
## 🤖 Agentic Code Review

Summary: the Dockerfile change itself is technically valid, but removing the baked-in `docker` profile changes the behavior of the published image in a way that no longer matches the existing compose setup and the `application-docker.properties` defaults. That introduces a real risk of broken container startup or incorrect runtime configuration unless the docs/configs are updated in the same PR.

## MEDIUM

### Runtime behavior of the container image changes materially without updating the surrounding deployment path
By removing `--spring.profiles.active=docker` from `Dockerfile:162-165`, the image now boots with `application.properties` by default instead of `application-docker.properties`.

Today those two profiles differ in important ways:

- `src/main/resources/application.properties:47-52` defaults to an H2 file DB and no `app.public-url`
- `src/main/resources/application-docker.properties:14-20` defaults to PostgreSQL and sets `app.public-url`

At the same time, the checked-in compose deployment still assumes the docker profile:

- `docker-compose.yml:9` explicitly sets `SPRING_PROFILES_ACTIVE: docker`
- `docker-compose.yml` provisions a Postgres sidecar and Postgres env vars

So the image’s default behavior and the repository’s documented deployment path now diverge. Anyone running the image directly as shown in `README.md:521` will silently get H2 instead of the previously intended Postgres-backed docker defaults, and may also miss docker-only settings such as `app.public-url`.

**Suggestion:** either:
1. update the docs/compose/deployment guidance in the same PR to clearly distinguish “plain `docker run` uses default/H2” vs “compose uses `docker` profile/Postgres”, or
2. keep the Dockerfile opinionated and make profile selection explicit only in the runtime manifests that need H2.

## LOW

### `application-docker.properties` may now be effectively dead for the main container image path
After this change, the only remaining checked-in reference to the docker profile I found is `docker-compose.yml:9`. If the intent is to standardize on default settings for containers too, `src/main/resources/application-docker.properties` becomes configuration debt and can drift unnoticed.

**Suggestion:** either remove/consolidate the docker profile, or document why both config sets must continue to exist.

---

PR #318, opened 2026-08-03
feat(docker): add linux/arm64 support to the published image

almnorth:
Summary

The published image is currently `linux/amd64` only, so anyone on Apple Silicon, AWS Graviton or a 64-bit Raspberry Pi has no native image to pull. This adds `linux/arm64` and publishes both as a single multi-arch manifest, so `docker pull tmseidel/ai-git-bot:latest` keeps working unchanged and just resolves to the right variant.

I audited every component the runtime stage installs against ARM64. **Only k6 has no ARM64 path** — everything else already resolves per-architecture and needed no change:

| Component | `arm64` | Notes |
| --- | --- | --- |
| `eclipse-temurin:21-jdk-alpine` / `:21-jre-noble` | ✅   | both are multi-arch manifests |
| NodeSource `node_22.x` | ✅   | feed advertises `amd64 arm64 armhf` |
| `dotnet-sdk-10.0` | ✅   | see note below |
| `golang-go`, `gcc`/`g++`, `ruby`, `python3`, `universal-ctags` | ✅   | Ubuntu Noble ports |
| Playwright 1.60 + Chromium | ✅   | Playwright ships `ubuntu24.04-arm64` builds |
| Cypress 15 | ✅   | `linux-arm64` binary exists (caveat below) |
| `rustup` | ✅   | aarch64 host triple |
| **k6 (`dl.k6.io/deb`)** | ❌   | the APT repo publishes only `amd64` and `i386` |

Incidental finding while checking this: `dotnet-sdk-10.0` does **not** actually come from the Microsoft feed. `packages.microsoft.com/ubuntu/24.04/prod` carries almost no .NET packages for Noble (`dotnet-apphost-pack-6.0` and nothing else). The SDK resolves from Ubuntu's own `noble-updates/main`, which has it for both `amd64` and `arm64`. I've left the Microsoft repo step alone since it's out of scope here, but it may be dead weight worth removing separately.

### `Dockerfile`

- **Build stage pinned to `--platform=$BUILDPLATFORM`.** The artifact is a plain JVM jar and is architecture independent, so there's no reason to run Maven under QEMU when cross-building. On a native arm64 builder `$BUILDPLATFORM` already *is* `linux/arm64`, so this is a no-op there.
- **k6 install is now `TARGETARCH`-aware.** `amd64` keeps the existing APT path byte-for-byte; `arm64` installs the official release tarball from GitHub, pinned by a new `K6_VERSION` build arg (defaulted to `2.0.0`, which is what the APT repo currently serves for amd64, so the two images ship the same k6). An unrecognised `TARGETARCH` now fails the build loudly instead of silently producing an image with no k6.
- Comments added noting the arm64 behaviour of Playwright/Cypress.

I deliberately **did not** switch amd64 over to the tarball as well. It would collapse this into one code path and drop the flaky `gpg --keyserver hkp://keyserver.ubuntu.com:80` step, but it changes how the currently-shipping architecture is built, which felt like the wrong thing to bundle into this PR. Happy to do it as a follow-up if you'd prefer the single path.

### `.github/workflows/docker-publish.yml`

Restructured into `prepare` → `build` (matrix) → `merge`:

- **Each architecture builds natively** on `ubuntu-latest` / `ubuntu-24.04-arm`, pushes **by digest only** (no tags), and `merge` combines the digests into one manifest list published under the real tags.
- **Why not just `platforms: linux/amd64,linux/arm64` + QEMU?** This image installs the .NET SDK, Go, Rust, Playwright and Cypress. Under emulation that's a multi-hour build and a realistic risk of hitting the 6h job limit. Native arm64 runners are free for public repos, so the matrix costs nothing and finishes in roughly the same wall-clock time as today.
- **GHA cache is scoped per architecture** (`scope=amd64` / `scope=arm64`) so the two builds don't evict each other's layers.
- **The pom version is resolved once** in a shared `prepare` job and passed to both downstream jobs. The arm64 runner image ships a reduced toolset and Maven isn't guaranteed to be on it, so I didn't want `mvn help:evaluate` running there.
- **OCI labels are preserved** — `docker/metadata-action` still runs in the build job and its labels are passed to `build-push-action`, so the published image keeps the same `org.opencontainers.image.*` metadata it has today.

### Docs

- `doc/DEPLOYMENT.md`: new **Architectures** section covering both variants, the two arm64 caveats, and how to build the manifest locally.
- `doc/DOCKERHUB_README.md`: supported-architectures table.
- `README.md`: one-line note under the quick-start `docker run`.
- `CHANGELOG.md`: entry under `[Unreleased] → Added`.
- Also corrected a stale line in `doc/DEPLOYMENT.md` that described the runtime base as `eclipse-temurin:21-jre-alpine` — it has been `:21-jre-noble` since the Noble migration. It sits in the block I was editing and would have directly contradicted the new platform note.

## Caveat worth knowing: Cypress on arm64

Cypress publishes a `linux-arm64` binary, but **no Arm builds of Chrome or Firefox** — on arm64 it can only drive its bundled Electron browser. Playwright runs Chromium on both architectures. So for E2E workflows that need a real Chromium on an Arm host, the Playwright runner is the one to use. This is documented in both the Dockerfile comment and `doc/DEPLOYMENT.md`.

## Verification

**Both images were built end-to-end from this branch** — `amd64` natively, `arm64` under QEMU on an x86-64 host. The arm64 build completing at all means every emulated `RUN` step succeeded, including `dotnet --info`, `playwright install chromium`, `cypress install` and the `rustup` bootstrap. Toolchain reported by each built image:

|     | `amd64` | `arm64` |
| --- | --- | --- |
| `uname -m` | `x86_64` | `aarch64` |
| k6  | `v2.0.0` (`linux/amd64`) | `v2.0.0` (`linux/arm64`) |
| `command -v k6` | `/usr/bin/k6` (APT) | `/usr/local/bin/k6` (tarball) |
| Java | `21.0.11` | `21.0.11` |
| Node | `v22.23.2` | `v22.23.2` |
| .NET | `10.0.110` | `10.0.110` |
| Go  | `go1.22.2 linux/amd64` | `go1.22.2 linux/arm64` |
| Playwright | `1.60.0` | `1.60.0` |
| Cypress | `15.19.0` | `15.19.0` |

Full parity — identical versions on both architectures, and the two k6 install paths land the same version. The differing `command -v k6` is the one intended divergence this PR introduces.

- `docker buildx build --check` passes with no warnings; the workflow YAML parses.
- Availability of every arm64 package above was verified directly against the upstream APT indices / release assets rather than assumed.

## Reviewer notes

1. **The multi-arch CI path itself is unexercised.** Both architectures build correctly, but the `prepare → build → merge` digest/manifest flow can only really run on a push to `main` with the Docker Hub secrets present. That's the one part worth watching on the first run after merge.
2. **`ubuntu-24.04-arm` runners are free for public repos.** `tmseidel/ai-git-bot` is public so this is fine, but the workflow would fail on a private fork.
3. **`K6_VERSION` will drift.** amd64 tracks whatever the APT repo serves as stable while arm64 stays pinned. They match today at `2.0.0`; it'll want a bump when the repo moves on. (Switching amd64 to the tarball too — see above — would make this self-correcting.)
4. The local arm64 verification above ran under emulation. CI builds natively on `ubuntu-24.04-arm`, which is strictly the easier case.


max-california (approved):
## 🤖 Agentic Code Review

Summary: the Dockerfile changes look sound and the workflow restructuring is close, but I found one correctness issue in the publish job that can prevent the manifest from being created on first release.

## BLOCKER

1. **Manifest creation can fail because the downloaded digest artifacts are not guaranteed to be in `/tmp/digests` root**
   - **File:** `.github/workflows/docker-publish.yml`
   - **Lines:** 111–153
   - The `merge` job downloads artifacts with `actions/download-artifact@v4` and then runs:
     ```sh
     working-directory: /tmp/digests
     docker buildx imagetools create ... $(printf '${{ env.IMAGE_NAME }}@sha256:%s ' *)
     ```
     This assumes the downloaded files end up directly under `/tmp/digests` and that `*` expands to bare digest filenames. In practice, artifact download layout is easy to get wrong across action versions/options, and if files land in subdirectories or `*` does not match, the command will build invalid image references (or literally pass `*`), causing the publish step to fail.
   - Please make the digest enumeration explicit and robust, e.g. by using `find /tmp/digests -type f -printf ...` or by listing the downloaded tree before invoking `imagetools create` and consuming only actual files.

## LOW

1. **Local multi-arch build example does not actually produce a locally loadable/tagged image by default**
   - **File:** `doc/DEPLOYMENT.md`
   - **Lines:** 176–177
   - `docker buildx build --platform linux/amd64,linux/arm64 -t ai-git-bot:local .` without `--push` or `--load` can confuse readers: the default `docker-container` builder won’t make a usable local image from that command alone.
   - Suggest clarifying that this command is for building with Buildx, and add `--push` (for publishing) or explain that a multi-platform local test typically needs a registry/`--output`.

---

tmseidel:
@almnorth Thanks for your contribution. I have to test the deployment before we can merge, but it looks good to me so far...

almnorth:
Ok, I have not yet tested on arm because I don't have the pipeline setup but I will try to use it in production once the image is there, I can still do some fixes if needed

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

tmseidel:
The following tests were executed:
1. Complete Build on a Raspberry PI + Execution with PostgresSQL setup -> ✅
2. Tested the extended workflow action with 'dry-run' flags before publishing (https://github.com/tmseidel/ai-git-bot/actions/runs/31252380192) -> ✅
3. k6 is flagged as experimental in the application, live-test is okay.

Overall, looks good to me, can be merged. 🚀

PR #319, opened 2026-08-07
security hardening: webhook signatures, credential-free clones, sandboxed execution

CaeruleusAqua:
## What was done and why?

A structured security review (static code audit + live pentest of a production deployment) surfaced several issues in the areas of webhook authentication, secret handling, untrusted code execution and encryption at rest. This PR hardens all of them in one cohesive changeset. 

### Inbound webhook signature verification (High)
Provider webhooks were authenticated solely by the URL path secret - no `X-Hub-Signature-256` / `X-Gitea-Signature` / `X-Gitlab-Token` verification existed anywhere. This PR adds optional per-bot **webhook signing secrets**: when configured, the raw request body is verified per provider (HMAC-SHA256 for GitHub/Gitea/Bitbucket, token equality for GitLab, constant-time comparison, `401` on mismatch). The secret is stored AES-GCM-encrypted, editable in the bot form, and the setup docs now recommend it for public repositories.

### Git tokens out of clone URLs (High)
Clone URLs embedded the integration token (`https://oauth2:<token>@host/...`), which persisted it in `.git/config` (readable by agent file tools) and leaked it into git error output - which is logged and posted verbatim into public PR/issue comments. Cloning/fetching/pushing now uses a git credential-store file placed **outside** the workspace, git output is redacted defensively, and `.git` is excluded from all agent file/search/tree tools.

### Sandboxed execution of untrusted build/test commands (Critical)
Unit-test/E2E workflows build and execute PR-controlled code on the host, and agent tool-calling runs LLM-chosen commands (`python3`, `mvn`, .) - with child processes inheriting the full JVM environment (DB password, encryption key, OAuth secrets). This PR:
- runs validation tools, generated test suites and PR-controlled `npm install` in an **ephemeral Docker container** when `agent.sandbox.enabled=true` and a Docker host is reachable (resource limits, `--cap-drop ALL`, `no-new-privileges`, read-only rootfs, scrubbed env, uid 1000, network off by default), with a scrubbed-environment direct execution fallback;
- scrubs the environment of **every** spawned process to a minimal allowlist regardless of sandboxing.

### Symlink-safe workspace path guards (High)
Agent write tools only checked the leaf for symlinks; directory symlinks committed in a PR allowed writes outside the workspace. Path guards now resolve the nearest existing ancestor against its real path.

### Smaller hardening items
- `/api/**` request-body size cap (default 2 MiB ? `413`; chunked bodies bounded too) - closes an unauthenticated parse-before-auth DoS.
- `giteabot.security.login-method` fails closed: unknown values abort startup instead of silently dropping the web security chain.
- Encryption at rest for `mcp_configurations.json_content` (MCP tokens/headers) and `event_hook_endpoint.custom_headers`.
- Docs: new "Sandboxed Execution" section in `DEPLOYMENT.md`.

### Verified unaffected
No SQL injection (JPA derived queries), no shell-string invocation (array-form `ProcessBuilder` everywhere), git option injection neutralized (`--`, `check-ref-format`), CSRF intact on the web UI, actuator exposure unchanged (health/info only).


tmseidel:
@CaeruleusAqua Julian, thanks for your contribution, this looks awesome. Some notes to get this merged as soon as possible:
1. Our target branch for contributions is `develop`, I've already changed it but there are some conflicts to resolve
2. The changes are huge, the review of such big changes needs a long time. Can you please split this PR into 5 Pull-Requests (for each item 1 PR) - so that we have a smaller scope for each PR; we can test the single changes more independently - This would give us the opportunity to be much faster in reviewing external contributions and get your changes into our codebase.
3. I noticed that your formatter removed unicode-characters like `—` or `→`. This needs to be reverted
4. The SQL-Scripts for data-migration are not idempotent
5. For transparency every external contribution should have a small section about how AI was used creating these code-changes.

Thanks again for your contribution 🚀

PR #320, opened 2026-08-07
verify inbound webhook signatures

CaeruleusAqua:
## What was done and why?

- Adds optional per-bot webhook signing secrets and verifies GitHub, Gitea, GitLab, and Bitbucket webhook authentication before processing callbacks.
- Stores signing secrets with the existing encryption service and keeps provider setup documentation alongside the implementation.
- Enforces the configured API request-body limit before webhook processing.
- Makes the V40 signing-secret schema migrations idempotent for H2 and PostgreSQL.
- Adds focused provider, migration, and payload-limit regression coverage.

## Testing

- Build and verify CI
- Focused webhook, migration, and payload-limit regression tests
- Git whitespace validation

### AI use

### DIFF
--- a/src/main/java/org/remus/giteabot/admin/BotService.java
+++ b/src/main/java/org/remus/giteabot/admin/BotService.java
@@ -23,6 +23,7 @@ public class BotService {
 
     private final BotRepository botRepository;
     private final BotToolConfigurationRepository botToolConfigurationRepository;
+    private final EncryptionService encryptionService;
 
     @Transactional(readOnly = true)
     public List<Bot> findAll() {
@@ -43,6 +44,18 @@ public class BotService {
         if (bot.getWebhookSecret() == null) {
             bot.setWebhookSecret(UUID.randomUUID().toString());
         }
+        // Encrypt a newly provided webhook signing secret; keep the stored one
+        // when the form field is left blank on update (same pattern as the
+        // git integration token).
+        String signingSecret = bot.getWebhookSigningSecret();
+        if (signingSecret != null && !signingSecret.isBlank()) {
+            bot.setWebhookSigningSecret(encryptionService.encrypt(signingSecret));
+        } else if (bot.getId() != null) {
+            botRepository.findById(bot.getId())
+                    .ifPresent(existing -> bot.setWebhookSigningSecret(existing.getWebhookSigningSecret()));
+        } else {
+            bot.setWebhookSigningSecret(null);
+        }
         if (bot.getToolConfiguration() == null) {
             // Defensive fallback for callers that bypass the BotController
             // (integration tests, scripts): assign the Default tool
@@ -62,6 +75,19 @@ public class BotService {
                             + "explicitly before saving the bot.");
         }
         return botRepository.save(bot);
+    }
+
+    /**
+     * Returns the bot's decrypted webhook signing secret, or {@code null} when
+     * none is configured (signature verification then stays disabled).
+     */
+    @Transactional(readOnly = true)
+    public String getDecryptedWebhookSigningSecret(Bot bot) {
+        String secret = bot.getWebhookSigningSecret();
+        if (secret == null || secret.isBlank()) {
+            return null;
+        }
+        return encryptionService.decrypt(secret);
     }
 
     public void deleteById(Long id) {
--- a/src/main/java/org/remus/giteabot/admin/BotService.java
+++ b/src/main/java/org/remus/giteabot/admin/BotService.java
@@ -41,15 +41,29 @@ public class BotService {
     }
 
     public Bot save(Bot bot) {
+        return save(bot, false);
+    }
+
+    /**
+     * Saves a bot, resolving the webhook signing secret from the form input.
+     *
+     * <p>The secret field is a one-way write: the stored value is never echoed
+     * back into the form. A blank field therefore means "keep the stored
+     * value", while {@code clearSigningSecret} requests explicit removal (the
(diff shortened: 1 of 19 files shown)


max-california (approved):
## 🤖 Agentic Code Review

Summary: the overall direction looks solid — signature verification before JSON parsing and the API payload cap both address real attack surfaces, and the migration/tests cover the happy path well. I found two correctness issues worth fixing before merge.

## BLOCKER

### 1. Clearing an existing webhook signing secret is impossible from the UI/service layer
`src/main/java/org/remus/giteabot/admin/BotService.java:53`

The new save logic treats a blank `webhookSigningSecret` on update as “keep the existing stored value”:

```java
} else if (bot.getId() != null) {
    botRepository.findById(bot.getId())
            .ifPresent(existing -> bot.setWebhookSigningSecret(existing.getWebhookSigningSecret()));
}
```

That matches the placeholder text in `templates/bots/form.html`, but it also means there is now no way to remove a previously configured signing secret. Once enabled, operators cannot revert to unsigned webhooks or rotate by first clearing and reapplying. This is a functional regression for bot administration.

Please add an explicit “clear secret” path (checkbox/button/flag) and wire it through `BotService.save`, or otherwise distinguish “field omitted, keep existing” from “user intentionally cleared it”.

## MEDIUM

### 2. `ApiPayloadSizeLimitFilter` can throw before your controller and likely returns 500 instead of 413 for oversized chunked bodies
`src/main/java/org/remus/giteabot/webhook/ApiPayloadSizeLimitFilter.java:109`

For unknown-length/chunked requests, the wrapper throws a plain `IOException` once the limit is exceeded:

```java
if (consumed > maxBytes) {
    throw new IOException("Request body exceeds the configured limit ...");
}
```

In real request handling, that exception will be raised later when Spring reads `@RequestBody byte[] rawBody` in `UnifiedWebhookController`, i.e. outside `doFilterInternal`. Unless there is a global exception mapper for this exact case, clients will typically see a generic 500/400 rather than the intended 413. The current test only asserts that the exception is thrown, not that the HTTP response is correct.

Recommend converting this into an HTTP 413 at the filter boundary, e.g. by catching the wrapper’s limit exception around `filterChain.doFilter(...)`, or by throwing a dedicated exception that a `@ControllerAdvice` maps to `REQUEST_ENTITY_TOO_LARGE`, and adding an integration-style test for the response status.

---

src/main/java/org/remus/giteabot/admin/BotService.java
```diff
@@ -43,6 +44,18 @@ public Bot save(Bot bot) {
         if (bot.getWebhookSecret() == null) {
             bot.setWebhookSecret(UUID.randomUUID().toString());
         }
+        // Encrypt a newly provided webhook signing secret; keep the stored one
+        // when the form field is left blank on update (same pattern as the
+        // git integration token).
+        String signingSecret = bot.getWebhookSigningSecret();
```
tmseidel, line 50: This implementation is a bit different than in the other cases (like in `org.remus.giteabot.admin.AiIntegration#apiKey`) but this looks to me a bit cleaner since all checks are in one layer, instead of dividing them into the Controller-Layer and Service-Layer...nice - I have to adapt the other fields (`org.remus.giteabot.admin.AiIntegration#apiKey` and `org.remus.giteabot.admin.GitIntegration#token`) to the same pattern.

src/main/java/org/remus/giteabot/webhook/UnifiedWebhookController.java
```diff
@@ -57,14 +64,29 @@ public ResponseEntity<String> handleWebhook(
...
+                    String signingSecret = botService.getDecryptedWebhookSigningSecret(bot);
+                    if (!WebhookSignatureVerifier.isValid(
+                            bot.getGitIntegration().getProviderType(), signingSecret, headers, rawBody)) {
+                        log.warn("Invalid webhook signature for bot '{}' from {}",
+                                bot.getName(), headers.getOrDefault("x-forwarded-for", "?"));
+                        return ResponseEntity.status(401).body("invalid signature");
+                    }
+                    Map<String, Object> payload;
+                    try {
+                        payload = objectMapper.readValue(rawBody, Map.class);
```
tmseidel, line 86: To get rid of the typecheck compiler-warning use `payload = objectMapper.readValue(rawBody, new TypeReference<>() {});` instead
```suggestion
                        payload = objectMapper.readValue(rawBody, new TypeReference<>() {});
```

src/main/resources/templates/bots/form.html
```diff
@@ -212,6 +212,20 @@ <h2>
                             </div>
                         </div>
 
+                        <div class="col-12">
```
tmseidel, line 215: Can you add a Clear-Button like the bot-review suggested (see proposal as image) - otherwise this would be a functional regression:

Image

tmseidel (changes requested):
Apart from the two issues in the comments (1. the compiler-warning and 2. the clear-feature) this PR looks good to me. If these two things are fixed, we can merge it
Tests:
Gitea -> ✅
Github -> ✅
(Gitlab and Bitbucket are tested after user-feedback)

src/main/resources/templates/bots/form.html
CaeruleusAqua: I'll try to get this done today...

CaeruleusAqua: Done

tmseidel (approved):
LGTM

PR #322, opened 2026-08-08
Harden untrusted command execution

CaeruleusAqua:
## What was done and why?

- Added pure-Java hardening for untrusted validation and E2E commands: child processes now run with a scrubbed environment (toolchain allowlist, no application secrets), a dedicated Linux process group via `setsid` where available, descendant tracking with forcible termination, and bounded output with UTF-8-safe truncation.
- Applied the hardened execution path to `ToolExecutionService` (validation tools), `WorkspaceProcessRunner` (E2E `pr-test-run`), and `PrTestWorkspaceManager` (scaffold `npm install`).
- `PrTestWorkspaceManager` now rejects a relative `ai-git-bot.e2e.workspace-root` at construction instead of silently resolving it against the process working directory.
- No container, no daemon, no new configuration: the hardening is always on and requires zero operational changes.

## Testing

- Full Docker Maven test suite green (1432 tests, 0 failures)
- Focused tests for environment scrubbing, process-group isolation, descendant termination, timeouts, output bounds, and the absolute workspace-root validation

### AI use

- Implementation and review used OpenCode (Kimi-K3).

### DIFF
--- a/src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java
+++ b/src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java
@@ -4,12 +4,11 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.remus.giteabot.agent.tools.ToolCatalog;
 import org.remus.giteabot.config.AgentConfigProperties;
+import org.remus.giteabot.util.ProcessSupport;
 import org.springframework.stereotype.Service;
 
-import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.LinkOption;
@@ -1283,34 +1282,26 @@ public class ToolExecutionService {
             ProcessBuilder pb = new ProcessBuilder(command);
             pb.directory(workspaceDir.toFile());
             pb.redirectErrorStream(true);
-
-            Process process = pb.start();
-
-            StringBuilder output = new StringBuilder();
-            try (BufferedReader reader = new BufferedReader(
-                    new InputStreamReader(process.getInputStream()))) {
-                String line;
-                while ((line = reader.readLine()) != null) {
-                    output.append(line).append("\n");
-                }
-            }
+            // Untrusted repository code (package-manager scripts, test fixtures) must
+            // never see application secrets such as database or AI provider keys.
+            ProcessSupport.scrubEnvironment(pb);
 
             int timeoutSeconds = agentConfig.getValidation().getToolTimeoutSeconds();
-            boolean finished = process.waitFor(timeoutSeconds, TimeUnit.SECONDS);
+            ProcessSupport.CommandResult result = ProcessSupport.run(pb, timeoutSeconds,
+                    TimeUnit.SECONDS, MAX_TOOL_OUTPUT_CHARS + 1_000);
 
-            if (!finished) {
-                process.destroyForcibly();
+            if (!result.finished()) {
                 return new ToolResult(false, -1, "",
                         "Tool execution timed out after " + timeoutSeconds + " seconds");
             }
 
-            int exitCode = process.exitValue();
+            int exitCode = result.exitCode();
             boolean success = exitCode == 0;
 
             log.info("Tool {} with exit code {}",
                     success ? "succeeded" : "failed", exitCode);
 
-            return new ToolResult(success, exitCode, truncateOutput(output.toString()), "");
+            return new ToolResult(success, exitCode, truncateOutput(result.output()), "");
 
         } catch (IOException e) {
             log.error("Failed to execute tool: {}", e.getMessage());
--- a/src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java
+++ b/src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java
@@ -1287,6 +1287,10 @@ public class ToolExecutionService {
             ProcessSupport.scrubEnvironment(pb);
 
             int timeoutSeconds = agentConfig.getValidation().getToolTimeoutSeconds();
+            // The byte cap is the hard memory bound while draining the pipe; the
+            // character cap below is the presentation limit. The +1_000 headroom
+            // makes the char cap the effective limit for ASCII-dominant output,
+            // while multi-byte-heavy output may hit the byte cap first.
             ProcessSupport.CommandResult result = ProcessSupport.run(pb, timeoutSeconds,
                     TimeUnit.SECONDS, MAX_TOOL_OUTPUT_CHARS + 1_000);
 
--- a/src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java
+++ b/src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java
@@ -1,6 +1,7 @@
 package org.remus.giteabot.agent.validation;
 
 import lombok.extern.slf4j.Slf4j;
+import org.remus.giteabot.util.ProcessSupport;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
@@ -17,7 +18,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -49,11 +49,6 @@ public class WorkspaceService {
     private static final Set<PosixFilePermission> OWNER_FILE_PERMISSIONS = Set.of(
             PosixFilePermission.OWNER_READ,
             PosixFilePermission.OWNER_WRITE);
-    private static final List<String> GIT_ENVIRONMENT_ALLOWLIST = List.of(
-            "PATH", "LANG", "LC_ALL", "LC_CTYPE", "TZ", "TMPDIR", "TMP", "TEMP",
-            "SystemRoot", "windir", "COMSPEC", "PATHEXT",
-            "HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY", "http_proxy", "https_proxy", "no_proxy",
-            "GIT_SSL_CAINFO", "GIT_SSL_CAPATH", "SSL_CERT_FILE", "SSL_CERT_DIR");
     private final Path workspaceBaseDir;
     private final ConcurrentMap<Path, Path> credentialsByWorkspace = new ConcurrentHashMap<>();
 
@@ -534,7 +529,7 @@ public class WorkspaceService {
             ProcessBuilder pb = new ProcessBuilder(gitCommand);
             pb.directory(workDir);
             pb.redirectErrorStream(true);
-            scrubEnvironmentForGit(pb);
+            ProcessSupport.scrubEnvironmentForGit(pb);
             pb.environment().put("GIT_CONFIG_NOSYSTEM", "1");
             pb.environment().put("GIT_CONFIG_GLOBAL", emptyGlobalGitConfig.toString());
 
@@ -579,18 +574,6 @@ public class WorkspaceService {
                     log.warn("Failed to remove empty Git hooks directory {}: {}",
                             disabledHooksDirectory, e.getMessage());
                 }
-            }
-        }
-    }
-
-    /** Keeps credentials and application secrets out of Git subprocesses. */
-    private void scrubEnvironmentForGit(ProcessBuilder processBuilder) {
-        Map<String, String> environment = processBuilder.environment();
-        environment.clear();
-        for (String name : GIT_ENVIRONMENT_ALLOWLIST) {
-            String value = System.getenv(name);
-            if (value != null) {
-                environment.put(name, value);
             }
         }
     }
--- a/src/main/java/org/remus/giteabot/prworkflow/e2e/tools/WorkspaceProcessRunner.java
+++ b/src/main/java/org/remus/giteabot/prworkflow/e2e/tools/WorkspaceProcessRunner.java
@@ -1,5 +1,6 @@
 package org.remus.giteabot.prworkflow.e2e.tools;
 
+import org.remus.giteabot.util.ProcessSupport;
 import org.springframework.stereotype.Component;
 
 import java.io.IOException;
@@ -18,6 +19,10 @@ import java.util.concurrent.TimeUnit;
  * stream because most test frameworks (Playwright, pytest, k6) interleave
  * them, and the agent only needs a single textual blob to reason about the
  * outcome.</p>
+ *
+ * <p>Commands execute with a scrubbed environment (no application secrets)
+ * and the strongest local lifetime control available, see
+ * {@link ProcessSupport}.</p>
  */
 @Component
 public class WorkspaceProcessRunner {
@@ -36,8 +41,8 @@ public class WorkspaceProcessRunner {
      * stdout/stderr (UTF-8) up to {@code maxOutputBytes} bytes and waiting
      * at most {@code timeout} ms before terminating the process.
      *
-     * @param extraEnv environment overrides applied on top of the inherited
-     *                 JVM environment (later entries win). Use this to inject
+     * @param extraEnv environment overrides applied on top of the scrubbed
+     *                 environment (later entries win). Use this to inject
      *                 {@code BASE_URL} for browser tests or
      *                 {@code PLAYWRIGHT_JSON_OUTPUT_NAME} to route the report.
      */
@@ -48,47 +53,19 @@ public class WorkspaceProcessRunner {
         ProcessBuilder pb = new ProcessBuilder(command)
                 .directory(workspace.toFile())
                 .redirectErrorStream(true);
+        ProcessSupport.scrubEnvironment(pb);
         // Make sure CI-style envs do not break the runner with interactive prompts.
-        pb.environment().putIfAbsent("CI", "1");
+        pb.environment().put("CI", "1");
         if (extraEnv != null) {
             for (Map.Entry<String, String> e : extraEnv.entrySet()) {
                 if (e.getKey() == null || e.getValue() == null) continue;
                 pb.environment().put(e.getKey(), e.getValue());
             }
         }
-        Process process = pb.start();
-        StringBuilder out = new StringBuilder();
-        Thread reader = new Thread(() -> {
-            try (var in = process.getInputStream()) {
-                byte[] buf = new byte[8192];
-                int read;
-                while ((read = in.read(buf)) > 0) {
-                    if (out.length() >= maxOutputBytes) {
-                        // Drain so the child does not block on a full pipe.
-                        continue;
-                    }
-                    int spare = Math.max(0, maxOutputBytes - out.length());
-                    int take = Math.min(read, spare);
-                    if (take > 0) {
-                        out.append(new String(buf, 0, take, java.nio.charset.StandardCharsets.UTF_8));
-                    }
-                }
-            } catch (IOException ignored) {
-                // process exited
-            }
-        }, "pr-test-run-reader");
-        reader.setDaemon(true);
-        reader.start();
-
-        boolean finished = process.waitFor(timeoutMs, TimeUnit.MILLISECONDS);
-        if (!finished) {
-            process.destroyForcibly();
-            reader.join(2_000);
-            long durationMs = (System.nanoTime() - start) / 1_000_000L;
-            return new ProcessResult(-1, out.toString(), durationMs, true);
-        }
-        reader.join(2_000);
+        ProcessSupport.CommandResult result = ProcessSupport.run(pb, timeoutMs,
+                TimeUnit.MILLISECONDS, maxOutputBytes);
         long durationMs = (System.nanoTime() - start) / 1_000_000L;
-        return new ProcessResult(process.exitValue(), out.toString(), durationMs, false);
+        return new ProcessResult(result.finished() ? result.exitCode() : -1,
+                result.output(), durationMs, !result.finished());
     }
 }
--- a/src/main/java/org/remus/giteabot/prworkflow/e2e/workspace/PrTestWorkspaceManager.java
+++ b/src/main/java/org/remus/giteabot/prworkflow/e2e/workspace/PrTestWorkspaceManager.java
@@ -2,6 +2,7 @@ package org.remus.giteabot.prworkflow.e2e.workspace;
 
 import lombok.extern.slf4j.Slf4j;
 import org.remus.giteabot.prworkflow.e2e.E2eTestFramework;
+import org.remus.giteabot.util.ProcessSupport;
 import org.remus.giteabot.util.WorkspacePaths;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
@@ -40,6 +41,13 @@ public class PrTestWorkspaceManager {
     private final Path root;
     private final boolean npmInstallEnabled;
 
+    /**
+     * Creates the manager with the configured workspace root.
+     *
+     * @param configuredRoot absolute host-visible root for generated workspaces;
+     *                       a relative value is rejected to avoid silently resolving
+     *                       it against the process working directory
+     */
     public PrTestWorkspaceManager(
             @Value("${ai-git-bot.e2e.workspace-root:#{null}}") String configuredRoot,
             @Value("${ai-git-bot.e2e.npm-install-enabled:true}") boolean npmInstallEnabled) {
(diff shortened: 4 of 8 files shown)


max-california (changes requested):
## 🤖 Agentic Code Review

Summary: the sandboxing/hardening direction looks solid, and the added tests cover a lot of the new behavior. I found two correctness issues where the new logic can silently skip protections or break valid deployments.

## BLOCKER

1. **Sandbox artifact export can be silently truncated before the marker, so artifacts are lost even on successful runs**
   - **File:** `src/main/java/org/remus/giteabot/prworkflow/e2e/tools/WorkspaceProcessRunner.java`
   - **Line:** 65
   - `WorkspaceProcessRunner` calls `sandboxedCommandExecutor.run(..., maxOutputBytes, ...)` with `maxOutputBytes = DEFAULT_RUN_OUTPUT_BYTES` (6 MiB). But the Docker-side protocol appends artifact payloads *after* command output, and `ProcessSupport.waitFor()` truncates the combined stream to `maxOutputBytes`.
   - That means a verbose test run can consume the whole 6 MiB before `__AI_GIT_BOT_ARTIFACTS__` is emitted, so `restoreArtifacts()` never sees the marker and silently restores nothing. The PR description says artifact restore stays inside the sandbox boundary with a bounded return channel, but in practice the return channel competes with normal stdout/stderr.
   - This is especially likely for failing browser runs that emit long stack traces plus screenshots/videos.
   - **Suggestion:** reserve space for the artifact section independently of normal output, e.g. pass `maxOutputBytes + MAX_EXPORTED_ARTIFACT_BYTES + protocol overhead` to the executor, or better yet separate stdout from the artifact channel entirely.

2. **`PrTestWorkspaceManager.allocate()` only rejects network-isolated sandboxes when Docker is currently reachable**
   - **File:** `src/main/java/org/remus/giteabot/prworkflow/e2e/workspace/PrTestWorkspaceManager.java`
   - **Line:** 74
   - `allocate()` blocks E2E workspaces only when `sandboxedCommandExecutor.isNetworkIsolated()` is true. But `isNetworkIsolated()` is implemented as `"none".equalsIgnoreCase(config.getNetwork()) && isSandboxed()`, and `isSandboxed()` probes Docker live.
   - If sandboxing is enabled with `network=none` but Docker is temporarily unavailable, `isNetworkIsolated()` returns false and allocation succeeds. Later `pr-test-run` either:
     - fails closed because Docker is unavailable, or
     - falls back to direct host execution if `fallback-to-direct=true`, which defeats the intent of the E2E-specific validation here.
   - The decision about whether an E2E workflow is compatible with the configured sandbox network should not depend on transient Docker availability.
   - **Suggestion:** make the allocation check key off configuration (`sandbox.enabled && network == none`) rather than `isSandboxed()`, or add a separate config-level predicate for “sandbox network is incompatible with preview access”.

## MEDIUM

1. **Configured workspace base directory is documented as needing to be “absolute”, but the code silently rewrites relative paths**
   - **File:** `src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java`
   - **Line:** 44
   - The constructor comment says workspaces are placed below the configured **absolute** directory, but `Path.of(configuredDir).toAbsolutePath().normalize()` accepts relative values and resolves them against the process cwd.
   - For a sandbox bind-mount root this can easily point somewhere unexpected in production and make Docker mounts fail in non-obvious ways.
   - **Suggestion:** validate `configuredDir` with `Path.of(configuredDir).isAbsolute()` and fail startup or log a clear error if it is relative.

---

CaeruleusAqua:
Addressed in `9b1715d` ("fix(security): reserve artifact channel and harden sandbox config checks"). Thanks for the thorough review — all three points are fixed.

**BLOCKER 1 — artifact export silently truncated before the marker:** Fixed. `WorkspaceProcessRunner` now reserves a dedicated budget for the artifact return channel (`ARTIFACT_CHANNEL_RESERVE_BYTES = MAX_EXPORTED_ARTIFACT_BYTES * 4/3 + 64 KiB`, covering the Base64 expansion and protocol overhead) and passes `maxOutputBytes + reserve` to the executor. The marker and its payload therefore never compete with the caller-visible output budget. After extraction the combined output is capped back, byte-safely (no split multi-byte chars), to `maxOutputBytes`. Regression tests: `run_restoresArtifactsEvenWhenOutputConsumesFullBudget` (a verbose log that fills the whole visible budget still restores the screenshot) and `run_capsCombinedOutputWithoutMarkerToCallerBudget`.

**BLOCKER 2 — allocation only rejected when Docker is currently reachable:** Fixed. `SandboxedCommandExecutor.isNetworkIsolated()` is now a deterministic, config-only check (`enabled && network == none`) with no live Docker probe, so its answer no longer flips with transient Docker outages. `PrTestWorkspaceManager.allocate()` keeps calling it, so an E2E workspace is rejected whenever `network=none` is configured — even with `fallback-to-direct=true` (which the reviewer rightly flagged as falling outside the sandbox boundary) and even while Docker is down. Tests updated accordingly: `isNetworkIsolated_reflectsConfigurationIndependentOfDockerAvailability`, `isNetworkIsolated_allowsPreviewReachableNetworkEvenWhenDockerIsUnavailable`, `allocate_rejectsNetworkNoneEvenWhenDockerIsUnavailable`.

**MEDIUM — configured workspace root silently resolved when relative:** Fixed. The `PrTestWorkspaceManager` constructor now requires an absolute `ai-git-bot.e2e.workspace-root` and throws an `IllegalArgumentException` with a clear message otherwise, instead of silently resolving it against the process working directory (which made Docker bind-mounts fail in non-obvious ways). The javadoc now documents that requirement. Regression test: `rejectsRelativeConfiguredRoot`.

tmseidel:
This is a very huge change, I need a bit time to review 😀

CaeruleusAqua:
> This is a very huge change, I need a bit time to review 😀

The last one was probably the fastest PR I've ever pushed through in an open-source project, so no stress!

tmseidel:
I have major concerns regarding this pull-request (seems to me overengineered). To recap the initial value this pr will bring I have done some notes to make things clearer:

## Threat model — what are we actually protecting?

The commands at risk are build/test invocations (`mvn`, `gradle`, `npm`, `pytest`, `k6`, …)
whose **executable** comes from a server-side allowlist (`ToolCatalog.validationToolNames()`)
but whose **arguments and workspace content** are influenced by the LLM and by untrusted PR
code (package-manager lifecycle scripts, test fixtures, Makefiles). Concrete threats:

| # | Threat | Branch's answer | Docker actually needed? |
|---|--------|-----------------|-------------------------|
| T1 | Exfiltration of app **environment** secrets (DB creds, Gitea/AI API keys are env vars) | `ProcessSupport.scrubEnvironment` allowlist | **No — pure Java, works today, no container involved** |
| T2 | Reading secrets from the **filesystem** (git credential-store files, app config, other workspaces) | Container user + mount masking | Partly — see §3.2; develop already keeps git credentials **outside** the workspace (`WorkspaceService.credentialConfigArgs`) and mounts nothing sensitive |
| T3 | **Network** abuse (exfiltration channel, malicious downloads) | `--network none` | Yes, for a hard guarantee — but `unshare --net` gives the same kernel primitive without a daemon (§3.3) |
| T4 | **Resource** exhaustion (fork bomb, memory, disk fill, runaway CPU) | `--memory/--cpus/--pids-limit`, bounded tmpfs | Timeout + process-group kill (pure Java, already in `ProcessSupport`) covers runaway processes; `prlimit` covers the rest cheaply (§3.3) |
| T5 | **Workspace tampering** (build modifies/deletes files) | tmpfs copy — workspace is immutable | Not needed: workspaces are disposable (agent clones are deleted after the run, E2E scaffolds are throwaway). Default direct mode offers zero protection here and nobody missed it |
| T6 | **Persistence** after the command (background daemons, planted git hooks, modified `.git/config`) | Container death + `.git/config` overlay | Process-group kill (done) + develop's existing hardening: host git commands already force `core.hooksPath=<empty dir>`, `credential.helper=`, `GIT_CONFIG_NOSYSTEM`, a blank global config (`WorkspaceService.runCommand`) |

Two observations fall out of this table:

1. **The highest-value protection (T1) is already pure Java and lives in `ProcessSupport`.**
   The branch's real security win over develop is the env scrubbing — that code never touches
   Docker.
2. **The Docker mount is protecting the least-valuable asset** (a disposable workspace, T5)
   while the genuinely valuable guards for T2/T6 either already exist on develop or don't
   strictly need a container.

### The Docker approach worsens the app's own attack surface

`docker-compose.sandbox.yml` mounts `/var/run/docker.sock` into the application container.
**Anyone who can run a command as the app user — precisely the scenario this branch defends
against — can start a privileged container and own the host.** A sandbox escape stops being
the worst case; the sandbox *plumbing* becomes the worst case. Add the operational price:
a sibling-container path-visibility requirement (`GITEABOT_WORKSPACES_DIR` must resolve
identically for app and daemon), a toolchain image to keep in sync, orphan reaping,
availability probing, and a fallback policy that silently degrades to non-isolated execution.

That is a lot of machinery to defend against `npm test` in a throwaway directory.

---

**TL;DR** The docker-approach seems to me not useful at all, I've already started to evaluate other more lightweight approaches that can be implemented in pure Java. Here are my suggestions:

---
### Layer 1 — pure Java (keep, already built)

Everything in `ProcessSupport` plus the surrounding discipline, none of which needs a
container:

- **No shell, ever.** Commands execute as direct argv (`List<String>` → `ProcessBuilder`).
  There is no `sh -c` anywhere in the replacement design, so there is no shell-injection
  surface and no place for a `WORKSPACE_SETUP_COMMAND`-style script to exist at all.
- **Tool allowlist.** `ToolCatalog.validationToolNames()` already restricts the executable.
  Arguments stay free-form (the LLM needs `-Dtest=…`), which is acceptable because argv
  without a shell cannot escape into command substitution.
- **Environment scrubbing.** `ProcessSupport.scrubEnvironment` replaces the child env with a
  minimal toolchain allowlist (T1). Git transport gets the stricter
  `scrubEnvironmentForGit` variant.
- **Lifetime control.** `setsid` process group + descendant tracking + timeout +
  `destroyForcibly` + bounded output drain (T4 runaway, T6 daemons).
- **Git hygiene.** `WorkspaceService` on develop already neutralizes repo-controlled git
  behavior for host-side commands (`core.hooksPath`, blank global config, disabled credential
  helper, `GIT_CONFIG_NOSYSTEM`) and stores clone credentials **outside** the workspace (T2,
  T6). The branch's `.git/config` overlay was load-bearing only because its base still
  embedded the token in the clone URL; on current develop that is gone.

This layer alone is a strict improvement over develop and requires **zero** operational
changes. It is the proposed default.

### Layer 2 — optional kernel-level confinement via util-linux, driven from Java

For deployments that want T2/T3/T4 as hard guarantees, the same kernel primitives Docker uses
(namespaces, rlimits, uid separation) are reachable through three single-purpose binaries
that ship with util-linux / are one package away. The Java executor composes them as an
**argv prefix** — still no shell:

```
setpriv --reuid=<uid> --regid=<gid> --clear-groups \   # T2: dedicated unprivileged user
unshare --user --map-current-user --net \             # T3: network namespace, unprivileged
prlimit --cpu=<s> --as=<bytes> --nproc=<n> --fsize=<bytes> \  # T4: rlimits
setsid <tool argv…>                                    # existing process-group discipline
```

Compared to Docker this keeps every security flag the branch actually values and deletes the
daemon, the socket mount, the image, the shared-path requirement, the orphan reaping, the
probing, the fallback policy, the tmpfs copy, and — because the real workspace is used
directly — the entire `WORKSPACE_SETUP_COMMAND` / `INSTALL_PACKAGE_ENV` / base64-artifact
apparatus. `node_modules` and build caches stay warm; E2E artifacts land where the consumers
already read them.

Honest limitations (to be stated in DEPLOYMENT.md):

- **uid separation needs privilege.** `setpriv --reuid` to a *different* uid requires the JVM
  to run as root (or with CAP_SETUID). In the stock deployment the app runs as uid 1000, so
  Layer 2 either (a) runs commands as the same uid — filesystem protection of app files then
  rests on ordinary permissions (chmod 700 on config/credential dirs, which is good practice
  anyway) — or (b) is enabled only where the operator provisions it. `unshare --net`, by
  contrast, works unprivileged via a user namespace on mainstream kernels
  (note: Ubuntu ≥ 23.10 restricts unprivileged user namespaces via AppArmor; probe at
  startup and fail closed or degrade per config).
- **rlimits are coarser than cgroups.** `prlimit --nproc` counts per real uid (effective fork
  bomb protection needs a dedicated uid, i.e. (b) above); `--as` is address space, not RSS.
  Acceptable for a build command under a timeout; not a multi-tenant guarantee.
- **No disk bound.** Same as develop's direct mode; the tmpfs disk bound was a Docker-only
  feature protecting a disposable asset (T5).

Layer 2 is **opt-in** (`agent.sandbox.hardened=true`), probed once at startup, fail-closed
when enabled-but-unavailable unless `fallback-to-direct` is explicitly set — same policy
shape as the branch, minus the Docker daemon.

---
AI-Disclaimer: I shared my concerns with Kimi and presented a vague idea for a solution. Kimi then helped me flesh out the ideas with concrete shell commands (especially the Layer 2 step), as my knowledge there is limited. So now, the next step is to double-check whether the proposed solutions are actually viable.

tmseidel (changes requested):
see my last comment

CaeruleusAqua:
I'll brainstorm a little bit too, and we'll see...

CaeruleusAqua:
Thanks for the detailed threat model — that's a really useful framing, and we agree with more of it than the current PR shape suggests. Let me respond to the core points, then make a concrete proposal.

## Where we agree

1. **T1 is the main win, and it's pure Java.** Env scrubbing, no-shell argv execution, setsid + process-group kill, descendant tracking — none of it needs a container.
2. **The socket mount is a real problem.** You're right: mounting `/var/run/docker.sock` into the app container turns "sandbox escape" into "host takeover". We don't want to defend that — we want to fix it.
3. **The Docker machinery is heavy for what it protects.** The embedded setup script, the base64 artifact channel, orphan reaping, availability probing, the shared-path requirement — that's a lot of moving parts defending a disposable workspace.

## Proposal: split the PR

**This PR becomes Layer 1 only** — env scrubbing, no-shell argv execution, setsid + process-group kill, descendant tracking, bounded output, fail-closed behavior. Zero operational change, platform-independent, no Docker anything. The review findings already fixed here (artifact-channel budget, config-only network check, absolute-path validation) mostly dissolve with the Docker code, which also removes a whole class of fragility you rightly criticized.

**The hard-isolation layer moves to a follow-up PR**, where we can discuss the approach properly instead of entangling it with the Java hardening.

## For that follow-up: the socket problem is solvable without dropping Docker

The established pattern is to keep the socket out of the app container entirely: a dedicated **sandbox sidecar** is the only container holding `/var/run/docker.sock`, and the app talks to a narrow API that only accepts predefined operations (fixed image, fixed mounts, fixed limits — no volume/bind parameters at all). This is the same model as GitHub Actions, where the runner executing untrusted workflow code has no Docker daemon access, and the Codex/OpenHands sidecar approach. Even a fully compromised app process can't reach the daemon, let alone start a privileged container — the attack surface you identified disappears structurally, not cosmetically.

## One addition to your Layer-2 limitations list (for the follow-up discussion)

Your "honest limitations" section already covers the main ones (uid separation needs root/CAP_SETUID, Ubuntu ≥ 23.10 AppArmor restriction, rlimits being coarser than cgroups). One more worth adding, specific to this project's target deployment: **the app itself runs in Docker** (as uid 1000, default seccomp profile, no CAP_SYS_ADMIN). In that environment, `unshare` is blocked by Docker's default seccomp profile regardless of the host distro (moby/moby#42441, repro moby/moby#43086). So even on a permissive host (no AppArmor restriction), Layer 2's `unshare --user --net` fails inside the stock app container with EPERM.

Making it work would mean granting the app container `--cap-add SYS_ADMIN` or `seccomp=unconfined` — arguably the same attack-surface class as the socket mount, just relocated. Not a dealbreaker, but worth documenting honestly alongside your existing list.

## The maintenance trade-off — your call

Whichever hardening path the follow-up takes (sidecar-Docker or util-linux), shipping **both** would mean maintaining both: two probe paths, two fail-closed policies, two sets of distro caveats, two test matrices. For a project with limited maintainer capacity, we'd recommend picking one.

So: OK if we reshape this PR to Layer 1 only, and continue the Docker-sidecar vs. util-linux discussion in a follow-up PR/issue? We can sketch the sidecar there so it's concrete to review.

tmseidel:
+1 for the iterational path. Let's scope this PR to Level 1 only. I need to do some additional research for a final decision, so yes we should move the rest to a separate PR.

CaeruleusAqua:
Done — the PR is now scoped to **Layer 1 only**, rebased onto current `develop` and force-pushed as `d092a48`.

**What remains (pure Java, no container):**
- `ProcessSupport` — environment scrubbing (toolchain allowlist), `setsid` process-group isolation on Linux, descendant tracking + forcible kill, bounded output with UTF-8-safe truncation.
- `ToolExecutionService.executeCommand` — now scrubs the child environment and runs through `ProcessSupport` (bounded output, descendant tracking, timeout).
- `WorkspaceProcessRunner` — direct execution with scrubbed environment + `ProcessSupport.run` (no Docker executor, no base64 artifact channel; E2E artifacts land in the workspace where consumers already read them).
- `PrTestWorkspaceManager` — scrubbed `npm install`; additionally keeps the absolute-path validation for `ai-git-bot.e2e.workspace-root` (fail-fast instead of silently resolving against the process cwd).

**Removed:** `SandboxedCommandExecutor` (Docker), `docker-compose.sandbox.yml`, the `docker.io` package in the Dockerfile, the `agent.sandbox.*` configuration, the DEPLOYMENT.md sandbox section, and all Docker-specific tests.

The hard-isolation discussion (Docker sidecar vs. util-linux) moves to a follow-up, as agreed. Full test suite: 1432 tests, 0 failures.

max-california (approved):
## 🤖 Agentic Code Review

## Summary

This is a solid hardening pass overall: the new shared `ProcessSupport` consolidates timeout/output handling, removes inherited secrets from untrusted child processes, and adds stronger cleanup semantics for subprocess trees. The new tests cover the main behaviors well.

I found one correctness issue in the Linux process-group path that can kill the wrong processes on timeout/successful cleanup, plus one maintainability/performance concern around the descendant tracker.

## Findings

### BLOCKER

1. **Linux process-group cleanup targets the `setsid` wrapper PID, not the child’s actual process group**
   - **Files:** `src/main/java/org/remus/giteabot/util/ProcessSupport.java`
   - **Lines:** around `runInNewProcessGroup()` and `terminateProcessGroup()`
   - `runInNewProcessGroup()` prepends `setsid` and then later calls `terminateProcessGroup(process.pid())`.
   - The PID returned by `processBuilder.start()` is the PID of the `setsid` utility, not necessarily the PID / PGID of the spawned target command after `setsid` re-execs or forks. Sending `kill -KILL -- -<leaderPid>` assumes that this PID is the process-group leader for the untrusted command tree.
   - If that assumption does not hold, cleanup can fail to kill the actual descendants, or worse, target an unrelated process group if the wrapper exits and the PID is reused before cleanup runs.
   - This is especially risky because the process-group kill is invoked both on timeout and after “successful” completion specifically to prevent background children from surviving.
   - **Suggested fix:** avoid inferring the PGID from the Java `Process` PID. Either:
     - invoke a tiny shell wrapper that prints its own PGID / child PID and `exec`s the target so you can kill the correct group deterministically, or
     - use a Java/native approach that obtains the real process group ID explicitly before cleanup.
   - At minimum, add a test that proves the timeout/success cleanup kills the intended descendant process tree when launched through `setsid`, not just when using Java descendant tracking.

### MEDIUM

1. **10ms descendant polling thread is fairly aggressive and runs for every non-Linux untrusted process**
   - **File:** `src/main/java/org/remus/giteabot/util/ProcessSupport.java`
   - **Lines:** `startDescendantTracker()`
   - The tracker wakes every 10ms and snapshots `process.toHandle().descendants().toList()` repeatedly until exit. That’s a lot of churn for long-running commands and can become surprisingly expensive if multiple validations/E2E runs execute concurrently.
   - It also introduces a maintainability burden because correctness depends on high-frequency polling to catch short-lived children.
   - **Suggested fix:** consider relaxing the interval substantially, or only enabling tracking on timeout paths / for commands known to spawn shells, or documenting why 10ms is required with evidence from the failure mode you’re preventing. If possible, prefer process-group based cleanup where available and a simpler best-effort fallback elsewhere.

### LOW

1. **`scrubEnvironmentForGit()` is introduced but currently unused**
   - **File:** `src/main/java/org/remus/giteabot/util/ProcessSupport.java`
   - This helper looks useful, but nothing in the current diff calls it.
   - If there are Git transport subprocesses elsewhere in the repo, it would be good to either wire them up in this PR or add a comment clarifying that follow-up adoption is intentional.

2. **Output truncation semantics differ between byte-bound capture and char-bound post-truncation**
   - **File:** `src/main/java/org/remus/giteabot/agent/validation/ToolExecutionService.java`
   - The command output is first bounded by bytes in `ProcessSupport.run(..., MAX_TOOL_OUTPUT_CHARS + 1_000)` and then truncated again by characters in `truncateOutput(...)`.
   - That’s not wrong, but it means the effective visible limit depends on UTF-8 density and the extra `+ 1_000` fudge factor. A short comment explaining why both limits exist would make the behavior easier to reason about later.

---

CaeruleusAqua:
Thanks for the review — all four findings are addressed in \124cd04\, with the rationale below.

## BLOCKER (setsid PGID inference)

The premise that the Java-visible PID may not be the process-group ID does not hold for the util-linux \setsid\ used here: it forks only when it is already a process-group leader (\getpgrp() == getpid()\), and a \ProcessBuilder\ child can never be one — it inherits the JVM's process group, and the kernel never allocates a PID that is still in use. \setsid\ therefore calls \setsid(2)\ and \execvp\s the target without forking, so the PID survives unchanged as the new PGID and \kill -KILL -- -<pid>\ targets the correct group. We documented this guarantee in the javadoc (including the 'as long as the JVM's group leader lives' precondition) and added the requested regression test for the timeout path: \
unInNewProcessGroup_terminatesBackgroundChildOnTimeout\ proves a background child launched through \setsid\ is killed on timeout, complementing the existing success-path test.

The real limitation your instinct points at is now documented: a child that daemonizes itself (calls \setsid\ again) leaves the group and survives cleanup, and on non-Linux a reparented daemon escapes the descendant tracker the same way. Closing that requires uid/cgroup separation — the follow-up hard-isolation layer.

## MEDIUM (10ms descendant polling)

Relaxed to 250ms with a comment explaining that the interval is not a correctness parameter: a child that spawns and dies between two ticks needs no kill, and any child alive at a tick is captured. Note the tracker only runs on non-Linux (Linux uses process-group cleanup), so the churn cost is limited to the best-effort fallback path.

## LOW (scrubEnvironmentForGit unused)

Wired up: \WorkspaceService.runCommand\ now calls \ProcessSupport.scrubEnvironmentForGit\, and the duplicated private variant (\GIT_ENVIRONMENT_ALLOWLIST\) was removed. The stricter variant still removes HOME/USERPROFILE/APPDATA; the additionally inherited variables are toolchain vars only (JAVA_HOME, NODE_PATH, …), not secrets. \GIT_CONFIG_NOSYSTEM\/\GIT_CONFIG_GLOBAL\ handling is unchanged.

## LOW (byte vs. char truncation)

Added a comment at the call site: the byte cap is the hard memory bound while draining the pipe, the character cap is the presentation limit, and the \+1_000\ headroom makes the char cap the effective limit for ASCII-dominant output, while multi-byte-heavy output may hit the byte cap first.

Verification: full Maven suite in Docker — 1433 tests, 0 failures; CI \Build and verify\ passes.

CaeruleusAqua:
Thanks for the review — all four findings are addressed in `124cd04`, with the rationale below.

### BLOCKER (setsid PGID inference)

The premise that the Java-visible PID may not be the process-group ID does not hold for the util-linux `setsid` used here: it forks only when it is already a process-group leader (`getpgrp() == getpid()`), and a `ProcessBuilder` child can never be one — it inherits the JVM's process group, and the kernel never allocates a PID that is still in use. `setsid` therefore calls `setsid(2)` and `execvp`s the target without forking, so the PID survives unchanged as the new PGID and `kill -KILL -- -<pid>` targets the correct group. We documented this guarantee in the javadoc (including the "as long as the JVM's group leader lives" precondition) and added the requested regression test for the timeout path: `runInNewProcessGroup_terminatesBackgroundChildOnTimeout` proves that a background child launched through `setsid` is killed on timeout, complementing the existing success-path test.

The real limitation your instinct points at is now documented: a child that daemonizes itself (calls `setsid` again) leaves the process group and survives cleanup, and on non-Linux a reparented daemon escapes the descendant tracker the same way. Closing that requires uid/cgroup separation — the follow-up hard-isolation layer.

### MEDIUM (10ms descendant polling)

Relaxed to 250ms with a comment explaining that the interval is not a correctness parameter: a child that spawns and dies between two ticks needs no kill, and any child alive at a tick is captured. Note the tracker only runs on non-Linux (Linux uses process-group cleanup), so the churn cost is limited to the best-effort fallback path.

### LOW (scrubEnvironmentForGit unused)

Wired up: `WorkspaceService.runCommand` now calls `ProcessSupport.scrubEnvironmentForGit`, and the duplicated private variant (`GIT_ENVIRONMENT_ALLOWLIST`) was removed. The stricter variant still removes HOME/USERPROFILE/APPDATA; the additionally inherited variables are toolchain vars only (JAVA_HOME, NODE_PATH, …), not secrets. The `GIT_CONFIG_NOSYSTEM` / `GIT_CONFIG_GLOBAL` handling is unchanged.

### LOW (byte vs. char truncation)

Added a comment at the call site: the byte cap is the hard memory bound while draining the pipe, the character cap is the presentation limit, and the `+1_000` headroom makes the char cap the effective limit for ASCII-dominant output, while multi-byte-heavy output may hit the byte cap first.

Verification: full Maven suite in Docker — 1433 tests, 0 failures; CI `Build and verify` passes.

tmseidel:
@CaeruleusAqua I'll do still a test, but the Code LGTM. Great work!

tmseidel (approved):

PR #323, opened 2026-08-08
encrypt webhook and MCP configuration secrets

CaeruleusAqua:
## What was done and why?

- Encrypts outgoing event-hook custom-header JSON and MCP configuration JSON with the existing `EncryptionService` before persistence when `APP_ENCRYPTION_KEY` is configured.
- Decrypts values only at the event-hook delivery, MCP discovery/execution, and admin edit boundaries.
- Blocks direct request binding of encrypted event-hook fields so plaintext cannot bypass the encryption path.
- Preserves readable legacy plaintext and encrypts it when the configuration is next saved.
- Adds focused coverage for encrypted persistence, delivery, form binding, MCP discovery, and decrypted edit views.

## Testing

- Focused Event Hook and MCP test suite: 72 tests
- Git whitespace validation

### AI use

### DIFF
--- a/src/main/java/org/remus/giteabot/eventhook/EventHookController.java
+++ b/src/main/java/org/remus/giteabot/eventhook/EventHookController.java
@@ -4,7 +4,9 @@ import lombok.extern.slf4j.Slf4j;
 import org.remus.giteabot.admin.BotService;
 import org.springframework.stereotype.Controller;
 import org.springframework.ui.Model;
+import org.springframework.web.bind.WebDataBinder;
 import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.InitBinder;
 import org.springframework.web.bind.annotation.ModelAttribute;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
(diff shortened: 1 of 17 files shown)


max-california (approved):
## 🤖 Agentic Code Review

Summary: the change is directionally solid and the test coverage is good. I did find one correctness issue around editing existing MCP configurations that can silently drop related tool-selection data.

## BLOCKER

### Editing an MCP configuration returns a detached partial entity and can wipe associations on save
`src/main/java/org/remus/giteabot/systemsettings/McpConfigurationService.java:73`  
`src/main/java/org/remus/giteabot/systemsettings/SystemSettingsController.java:63`  
`src/main/java/org/remus/giteabot/systemsettings/McpConfiguration.java:35`

`decryptedView()` creates a brand-new `McpConfiguration` containing only `id`, `name`, and decrypted `jsonContent`, and `editMcpForm()` binds that object back through the form. On submit, `save()` persists that partial entity directly.

Because `McpConfiguration` has `selectedTools` mapped with `cascade = ALL, orphanRemoval = true`, saving a fresh detached instance with the collection left at its default empty list risks replacing the managed entity state with an empty association set, orphan-removing previously selected tools during a normal config edit.

This is especially likely because the edit form does not include `selectedTools`, `createdAt`, etc., so the detached object is incomplete by design.

**Suggested fix:** on update, load the existing entity from the repository and copy only editable fields (`name`, plaintext/decrypted `jsonContent`) onto the managed instance before encrypting and saving. Alternatively, have `decryptedView()` preserve `selectedTools`, but merging a partial detached entity is still fragile compared to updating the managed entity.

## LOW

No other concrete issues stood out from the reviewed paths; the event-hook binding hardening and decrypt-at-use boundaries look consistent.

---

tmseidel (approved):

PR #324, opened 2026-08-08
fail closed invalid login configuration

CaeruleusAqua:
## What was done and why?

- Validates `giteabot.security.login-method` during startup and aborts for every value except `native` and `oauth`, preventing an undefined web-security configuration.
- Redirects local OAuth logout to a public `/signed-out` page instead of immediately re-entering the provider-login flow.
- Adds coverage for supported and unsupported login methods, logout redirect behavior, and anonymous access to the signed-out page.

## Testing

- Focused login and OAuth MVC suite: 6 tests
- Git whitespace validation

### AI use

### DIFF
--- a/src/main/java/org/remus/giteabot/admin/LoginMethodValidator.java
+++ b/src/main/java/org/remus/giteabot/admin/LoginMethodValidator.java
@@ -1,0 +1,22 @@
+package org.remus.giteabot.admin;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Fail-fast validation of the web login method. A typo in
+ * {@code giteabot.security.login-method} must not silently leave the app
+ * without any web security chain (fail-open) - refuse to start instead.
+ */
+@Configuration
+public class LoginMethodValidator {
+
+    public LoginMethodValidator(
+            @Value("${giteabot.security.login-method:native}") String loginMethod) {
+        if (!"native".equals(loginMethod) && !"oauth".equals(loginMethod)) {
+            throw new IllegalStateException(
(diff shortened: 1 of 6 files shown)


max-california (approved):
## 🤖 Agentic Code Review

Summary: this looks like a solid hardening change. The startup validation closes the “no matching security chain” gap, and the OAuth logout redirect avoids the confusing immediate re-login loop. I didn’t find any correctness or security regressions in the touched paths.

## Findings

### LOW
- **`LoginMethodValidatorTest` only unit-tests the constructor, not actual application startup wiring.**  
  The intent of this change is specifically “fail during startup” for unsupported `giteabot.security.login-method` values, but `src/test/java/org/remus/giteabot/admin/LoginMethodValidatorTest.java` currently verifies that by directly instantiating the class. That proves the guard logic, but not that Spring actually creates this bean early enough in a real application context.  
  Suggestion: add a small context-level test (e.g. `ApplicationContextRunner` or a minimal `@SpringBootTest`) asserting that `login-method=disabled` fails context initialization and `native` / `oauth` start successfully.

### LOW
- **The new signed-out page is tightly coupled to a single OAuth registration id in the template.**  
  In `src/main/resources/templates/signed-out.html:20`, the “Sign in again” link hardcodes `/oauth2/authorization/giteabot`. That matches the current design (`OAuthLoginConfiguration.REGISTRATION_ID` is constant), so it’s fine today, but it means the template must be updated if the registration id ever changes.  
  Suggestion: if you want to reduce that coupling, pass the re-login URL from the controller/model (or expose the registration id centrally) rather than hardcoding it in the template.

---

tmseidel (approved):

PR #332, opened 2026-08-12
feat: enable prompt caching support in Anthropic DTOs (system + cache token fields)

### LINKED ISSUE #331: Anthropic client: enable prompt caching
We want to use Claude for our reviews with the agentic review bot.
But it is very expensive, because it is "chatting" a lot with the API which uses many tokens.

With activating caching it should be reducing the costs by a lot.

The following text is generated by Claude AI.

---

## Problem

The Anthropic client cannot emit `cache_control`, so every round of the agentic
review loop re-sends the whole prefix (system prompt + diff + all previous tool
results) at full input price.

Measured on one agentic review of a single PR with `claude-opus-4-7`: 18 requests
in 96 seconds, **709,545 input vs. 5,001 output tokens** — input is 97 % of the
cost, ~$3.67 for that one review. The non-agentic workflow on a comparable PR
costs ~$0.25.

The prefix reaches ~37k tokens by round 4 and is re-sent unchanged from there on.
Rounds 4–11 alone consume 296,052 input tokens — 42 % of the bill — to add 5,518
tokens of new context (small `grep` / `cat` results). With cache reads at 0.1x and
writes at 1.25x, applying those multipliers to the observed prefix sequence puts
input cost at roughly **1/5**. Rounds are 3–16 s apart, so the default 5-minute
TTL covers the whole loop.

## Root cause

At `9f6db9d`, `system` is typed as `String` — only the array-of-text-blocks form
of `system` can carry a cache breakpoint, so this is a DTO-level blocker, not a
missing config flag:

* `ai/anthropic/AnthropicRequest.java` — L22 `private String system;`; `ContentBlock`
  (L41) has no `cache_control` field.
* `ai/anthropic/AnthropicAiClient.java` — L148 `.system(systemPrompt)` in
  `chatWithTools(...)` (the agentic loop; L68 / L91 are the single-request and chat
  paths, lower priority).
* `ai/anthropic/AnthropicResponse.java` — `Usage` (L36) lacks the cache token
  fields, so cache activity is neither observable nor billable.

## Proposed change

```java
// AnthropicRequest: was  private String system;
private List<ContentBlock> system;

// AnthropicRequest.ContentBlock
@JsonProperty("cache_control")
private CacheControl cacheControl;

@Data @Builder @JsonInclude(JsonInclude.Include.NON_NULL)
public static class CacheControl {
    @Builder.Default private String type = "ephemeral";
    private String ttl;   // null = 5 min, "1h" for the long TTL
}

// AnthropicResponse.Usage
@JsonProperty("cache_creation_input_tokens") private int cacheCreationInputTokens;
@JsonProperty("cache_read_input_tokens")     private int cacheReadInputTokens;
```

In `chatWithTools` two breakpoints: one on the system block, and a rolling one on
the last content block of the last message. The rolling one is what matters — it
lets round N+1 read round N's whole prefix at 0.1x. A per-integration toggle
(default on for Anthropic) would keep this out of the way for providers and local
models that don't support caching.

## Implementation notes

* **Render order is `tools` → `system` → `messages`.** Tool definitions sit at the
  front of the prefix, so a non-deterministic tool order silently invalidates
  *everything* — no error, just permanent misses. If the tool list at
  `AnthropicAiClient` L141–143 comes from a set or an unordered query, it needs a
  stable sort.
* **20-block lookback.** A breakpoint searches back at most 20 content blocks. A
  round adds ~2 blocks (`tool_use` + `tool_result`), so from ~round 10 a single
  rolling breakpoint stops finding its predecessor. Keeping an older breakpoint
  alive every ~8 rounds helps (4 allowed per request).
* **`input_tokens` changes meaning.** On a hit it reports only the uncached
  remainder; total size is `input_tokens + cache_creation + cache_read`. Cost
  accounting and metrics that read `input_tokens` alone will under-report after
  this change (~300 instead of ~37,000 per round in the example) — they should
  weight the three fields 1.0 / 1.25 / 0.1.

## Verification

`cache_read_input_tokens > 0` from round 2 onwards; in the example round 2 should
read ~14,500 tokens instead of 0. `AnthropicAiClientTest` can pin the
serialization (`system` as array, `cache_control` on the intended block,
deterministic tool order).

Docs: <https://platform.claude.com/docs/en/build-with-claude/prompt-caching>

MakerYuichi:
Fixes #331 (partial — DTO/serialization layer only, see scope note below)

**What has been done?**
Restructured `AnthropicRequest.system` from a plain `String` to a `List<ContentBlock>`, since only the array form of `system` can carry a `cache_control` breakpoint. Added a `CacheControl` DTO and a `cacheControl` field on `ContentBlock`. Added `cache_creation_input_tokens` and `cache_read_input_tokens` to `AnthropicResponse.Usage` so cache activity is observable in usage reporting. Updated the three call sites in `AnthropicAiClient` (`sendReviewRequest`, `sendChatRequest`, `chatWithTools`) that previously passed a plain system-prompt string, via a new `toSystemBlocks()` helper.

**Why?**
The agentic review loop currently re-sends the full prefix (system prompt + diff + all prior tool results) at full input price on every round, because the client has no way to emit `cache_control`. On a measured example PR review, input tokens were 97% of the cost (~$3.67 for one review); with cache reads billed at 0.1x, the same review's input cost would drop to roughly 1/5. This PR lays the DTO groundwork; it does not yet place breakpoints or change request behavior.

**What has been tested?**
Added `AnthropicRequestTest` (4 new tests): confirms `system` serializes as a JSON array rather than a string, confirms `cache_control` appears only on blocks that explicitly set it, confirms a null `ttl` is correctly omitted per `@JsonInclude(NON_NULL)`, and confirms `Usage` deserializes the new cache token fields from a sample response. Full suite passes locally (inside the project's Docker build environment): `mvn clean test` — 1419/1419 tests green (1415 pre-existing + 4 new), including `ArchitectureTest`.

**How AI was used?**
Used Claude as a coding assistant throughout this change.

**Scope note**
This covers only the DTO/serialization layer described in the issue's "Proposed change" section. The actual cache-breakpoint placement logic in `chatWithTools` (rolling breakpoint on the last content block, the
20-block lookback constraint, and the deterministic-tool-ordering fix noted in the issue) is intentionally left for a follow-up PR, per my comment on #331.

### DIFF
--- a/src/main/java/org/remus/giteabot/ai/anthropic/AnthropicAiClient.java
+++ b/src/main/java/org/remus/giteabot/ai/anthropic/AnthropicAiClient.java
@@ -10,6 +10,7 @@ import org.remus.giteabot.ai.StopReason;
 import org.remus.giteabot.ai.ToolCall;
 import org.remus.giteabot.ai.ToolDescriptor;
 import org.remus.giteabot.ai.ToolNameSanitizer;
+import org.remus.giteabot.ai.anthropic.AnthropicRequest.ContentBlock;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestClient;
 import tools.jackson.databind.JsonNode;
@@ -65,7 +66,7 @@ public class AnthropicAiClient extends AbstractAiClient {
         AnthropicRequest request = AnthropicRequest.builder()
                 .model(effectiveModel)
                 .maxTokens(maxTokens)
-                .system(systemPrompt)
+                .system(toSystemBlocks(systemPrompt))
                 .messages(List.of(
                         AnthropicRequest.Message.builder()
                                 .role("user")
@@ -88,7 +89,7 @@ public class AnthropicAiClient extends AbstractAiClient {
         AnthropicRequest request = AnthropicRequest.builder()
                 .model(effectiveModel)
                 .maxTokens(maxTokens)
-                .system(systemPrompt)
+                .system(toSystemBlocks(systemPrompt))
                 .messages(anthropicMessages)
                 .build();
 
@@ -145,7 +146,7 @@ public class AnthropicAiClient extends AbstractAiClient {
         AnthropicRequest request = AnthropicRequest.builder()
                 .model(effectiveModel)
                 .maxTokens(effectiveMaxTokens)
-                .system(systemPrompt)
+                .system(toSystemBlocks(systemPrompt))
                 .messages(messages)
                 .tools(toolPayloads)
                 .build();
@@ -448,6 +449,21 @@ public class AnthropicAiClient extends AbstractAiClient {
                 .body(request)
                 .retrieve()
                 .body(AnthropicResponse.class);
+    }
+
+    /**
+    * Wraps a plain system-prompt string as a single-element content-block
+    * list, the shape required since {@link AnthropicRequest#getSystem()}
+    * became cache-controllable (must be an array, not a bare string)
+    */
+
+    private List<AnthropicRequest.ContentBlock> toSystemBlocks(String systemPrompt){
+        return List.of(AnthropicRequest.ContentBlock.builder()
+        .type("text")
+        .text(systemPrompt)
+        .build()
+
+    );
     }
 }
 
--- a/src/main/java/org/remus/giteabot/ai/anthropic/AnthropicAiClient.java
+++ b/src/main/java/org/remus/giteabot/ai/anthropic/AnthropicAiClient.java
@@ -10,7 +10,6 @@ import org.remus.giteabot.ai.StopReason;
 import org.remus.giteabot.ai.ToolCall;
 import org.remus.giteabot.ai.ToolDescriptor;
 import org.remus.giteabot.ai.ToolNameSanitizer;
-import org.remus.giteabot.ai.anthropic.AnthropicRequest.ContentBlock;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestClient;
 import tools.jackson.databind.JsonNode;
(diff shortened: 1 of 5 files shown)


max-california (approved):
## 🤖 Agentic Code Review

Summary: The DTO changes look directionally correct and the added serialization tests cover the new request/response shapes well. I did not find a functional bug in the main request-building flow. The main gap I see is test coverage at the client-call-site level: the PR changes the wire shape for every Anthropic request, but only DTO-only tests were added.

## MEDIUM

1. **No client-level regression test for the changed `system` payload shape**
   - **Why it matters:** `AnthropicAiClient` now sends `system` as a `List<ContentBlock>` in all three request paths (`sendReviewRequest`, `sendChatRequest`, `chatWithTools`), which is the core behavior change of the PR. The current tests only prove that `AnthropicRequest` serializes correctly when manually constructed, not that these production call sites actually build the intended payload.
   - **Where:** `src/main/java/org/remus/giteabot/ai/anthropic/AnthropicAiClient.java:66-76`, `89-94`, `146-152`
   - **Suggestion:** Add an `AnthropicAiClientTest` that captures the request body sent to `RestClient` (or exercises the request construction indirectly) and asserts that:
     - `system` is emitted as an array with a single `{"type":"text","text":...}` block
     - legacy user messages still serialize as plain strings
     - native-tool messages still serialize their content blocks correctly alongside the new `system` shape

## LOW

1. **Style/readability issue in `toSystemBlocks()`**
   - **Where:** `src/main/java/org/remus/giteabot/ai/anthropic/AnthropicAiClient.java:454-460`
   - **Why it matters:** The helper is correct, but its formatting/Javadoc indentation is inconsistent with the surrounding file, which makes the change look rougher than the rest of the class.
   - **Suggestion:** Reformat the Javadoc and method body to standard project style (`private List<ContentBlock> toSystemBlocks(String systemPrompt) { ... }`).

2. **Unused import introduced**
   - **Where:** `src/main/java/org/remus/giteabot/ai/anthropic/AnthropicAiClient.java:13`
   - **Why it matters:** `import org.remus.giteabot.ai.anthropic.AnthropicRequest.ContentBlock;` does not appear to be used, since the method still references `AnthropicRequest.ContentBlock` fully qualified.
   - **Suggestion:** Remove the unused import or use the imported short name consistently.

---

tmseidel:
@MakerYuichi Thanks for your contribution. I'll test the code if it will fulfill the promised value. Can you please add the following:
1. Add some tests for the `AnthropicAiClient` (like the Review suggested)
2. Check `AnthropicAiClient` for proper formatting (like the Review suggested)

MakerYuichi:
Done

Added AnthropicAiClientRequestShapeTest covering all three call sites (sendReviewRequest, sendChatRequest, chatWithTools), confirming system is emitted as a single-element array and legacy/tool message shapes are unaffected. Also cleaned up the formatting and removed the unused import in AnthropicAiClient. 
Full suite green locally — 1422/1422 tests passing.

tmseidel (approved):
LGTM

PR #339, opened 2026-08-15
feat(anthropic): add prompt caching to cut agentic review cost

### LINKED ISSUE #331: Anthropic client: enable prompt caching

tmseidel:


max-california (approved):
## 🤖 Agentic Code Review

Summary: the Anthropic prompt-caching implementation looks well thought out overall, and the added tests cover the request-shape and accounting changes nicely. I found one maintainability/regression issue in the factory caching layer; otherwise this looks solid.

## MEDIUM

### Deleted `AiClientFactory.evict()` leaves stale client entries behind after integration deletion
In `src/main/java/org/remus/giteabot/admin/AiClientFactory.java`, this PR removes the `evict(Long integrationId)` method, but `AiClientFactory` still keeps a process-local `ConcurrentMap<Long, CachedClient>` keyed by integration ID.

That means deleting an integration now no longer provides any way to clear its cached client. If any code path still holds or reuses the deleted `AiIntegration` object (or if the same ID is reintroduced in tests/dev scenarios), the factory can continue returning a stale client with now-deleted credentials/config. Even if this is not an immediate functional bug today, removing the only explicit invalidation hook increases the chance of long-lived stale entries and makes future delete/update flows harder to reason about.

Suggested fix:
- keep `evict(Long integrationId)` in `AiClientFactory`, and
- call it from the delete flow (or document clearly why cached clients are intentionally allowed to survive deletion for the life of the process).

## LOW

### Consider adding an export/API representation for the new cache token columns
The UI table in `src/main/resources/templates/usage/list.html` now shows `cacheCreationInputTokens` and `cacheReadInputTokens`, but `AiUsageService.exportErrors(...)` is the only export-like path I found and there doesn’t appear to be a matching usage export/update in this PR.

If there is a usage export or API elsewhere, it would be worth checking that the new fields are included there too, otherwise the UI and exported/audited data can drift.

---

PR #340, opened 2026-08-15
refactor(bot): remove per-bot agentEnabled toggle

### LINKED ISSUE #338: Remove org.remus.giteabot.admin.Bot#agentEnabled
This property should be removed. The application should behave like if it would be enabled today

tmseidel:


max-california (approved):
## 🤖 Agentic Code Review

Summary: this refactor looks consistent overall: the entity field, UI toggle, docs, migration, and the routing/tests were updated together, and I didn’t find any obvious broken references. The behavioral intent also seems coherent with the newer workflow-based model: `issue-coding` now always invokes the coding agent when that workflow is configured, instead of being additionally gated by a per-bot flag.

## Findings

### MEDIUM
- **Potential migration compatibility risk for non-clean databases**  
  The new Flyway migration only drops the column:
  - `src/main/resources/db/migration/postgresql/V44__bots_drop_agent_enabled.sql`
  - `src/main/resources/db/migration/h2/V44__bots_drop_agent_enabled.sql`

  This is fine if every deployed database already has `agent_enabled`, but it can fail on environments that were created after the workflow-based refactors and never had the column, or on divergent test/preview schemas. You used `IF EXISTS`, which helps on PostgreSQL/H2, so this may be acceptable depending on your supported DB versions; the main thing I’d suggest is verifying Flyway compatibility for the exact H2/Postgres versions you test against, because `DROP COLUMN IF EXISTS` is not universally supported the same way across engines/versions. If older H2 versions are still in play, this migration may become the first failure point.

### LOW
- **No explicit regression test covering “coding workflow configured => assignment/comment always dispatches”**  
  The removed tests covered the old `agentEnabled` gate, but I don’t see a focused replacement asserting the new contract in `CodingIssueWorkflow` itself:
  - `src/main/java/org/remus/giteabot/issueworkflow/coding/CodingIssueWorkflow.java`

  Existing integration-style tests may already exercise it indirectly, but a small unit test here would make the intent durable: when `issue-coding` is enabled, `onIssueAssigned` and `onIssueComment` should always call `createIssueImplementationService(bot)` regardless of legacy bot shape.

### LOW
- **Minor template indentation drift after column removal**  
  In `src/main/resources/templates/bots/list.html`, the removed “Agent” column left some uneven indentation/alignment around the `<th>`/`<td>` blocks. Not a functional issue, but worth cleaning up for maintainability/readability in future template edits.

---

PR #347, opened 2026-08-21
feature/usage-log

### LINKED ISSUE #342: Log full LLM Responses
The usage -> errors section already shows the response from the llm in case of an error (401 / 403 etc).

For debugging purposes, I would like to also log the responses on any sucessfull ai calls (in the usage -> ai usage section).

This extended Logging configuration should have to be manually activated first (e.g. in the AI Integrations tab). 
Default is off - to save disk space.

Background: Right now I am wondering why deepseek takes 20 Rounds to review my PRs, and am therefore still testing around with what tools I should give to this bot. Having the LLM response would make debugging / configuring easier.

tmseidel:


max-california (approved):
## 🤖 Agentic Code Review

Summary: the feature is mostly wired through correctly end-to-end: provider clients now pass request/response payloads into the audit path, the DB schema and entity were extended, and the usage page can render the stored payloads. I found one substantive security/privacy regression in what gets exposed and retained.

## BLOCKER

### Raw payloads likely expose secrets and sensitive user/repo content without any redaction
The new audit trail stores and displays full provider request/response bodies (`src/main/java/org/remus/giteabot/ai/AbstractAiClient.java:62`, `src/main/java/org/remus/giteabot/aiusage/AiUsageService.java:56`, `src/main/resources/templates/usage/list.html:127`). Those payloads can include:

- complete prompts containing repository code / PR diffs / comments
- tool-call arguments and outputs
- provider-specific metadata that may contain confidential content
- potentially credentials if any provider request model ever carries auth/config fields in-body

At the moment this is persisted verbatim and shown in the admin UI verbatim. That’s a significant data-retention/privacy expansion, and there’s no redaction, truncation, opt-in flag, or warning in this path.

Actionable fixes:
- redact known-sensitive fields before persistence
- cap stored payload size server-side (not just DB column length)
- consider making raw payload logging configurable/off by default
- document/admin-gate this as sensitive audit data explicitly

## LOW

### Entity length hint does not match the actual `TEXT` migration and may truncate unexpectedly on some dialects
`AiUsageLog` declares `@Column(name = "raw_request", length = 65535)` and same for `raw_response` (`src/main/java/org/remus/giteabot/aiusage/AiUsageLog.java:62`), while the migrations add `TEXT` columns (`src/main/resources/db/migration/postgresql/V45__ai_usage_raw_request_response.sql:5`). This is not necessarily broken today, but the annotation suggests a max length that the schema does not enforce consistently across dialects, and future schema generation/tests may behave differently than production.

Suggestion: either remove the `length` attribute and rely on `columnDefinition`/migration-managed `TEXT`, or enforce truncation in `AiUsageService` to a clearly documented maximum.

---

max-california (approved):
## 🤖 Agentic Code Review

Summary: the feature is mostly well integrated — provider clients now pass request/response payloads through the existing audit path, persistence is guarded so logging failures won’t break inference, and the UI/doc updates are coherent. I found one correctness issue around configured payload length handling that can break persistence in production.

## BLOCKER

1. **Negative `ai-usage.max-raw-payload-length` can crash usage logging instead of safely truncating**
   - **Where:** `src/main/java/org/remus/giteabot/config/AiUsageProperties.java:31`, `src/main/java/org/remus/giteabot/aiusage/AiUsageService.java:221`
   - `getEffectiveMaxRawPayloadLength()` currently returns `Math.min(maxRawPayloadLength, 65535)`. If an operator sets `AI_USAGE_MAX_RAW_PAYLOAD_LENGTH` to a negative value, the effective length is negative.
   - `truncateRawPayload()` then executes `value.substring(0, maxLength)`, which throws `StringIndexOutOfBoundsException`. Because `recordUsage()` catches exceptions around the whole save path, the result is that *the entire usage record is dropped*, not just the raw payloads.
   - This is externally configurable and easy to hit via env var typo or bad config, so it should be clamped to a non-negative range, e.g. `Math.max(0, Math.min(maxRawPayloadLength, 65535))`, with a test for negative input.

## LOW

1. **UI renders a “Raw” button/modal even when both payloads are null**
   - **Where:** `src/main/resources/templates/usage/list.html:114`
   - When raw capture is enabled, every row shows a Details column and a modal trigger, even if the stored request/response are both `null` (for older rows, serialization failures, or requests recorded before the feature was enabled). That produces unnecessary empty modals.
   - Consider showing the button only when `entry.rawRequest != null || entry.rawResponse != null`, and render a muted placeholder otherwise.

---

PR #348, opened 2026-08-22
feature/anthropic-thinking

### LINKED ISSUE #346: Model reasoning is published into PR review comments (Anthropic provider)
We have the problem that with claude the reasoning/thinking text appears at top of the actual expected review text.
I asked claude to create a issue text for this problem.

The following text is generated by Claude:

---

# Model reasoning is published into PR review comments (Anthropic provider)

## Summary

With the Anthropic provider, both review workflows regularly publish the model's own reasoning above the review itself. The narration is part of the assistant message, so it ends up verbatim in the posted comment.

Affected:

- **PR Review** (`review`) — the non-agentic, diff-based workflow
- **Agentic PR Review** (`agentic-review`)

This is not fixable from the operator-editable system prompts with any reliability.

Two independent changes would address it. The first is the real fix; the second is a cheap, provider-independent safety net.

## What happens

Illustrative shape of a posted review comment, with a system prompt that requires the output to start at a fixed literal (`Summary:`):

> The refactored logic is equivalent, so no reset is needed. Behaviour is preserved. Tests confirm the expected output. No issues found.
>
> Summary: No issues found in this change.

Only the last line was supposed to be published. Everything above it is the model reasoning its way to that conclusion.

Another observed variant ends the narration with a sentence like "I have enough to write the review", followed by the correctly formatted review — all inside one assistant message.

## Why the system prompt cannot solve this

We iterated on both system prompts several times: an explicit ban on narration, a rule that the response must begin with a fixed literal, a list of forbidden sentence openers, and finally a worked negative/positive example. Each version reduced the frequency but none eliminated it.

The underlying reason is structural: extended thinking is never requested, so the model has no channel for its reasoning other than the text it emits — and every character it emits is published.

## Proposal A — support Anthropic extended thinking

When extended thinking is enabled, reasoning is returned in separate content blocks of type `thinking` rather than mixed into `text`.

**The response side already does the right thing.** In `AnthropicAiClient`:

- `extractText(...)` keeps only blocks where `type == "text"`.
- `interpret(...)` handles `text` and `tool_use` and ignores every other block type.

So `thinking` blocks would be discarded automatically and never reach the review comment. No change is needed on that path.

**What needs to change:**

1. `AnthropicRequest` — add the `thinking` object (`{"type": "enabled", "budget_tokens": N}`).
2. `AnthropicAiClient` — set it on the request when enabled.
3. `AiIntegration` — add per-integration fields (e.g. `thinkingEnabled`, `thinkingBudgetTokens`) plus a Flyway migration and the corresponding fields in the AI-integration settings form. This follows the existing pattern of `promptCachingEnabled`, `maxTokens` and `contextWindowTokens`.

### Suggested staging: non-agentic first

The two workflows differ substantially in effort, so they can be shipped separately.

**Non-agentic PR Review — small change.** `submitReviewPrompt(...)` is a single-shot call with no tool use, and its result goes through `extractText(...)`, which already filters to `text` blocks. Setting the `thinking` parameter is essentially all that is required.

**Agentic PR Review — needs more work.** When extended thinking is combined with tool use, the `thinking` block returned alongside a `tool_use` turn must be replayed unchanged (including its `signature`) in the assistant message when the tool results are sent back. Otherwise the API rejects the follow-up request.

Today that is not possible: `ChatTurn` carries only text, tool calls, stop reason and token counts, and the assistant history is rebuilt in `AnthropicAiClient` from `text` + `tool_use` blocks only. Enabling thinking without also capturing, persisting and replaying the thinking blocks would break the agent loop rather than fix it.

### Settings validation

`budget_tokens` must be smaller than `max_tokens`. The default `maxTokens` of 4096 leaves little room, so operators will need either guidance in the form or a validation rule.

## Proposal B — strip leading narration before posting

Independent of Proposal A, the posted comment could be sanitised: cut everything before the first occurrence of a configured start marker, and leave the text untouched when the marker is not found, so a review is never lost.

This is the same defensive pattern already used by `parseFormalReviewResult(...)`, which strips the trailing severity-classification JSON so it never leaks into the comment.

Choke points:

- **Agentic:** `AgentReviewService.formatReview(...)` — a single place, applied both on the normal path and on the budget-exhausted fallback.
- **Non-agentic:** the per-chunk result inside `reviewDiffWithChunking(...)` rather than `formatReviewComment(...)`, because with `totalChunks > 1` each chunk is a separate model response and each one carries its own narration. Stripping only at the end would leave the narration of chunks 2..n in place.

The marker should not be hardcoded, since the output format comes from the operator-editable system prompt. A per-bot or per-workflow setting (empty = disabled, current behaviour) would keep it opt-in.

tmseidel:


max-california (approved):
## 🤖 Agentic Code Review

Summary: the code changes are small and mostly well-covered by tests, but I found one correctness issue in the new feature gating. The new request fields are wired through cleanly, serialization is tested, and the response parsing already ignores non-`text` blocks as intended.

## BLOCKER

1. **Extended thinking is enabled for every Anthropic integration, even models the docs say do not support it**
   - **Files:** `src/main/java/org/remus/giteabot/ai/anthropic/AnthropicProviderMetadata.java:76`, `src/main/java/org/remus/giteabot/ai/anthropic/AnthropicAiClient.java:86`
   - The new properties are global and are passed unconditionally into every Anthropic client. Once `anthropic.extended-thinking.enabled=true`, requests for older/non-5.x models will always include:
     - `thinking: { "type": "adaptive" }`
     - `output_config: { "effort": ... }`
   - Your deployment docs explicitly say this requires Claude 5.x models, but the runtime code does not enforce that. That creates a production failure mode where an existing Anthropic integration using another model starts returning 4xx errors as soon as the global flag is turned on.
   - Suggested fix: gate these fields by model capability in `createClient(...)` or `AnthropicAiClient` itself, e.g. only enable for an allowlist/pattern of supported model names, or fail fast with a clear configuration error when the flag is on for an unsupported model.

## LOW

1. **`effort` is accepted verbatim with no validation**
   - **Files:** `src/main/java/org/remus/giteabot/config/AnthropicExtendedThinkingProperties.java:37`, `src/main/java/org/remus/giteabot/ai/anthropic/AnthropicAiClient.java:104`
   - Docs constrain `effort` to `low|medium|high|xhigh|max`, but the property is a free-form `String`. A typo will silently produce invalid provider requests.
   - Suggestion: validate at bind time (enum or `@Pattern`) and/or normalize case before sending.

---

PR #351, opened 2026-08-22
release/1.22.0

tmseidel:


max-california (approved):
## 🤖 Agentic Code Review

Summary: the PR looks generally solid and well-tested. The secret-clearing flows and process-environment scrubbing are sensible improvements, and the raw AI payload capture is appropriately opt-in. I found one correctness issue in the new configuration wiring.

## BLOCKER

1. **`ai-usage.*` properties are introduced but never exposed in `application.properties`, so the new raw-payload feature cannot actually be configured in normal deployments**
   - `src/main/java/org/remus/giteabot/config/AiUsageProperties.java`
   - `src/main/java/org/remus/giteabot/aiusage/UsageController.java`
   - `src/main/java/org/remus/giteabot/ai/AbstractAiClient.java`
   - `src/main/resources/application.properties`
   
   The code now depends on `AiUsageProperties` for:
   - enabling/disabling raw payload capture,
   - truncation length,
   - showing the “Details / Raw” column in `/usage`.

   But unlike the new Anthropic extended-thinking properties, there are no corresponding entries in `application.properties` documenting/binding `ai-usage.raw-payloads-enabled` and `ai-usage.max-raw-payload-length`. Since this project appears to rely on `application.properties` as the central contract for env-var-based deployment config, operators currently have no documented/defaulted way to turn this on via env vars in Docker/K8s style setups. In practice, the feature is effectively dead-on-arrival unless someone knows to supply Spring’s relaxed property names manually.

   Please add the missing properties, e.g. with env-backed defaults, alongside the other top-level config:
   - `ai-usage.raw-payloads-enabled=${AI_USAGE_RAW_PAYLOADS_ENABLED:false}`
   - `ai-usage.max-raw-payload-length=${AI_USAGE_MAX_RAW_PAYLOAD_LENGTH:65535}`

## LOW

1. **Good defensive change, but controller/service duplication could drift**
   - `src/main/java/org/remus/giteabot/admin/AiIntegrationController.java`
   - `src/main/java/org/remus/giteabot/admin/GitIntegrationController.java`
   - `src/main/java/org/remus/giteabot/admin/AiIntegrationService.java`
   - `src/main/java/org/remus/giteabot/admin/GitIntegrationService.java`

   The “one-way write” handling is implemented both in the controller and service. It works today, but the contract is split across layers. Consider centralizing all blank/new/clear resolution in the service and letting controllers pass raw form values only; that would reduce the chance of future divergence.

2. **`ProcessSupport.runInNewProcessGroup` hard-depends on `setsid`; worth calling out in deployment docs**
   - `src/main/java/org/remus/giteabot/util/ProcessSupport.java`

   The Linux path now requires the external `setsid` utility. The code handles absence with an `IOException`, but for minimal/distroless images this may surprise operators. Not a merge blocker because the failure mode is explicit, but I’d recommend documenting that dependency wherever runtime/container requirements are listed.

---

max-california (approved):
## 🤖 Agentic Code Review

Summary: overall this release looks solid. The token/API-key “clear vs keep” behavior is well thought through and the process isolation work is a meaningful security improvement. I only found one medium-risk issue around access to newly stored raw AI payloads.

## MEDIUM

### Raw request/response auditing can expose secrets in the admin UI and DB without any redaction
The new raw-payload capture stores full provider request/response JSON and exposes it directly in `/usage` when enabled (`src/main/java/org/remus/giteabot/ai/AbstractAiClient.java`, `src/main/java/org/remus/giteabot/aiusage/AiUsageService.java`, `src/main/resources/templates/usage/list.html`).

Even with the feature off by default, once enabled this will persist and display:
- full prompts and repository content,
- tool arguments/results,
- potentially user-supplied secrets included in prompts,
- provider-side sensitive response metadata.

Because the implementation serializes the provider request/response objects wholesale, there is no field-level redaction before persistence or display. This creates a durable sensitive-data sink in the application database and admin UI.

Suggested fix:
- redact or omit known sensitive fields before persistence (at minimum API/auth headers if they ever become part of request objects, and ideally tool outputs / prompt bodies behind a stricter opt-in),
- add an explicit warning in the UI near the feature toggle/docs that this stores sensitive prompt content,
- consider gating the modal behind a stronger permission boundary or separate endpoint if your admin surface is shared.

## LOW

### Nice improvement: clear/keep secret handling is now consistent and avoids double-encryption
The controller/service split for AI and Git integrations looks correct, and the new tests cover the important cases well (`AiIntegrationService`, `GitIntegrationService`, corresponding controller/E2E tests).

### Nice improvement: subprocess environment scrubbing and process-group cleanup materially reduce secret leakage risk
`ProcessSupport` plus the integration into validation/E2E runners is a meaningful hardening step. The tests around timeouts, background children, and UTF-8 truncation are especially helpful.

---

PR #355, opened 2026-08-24
Add i18N for the UI

tmseidel:
Add internationalization (i18n) support for the server-rendered UI

### Usage of AI
The translations were generated by deepseek-pro and were checked and improved by opus-5


max-california (approved):
## 🤖 Agentic Code Review

Summary: the i18n plumbing looks solid overall — locale resolution, message bundles, and most controller/template conversions are coherent, and the added tests cover the main login/locale flows. I found one correctness issue that will surface at runtime, plus one maintainability gap around locale handling.

## BLOCKER

1. **Missing message key causes delete success path to fail for event hooks**
   - **File:** `src/main/java/org/remus/giteabot/eventhook/EventHookController.java`
   - **Line:** around the new `flash.hookDeleted` lookup in `delete(...)`
   - The controller now uses `messageSource.getMessage("flash.hookDeleted", ...)`, but `messages.properties` in the diff does not define `flash.hookDeleted` (I can see `flash.hookSaved`, `flash.hookEnabled`, `flash.hookDisabled`, etc., but not this one).
   - Because this call does **not** provide a default message and does not catch `NoSuchMessageException`, a successful delete will now throw instead of redirecting with a flash message.
   - Please add `flash.hookDeleted` to the bundles (or supply a fallback/default message).

## MEDIUM

1. **Locale selector active-state check won’t mark Simplified Chinese as active**
   - **File:** `src/main/resources/templates/fragments/locale-selector.html`
   - **Line:** `th:classappend="${localeOption.code == #locale.toString()} ? 'active'"`
   - For most languages this works (`fr`, `de`, etc.), but for Chinese the supported code is `zh_CN` while `#locale.toString()` may resolve to `zh_CN_#Hans` / another JVM-specific variant when using `Locale.SIMPLIFIED_CHINESE`. That means the active language can fail to highlight even though the locale is selected.
   - Suggestion: compare normalized language tags/codes from a model attribute, or compare against `#locale.language`/country explicitly instead of raw `toString()`.

Overall, aside from the missing message key, this looks like a good and well-tested addition.

---

max-california (approved):
## 🤖 Agentic Code Review

Summary: solid i18n wiring overall, with good test coverage around locale resolution and login-page behavior. I found one functional regression on the setup page and one broader maintainability/correctness gap around untranslated hard-coded UI strings that are still visible after switching locales.

## BLOCKER

1. **Setup page no longer exposes any language switcher, so first-run UI cannot actually be localized by user choice**
   - **File:** `src/main/resources/templates/setup.html`
   - The PR adds a locale selector to `/login` and to authenticated pages via system settings, but not to `/setup`. That means the initial bootstrap page cannot be switched explicitly by a user, despite being translated and being one of the most important anonymous pages.
   - This is especially noticeable because `/setup` is shown before authentication and before a locale cookie exists; users are then limited to browser `Accept-Language` only.
   - Suggestion: include the same `fragments/locale-selector :: localeSelector('/setup')` used on the login page.

## MEDIUM

1. **A number of user-visible strings remain hard-coded in English, resulting in a partially localized UI**
   - **Files:** notably `src/main/resources/templates/layout.html`, plus likely others
   - Examples visible in the diff:
     - `layout.html`: `alt="AI Git Bot logo"`
     - `layout.html`: `data-tour-title="Dashboard"`, `data-tour-text="Your starting point: ..."`
     - same pattern for Bots / AI Integrations / Git Integrations / Usage / System settings
   - Because these strings are still rendered directly, users in non-English locales will get mixed-language pages. The tour/help text looks especially prominent.
   - This may not block the entire feature, but it undermines the stated goal of “Add i18N for the UI”.
   - Suggestion: either move these remaining strings into message bundles now, or clearly scope/document that tour/help text and certain accessibility strings are intentionally left English-only.

## LOW

1. **`lang` cookie name is very generic**
   - **File:** `src/main/java/org/remus/giteabot/config/I18nConfig.java`
   - `LOCALE_COOKIE = "lang"` increases the chance of collisions with other apps on the same host/path.
   - Suggestion: use an app-scoped cookie name such as `giteabot.lang`.

2. **Unused import**
   - **File:** `src/main/java/org/remus/giteabot/systemsettings/BuiltinToolRegistry.java`
   - `import java.util.Locale;` appears unused in the diff.
   - Minor cleanup.

---

max-california (approved):
## 🤖 Agentic Code Review

Summary: the i18n integration is generally well thought through and the new tests cover the main locale-resolution paths. I did find one behavioral regression in the workflow editor, plus a couple of maintainability/UX issues worth addressing.

## BLOCKER

1. **Workflow rows without parameters now render an expand/collapse chevron that controls no element**
   - **File:** `src/main/resources/templates/system-settings/workflow-configurations/workflows.html`
   - **Lines:** around 45–53, 77–79, 161–162
   - The comment says the collapse toggle is “only rendered when the workflow has params”, but the condition is `th:if="${row.workflow != null}"`, which is true for *all* registered workflows. For workflows with an empty params schema, the template still renders:
     - a chevron button with `data-bs-target="#wf-collapse-..."`, and
     - an empty collapsible body containing only “This workflow has no configurable parameters.”
   - That’s a regression in the editor UX: workflows that have nothing configurable now look collapsible/editable, and users can expand a meaningless empty section. If the intent was really “only when params exist”, this should be keyed off `row.localizedSchema.empty` / `row.workflow.paramsSchema().empty` instead of `row.workflow != null`.
   - Suggested fix: only render the toggle and collapse section when the schema is non-empty; render the “no configurable parameters” text inline in the header or body only when expanded content actually exists.

## MEDIUM

1. **Several newly added explanatory paragraphs remain hard-coded in English**
   - **Files:** e.g. `src/main/resources/templates/system-settings/bot-tools-selection.html:18-23`
   - A lot of visible strings were moved to `messages*.properties`, but some user-facing paragraphs are still plain English HTML text. Example in the built-in tool selection page:
     - “Whitelist of built-in agent tools ... runtime is rejected.”
   - This undermines the “Add i18N for the UI” goal because parts of the UI will stay untranslated even when the locale is switched.
   - I’d recommend grepping for remaining raw text in templates and either moving them into message keys or explicitly documenting that those pages are intentionally only partially localized.

## LOW

1. **`message.confirmDelete` is now used for actions with materially different consequences**
   - **File:** `src/main/resources/templates/system-settings/list.html`
   - Multiple delete confirmations that were previously specific are now collapsed to the generic `message.confirmDelete`, including deployment targets where the old text warned that bots referencing the target must be updated first.
   - This is not a correctness bug, but it loses important context for destructive actions. It would be better to keep resource-specific confirmation keys where there are special side effects.

2. **The locale selector drops existing query parameters**
   - **File:** `src/main/resources/templates/fragments/locale-selector.html`
   - Current links are built as `@{${baseUrl}(lang=${localeOption.code})}`. On pages that already have query parameters (filters, pagination, etc.), switching language will replace the query string instead of preserving it.
   - That’s acceptable on `/login` and `/setup`, but less ideal on pages like usage/event deliveries where users may already have filters applied. If you want a smoother UX, consider preserving the current request parameters and only overriding `lang`.

---

max-california (approved):
## 🤖 Agentic Code Review

Summary: the i18n plumbing looks broadly sound and the test coverage is good, but I found one functional gap where OAuth-mode pages remain effectively English-only because the localized text is still being supplied from the controller as a hard-coded string.

## BLOCKER

None.

## MEDIUM

### OAuth login error message is still hard-coded in English
`src/main/java/org/remus/giteabot/admin/OAuthLoginController.java:27`

The templates for `oauth-login-error.html` and `signed-out.html` were localized, but `OAuthLoginController#loginError()` still injects the main error body via:

```java
model.addAttribute("error", "OAuth sign-in failed. Check the server logs and your OAuth client configuration, then try again.");
```

Since the template renders `${error}` directly, this page will stay English regardless of the selected locale. That leaves the OAuth sign-in flow partially untranslated.

Suggested fix: inject `MessageSource` into `OAuthLoginController` and resolve this message through the bundle (similar to the other controllers), or move the whole sentence into the template with `th:text="#{...}"`.

## LOW

### Unused import in `BuiltinToolRegistry`
`src/main/java/org/remus/giteabot/systemsettings/BuiltinToolRegistry.java:11`

`java.util.Locale` is imported but not used. Small cleanup, but worth removing.

---

PR #358, opened 2026-08-25
fix: Show favicon on pages after login again

Postremus:
### What was done and why?
The favicon was not shown on any pages after login - like the dashboard - because html was incorrectly structured. The body tag was opened before the head tag (like: `<body><div><head>`). Chrome then just ignored everything inside head.

The head tag fragment was basically included into a section inside body.

Idea behind the fix is, that the layout now provides a fragment for the whole page (top level html tag); including both head and body tag in the layout fragment.

### AI Usage

I wrote the fragment in the layout and dashboard by hand. Opus then applied the same pattern to all other pages, and wrote the tests.

I manually confirmed that the pages still render correctly. The favicon is now shown on every page.


max-california (approved):
## 🤖 Agentic Code Review

Summary: This looks like a correct fix for the malformed document structure. Moving the shared fragment to the top-level `<html>` and updating the content pages to replace that element should ensure `<head>` is emitted before `<body>`, which matches the favicon symptom described. I didn’t find any functional regressions in the changed templates from inspection.

## Findings

### LOW
- **Test coverage is narrower than the scope of the template migration.**  
  The change updates 20+ templates, but the new regression test only exercises `dashboard.html` (`src/test/java/org/remus/giteabot/admin/DashboardControllerTest.java`). Since the fix pattern was applied broadly and partially generated by AI, it would be safer to add at least one more representative MVC test for another page family using the shared layout (for example bots list or system settings) to catch accidental malformed markup in another migrated template.

No blocking issues found.

---

tmseidel (approved):

tmseidel:
Thx for your contribution. LGTM 🚀

PR #361, opened 2026-08-26
fix(ai): report token-budget exhaustion instead of generic empty response

CaeruleusAqua:
## What has been done?

When an OpenAI-compatible API returns a choice without visible content (`message.content` null or blank) and `finish_reason=length`, `OpenAiClient.extractText` now produces an actionable message instead of the generic "empty response from AI":

> Unable to generate review - the model used its entire token budget without producing visible output (finish_reason=length; typical for reasoning models). Increase the max tokens setting of this AI integration.

Additionally:

- Blank content is now treated as empty. Previously it passed through and produced an empty review/comment body.
- The finish reason is included in the warning log line for easier diagnosis.

## Why?

Reasoning models (e.g. DeepSeek thinking variants served through OpenRouter) spend their completion budget on hidden reasoning tokens before producing visible output. When that budget is exhausted, `content` comes back empty and users only saw "Unable to generate review - empty response from AI.", which gives no hint about the actual cause or fix. The new message names the real problem and points at the max tokens setting of the AI integration.

## What has been tested?

- 4 new unit tests in `OpenAiClientTest`: budget-hint for null content + `finish_reason=length`, budget-hint for blank content + `length`, generic message retained for other finish reasons, and the happy path returning content.
- Full Maven suite green in Docker (Java 21): 1538 tests, 0 failures.

## How AI was used?

Diagnosis of the failure mode (root-cause analysis against the codebase and OpenRouter reasoning-token behavior) and implementation were done with OpenCode (model: ox-alpha); all changes were reviewed and verified by a human.

### DIFF
--- a/src/main/java/org/remus/giteabot/ai/openai/OpenAiClient.java
+++ b/src/main/java/org/remus/giteabot/ai/openai/OpenAiClient.java
@@ -261,10 +261,10 @@ public class OpenAiClient extends AbstractAiClient {
             String finishReason = firstChoice != null ? firstChoice.getFinishReason() : null;
             log.warn("Empty response from OpenAI API (finish_reason={})", finishReason);
             if ("length".equals(finishReason)) {
-                return "Unable to generate " + context
-                        + " - the model used its entire token budget without producing visible output"
-                        + " (finish_reason=length; typical for reasoning models)."
-                        + " Increase the max tokens setting of this AI integration.";
+                return """
+                        Unable to generate %s - the model used its entire token budget without producing visible output \
+                        (finish_reason=length; typical for reasoning models). \
+                        Increase the max tokens setting of this AI integration.""".formatted(context);
             }
             return "Unable to generate " + context + " - empty response from AI.";
         }
--- a/src/main/java/org/remus/giteabot/ai/openai/OpenAiClient.java
+++ b/src/main/java/org/remus/giteabot/ai/openai/OpenAiClient.java
@@ -247,21 +247,29 @@ public class OpenAiClient extends AbstractAiClient {
                 .body(OpenAiResponse.class);
     }
 
-    private String extractText(OpenAiRequest request, OpenAiResponse response, String context) {
+    String extractText(OpenAiRequest request, OpenAiResponse response, String context) {
         if (response == null || response.getChoices() == null || response.getChoices().isEmpty()) {
             log.warn("Empty response from OpenAI API");
             return "Unable to generate " + context + " - empty response from AI.";
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/remus/giteabot/ai/openai/OpenAiClient.java
```diff
@@ -247,21 +247,29 @@ private OpenAiResponse executeRequest(OpenAiRequest request) {
...
+                : null;
+        if (content == null || content.isBlank()) {
+            String finishReason = firstChoice != null ? firstChoice.getFinishReason() : null;
+            log.warn("Empty response from OpenAI API (finish_reason={})", finishReason);
+            if ("length".equals(finishReason)) {
+                return "Unable to generate " + context
+                        + " - the model used its entire token budget without producing visible output"
+                        + " (finish_reason=length; typical for reasoning models)."
+                        + " Increase the max tokens setting of this AI integration.";
+            }
```
Postremus: Maybe using a text block with line continuation would be a bit nicer to look at?

```
"""
Unable to generate  % \
- the model used its entire token budget without producing visible output \
(finish_reason=length; typical for reasoning models). \
Increase the max tokens setting of this AI integration.""".formatted(context);
```

CaeruleusAqua: Applied in 304097a - text block with line continuation reads much better, thanks.

max-california (approved):
## 🤖 Agentic Code Review

Summary: this change looks correct and well targeted. The new `finish_reason=length` handling matches the reported failure mode, blank content is now treated consistently, and the added tests cover the important branches. I did not find any blocking correctness or security issues.

## Findings

### LOW
- **Package-private visibility on `extractText` is a small encapsulation regression**  
  In `src/main/java/org/remus/giteabot/ai/openai/OpenAiClient.java:250`, `extractText` was changed from `private` to package-private solely so the test can call it directly. The logic is fine, but widening visibility for testability slightly weakens encapsulation of production code.  
  **Suggestion:** keep the method `private` and test it through a higher-level entry point, or extract the response-parsing logic into a small helper/component with its own focused unit tests.

- **The new tests don’t cover whitespace-only content explicitly**  
  The implementation correctly uses `isBlank()`, but `src/test/java/org/remus/giteabot/ai/openai/OpenAiClientTest.java` only tests `""`, not `"   "` or similar whitespace-only content. Since the PR description explicitly calls out “null or blank”, it would be good to lock that behavior in.  
  **Suggestion:** add one test case with whitespace-only content and `finish_reason=length`.

---

tmseidel (approved):
LGTM

PR #363, opened 2026-08-28
feat(ui): add reusable CodeMirror 6 JSON editor with Prettier formatting

### LINKED ISSUE #359: feature: Enhancements for the mcp config editor
Right now the mcp config is a simple textarea.
This works fine when entering a single or maybe 2 mcp configs.

Some ideas to improve it:
* Add line numbers on the left side of the editor panel.
* Add a button to pretty print the config. This would replace the current coontent of the text area with a pretty printed version of the json

* When saving an invalid json, a generic error message is shown. However, at least jackson provides detailed information at which line number and position the error occured. Could this also be shown.

Image

tmseidel:
- Add src/main/resources/static/js/json-editor.js with JSON linting,
  dark-mode awareness, and a "Format JSON" button.
- Extend layout.html with layoutWithExtras to inject per-page head and
  script fragments.
- Wire the editor into system-settings/mcp-form.html and
  event-hooks/form.html.
- Update e2e tests to fill the CodeMirror .cm-content element.

image

### DIFF
--- a/src/main/resources/static/js/json-editor.js
+++ b/src/main/resources/static/js/json-editor.js
@@ -11,11 +11,12 @@
  * so that standard form submission continues to work.
  */
 
-import { EditorView, basicSetup } from "https://esm.sh/codemirror@6.0.1";
-import { json, jsonParseLinter } from "https://esm.sh/@codemirror/lang-json@6.0.1";
-import { linter, lintGutter } from "https://esm.sh/@codemirror/lint@6.8.1";
-import { oneDark } from "https://esm.sh/@codemirror/theme-one-dark@6.1.2";
-import { placeholder as cmPlaceholder } from "https://esm.sh/@codemirror/view@6.33.0";
+import { EditorView, basicSetup } from "https://esm.sh/codemirror@6.0.1?deps=@codemirror/state@6.4.1";
+import { Compartment } from "https://esm.sh/@codemirror/state@6.4.1";
+import { json, jsonParseLinter } from "https://esm.sh/@codemirror/lang-json@6.0.1?deps=@codemirror/state@6.4.1";
+import { linter, lintGutter } from "https://esm.sh/@codemirror/lint@6.8.1?deps=@codemirror/state@6.4.1";
+import { oneDark } from "https://esm.sh/@codemirror/theme-one-dark@6.1.2?deps=@codemirror/state@6.4.1";
+import { placeholder as cmPlaceholder } from "https://esm.sh/@codemirror/view@6.33.0?deps=@codemirror/state@6.4.1";
 import * as prettier from "https://esm.sh/prettier@3.3.3/standalone";
 import * as estreePlugin from "https://esm.sh/prettier@3.3.3/plugins/estree";
 import * as babelPlugin from "https://esm.sh/prettier@3.3.3/plugins/babel";
@@ -34,6 +35,10 @@ function jsonLinter() {
 
 function isDarkMode() {
     return document.documentElement.getAttribute('data-bs-theme') === 'dark';
+}
+
+function themeExtension() {
+    return isDarkMode() ? oneDark : [];
 }
 
 async function formatJson(text) {
@@ -101,15 +106,20 @@ export function initJsonEditor(textarea, options) {
         extensions.push(cmPlaceholder(textarea.placeholder));
     }
 
-    if (isDarkMode()) {
-        extensions.push(oneDark);
-    }
+    const themeCompartment = new Compartment();
+    extensions.push(themeCompartment.of(themeExtension()));
 
     const view = new EditorView({
         doc: textarea.value,
         extensions: extensions,
         parent: wrapper
     });
+
+    // Reconfigure the editor theme when the app theme is toggled at runtime.
+    const applyTheme = () => {
+        view.dispatch({ effects: themeCompartment.reconfigure(themeExtension()) });
+    };
+    document.addEventListener('themeChanged', applyTheme);
 
     // Hide the original textarea but keep it in the DOM for form submission.
     textarea.classList.add('d-none');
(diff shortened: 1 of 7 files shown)


max-california (approved):
## 🤖 Agentic Code Review

Summary: nice UI improvement overall, but I found one functional regression in dark-mode handling. The editor only reads the theme once at initialization, so after the user toggles the site theme the CodeMirror instance keeps the old theme and becomes visually inconsistent with the rest of the page.

## BLOCKER

1. **CodeMirror theme does not update when the app theme is toggled**
   - **Files:** `src/main/resources/static/js/json-editor.js`, `src/main/resources/static/js/theme-init.js`
   - `initJsonEditor()` decides whether to add `oneDark` only once via `isDarkMode()` during construction. However, the app’s theme is explicitly mutable at runtime (`window.toggleTheme()` in `theme-init.js` updates `data-bs-theme` after page load).
   - Result: if a user opens one of these forms and then toggles light/dark mode, the rest of the page updates but the editor does not. In dark mode this can easily leave a light editor embedded in a dark page (or vice versa), which is a visible correctness/UI regression for the new component.
   - Suggestion: either
     - listen for theme changes and reconfigure the editor theme dynamically, or
     - expose a custom event from `theme-init.js` whenever `data-bs-theme` changes and have `json-editor.js` react to that event.

## LOW

1. **`extraHead` fragments are currently empty in both adopting templates**
   - **Files:** `src/main/resources/templates/system-settings/mcp-form.html`, `src/main/resources/templates/event-hooks/form.html`
   - Both pages pass `~{::extraHead}` but the fragment only contains a comment. This isn’t harmful, but it adds ceremony without current value.
   - Suggestion: drop `extraHead` on these pages until it is actually needed, while keeping `layoutWithExtras` available for future use.

2. **E2E selectors now depend on a generated wrapper ID**
   - **Files:** `tests/e2e/pr-326/create-mcp-configuration-and-select-tools.spec.ts`, `tests/e2e/pr-326/preserve-selected-tools-across-edit.spec.ts`
   - The tests target `#jsonContent-editor .cm-content`, which couples them to `json-editor.js`’s implementation detail of generating `textarea.id + '-editor'`.
   - A less brittle selector would target relative to the original field/container, or add a stable `data-testid` around the editor host.
   - Not blocking, but it will make refactors of the wrapper structure harder.

---

max-california (dismissed):
## 🤖 Agentic Code Review

Summary: the overall approach looks sound, and the server-side validation still protects the new JSON inputs. I found one correctness issue in the new reusable editor that can break forms when a page has more than one submit action.

## BLOCKER

1. **Pressing Enter inside the CodeMirror editor no longer triggers the form’s implicit submit**
   - **File:** `src/main/resources/static/js/json-editor.js`
   - **Lines:** around 109–116 / editor initialization
   - The implementation hides the original `<textarea>` (`textarea.classList.add('d-none')`) and replaces it with a `div`-based CodeMirror editor. That changes native form behavior: when a form has a single-line text input and a textarea, users often rely on pressing Enter from the textarea or focus flow that still leaves implicit submit working predictably. With CodeMirror, key events stay inside the editor widget and no longer participate in native form semantics.
   - This is especially risky on `system-settings/mcp-form.html`, where the primary action is “Save and select tools”. If keyboard users type JSON and hit Enter expecting submission, nothing happens.
   - Suggestion: either document and test the changed keyboard behavior, or better, add an explicit keybinding/handler for `Mod-Enter` (or similar) to submit the nearest form, and keep the visible submit affordance accessible.

## MEDIUM

1. **Event listeners are never cleaned up, which makes `initJsonEditor` non-reusable in dynamic pages**
   - **File:** `src/main/resources/static/js/json-editor.js`
   - **Lines:** around 121–138 and 157–166
   - `initJsonEditor` registers listeners on `document`, `textarea`, `formatButton`, and `view.dom`, but never unregisters them. That is fine for today’s static pages, but the helper is introduced as a reusable module and returns the `EditorView`; in any future partial-render/modal/turbo-like flow this will leak listeners and can re-run theme updates/format handlers on detached editors.
   - Suggestion: either expose a small destroy wrapper (remove listeners + `view.destroy()`), or attach listeners via CodeMirror extensions where lifecycle is managed with the editor.

## LOW

1. **`extraHead` fragments are currently empty in both consuming pages**
   - **Files:** `src/main/resources/templates/system-settings/mcp-form.html`, `src/main/resources/templates/event-hooks/form.html`
   - The new `layoutWithExtras` abstraction is useful, but both new `extraHead` fragments only contain comments. That adds template complexity without current benefit.
   - Suggestion: omit the empty fragments for now and pass `null` for `extraHead`, or keep one example only if this is meant as a pattern for future pages.

2. **E2E comments now describe the old DOM structure**
   - **Files:** `tests/e2e/pr-326/create-mcp-configuration-and-select-tools.spec.ts`, `tests/e2e/pr-326/preserve-selected-tools-across-edit.spec.ts`
   - The comments still say “textarea#jsonContent” / “multi-form field”, but the test now targets `#jsonContent-editor .cm-content`.
   - Suggestion: update the comments so future maintainers understand the test is intentionally driving CodeMirror, not the hidden textarea.

---

tmseidel:
@max-california This is a multiline-textearea. Not submitting the form when hitting enter is the desired behavior

max-california:
## 🤖 Follow-up

Agreed — for a multiline textarea/CodeMirror editor, **Enter should insert a newline, not submit the form**.

In this PR, I don’t see any code that overrides Enter to submit:

- `src/main/resources/static/js/json-editor.js` only adds:
  - sync from CodeMirror back to the hidden `<textarea>`
  - JSON linting
  - theme switching
  - a format action
  - a `keydown` handler for **Shift+Alt+F** only

There is **no Enter key handler** in the new editor code, so the expected behavior remains “newline on Enter”.

So if my earlier concern was about Enter not submitting, you can disregard it for these JSON editor fields — **not submitting on Enter is the correct UX here**.

---

PR #364, opened 2026-08-28
feat(ai): surface provider errors from OpenAI API responses

### LINKED ISSUE #362: feature: Log unmapped json properties in raw usage log
https://github.com/tmseidel/ai-git-bot/pull/347 introduced the ability, to log the input and output of the model.

However, this logging is based upon the already created OpenAiRequest request and OpenAiResponse response.
https://github.com/tmseidel/ai-git-bot/pull/347/changes#diff-9aace423499b495df10f399dfd590cecb69fa23d9f540581e7b39fce9ad2408fR250

OpenRouter has the quirk, that it adds another stop reason - "error".
When this stop reason is used however another json object is added as response to the completion api, describing the error in more detail.
https://openrouter.ai/docs/api_reference/errors-and-debugging#chat-completions-/api/v1/chat/completions

This "error" property however is removed, since OpenAiRequest and OpenAiResponse don't know about it.
```
{"choices":[{"finish_reason":"error","index":0,"message":{"content":null,"role":"assistant","tool_calls":null}}],"id":"gen-1787823563-6MMoJYv0lz80hJrrxL3J","model":"deepseek/deepseek-v4-pro-0813","object":"chat.completion","usage":{"completion_tokens":838,"prompt_tokens":15765,"total_tokens":16603}}
```

Would it be possible to log the full raw response of the model? Or we could add openrouter as a specific ai provider, but that might be overkill for now. Logging the raw response (string) might benefit all llms.

tmseidel:
Add an OpenRouter-compatible Error type to OpenAiResponse with code, message, and metadata fields. Check for finish_reason=error before falling through to the empty-response path, extracting the provider error message (choice-level → response-level → "unknown provider error") and surfacing it in a user-friendly reply. Add unit tests for the error, fallback, and missing-error cases, plus a JSON serialization check.

### DIFF
--- a/src/main/java/org/remus/giteabot/ai/openai/OpenAiClient.java
+++ b/src/main/java/org/remus/giteabot/ai/openai/OpenAiClient.java
@@ -257,8 +257,18 @@ public class OpenAiClient extends AbstractAiClient {
         String content = firstChoice != null && firstChoice.getMessage() != null
                 ? firstChoice.getMessage().getContent()
                 : null;
+        String finishReason = firstChoice != null ? firstChoice.getFinishReason() : null;
+        if ("error".equals(finishReason)) {
+            OpenAiResponse.Error error = firstChoice.getError() != null
+                    ? firstChoice.getError()
+                    : response.getError();
+            String errorMessage = error != null && error.getMessage() != null
+                    ? error.getMessage()
+                    : "unknown provider error";
+            log.warn("OpenAI API returned finish_reason=error: {}", errorMessage);
+            return "Unable to generate %s - provider returned an error (%s).".formatted(context, errorMessage);
+        }
         if (content == null || content.isBlank()) {
-            String finishReason = firstChoice != null ? firstChoice.getFinishReason() : null;
             log.warn("Empty response from OpenAI API (finish_reason={})", finishReason);
             if ("length".equals(finishReason)) {
                 return """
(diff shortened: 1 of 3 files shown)


max-california (approved):
## 🤖 Agentic Code Review

Summary: the change looks correct and well-targeted. It fixes a real misclassification bug by handling `finish_reason=error` before the generic empty-response path, models the provider error payload on both response and choice levels, and adds good coverage for the new branches. I didn’t find any blocking correctness or security issues.

## Findings

### LOW
- **`src/test/java/org/remus/giteabot/ai/openai/OpenAiClientTest.java:159`**  
  The serialization test only checks raw string containment. That proves the field is emitted, but it’s brittle and won’t catch shape regressions cleanly. A slightly stronger test would serialize and then parse back to a tree/assert on `choice[0].error.code` and `message`, which is less sensitive to formatting and unrelated serializer changes.

### LOW
- **`src/main/java/org/remus/giteabot/ai/openai/OpenAiClient.java:261`**  
  `extractText(...)` now handles provider errors, but `interpret(...)` for tool-calling requests still treats all non-empty `choices` the same and does not surface `finish_reason=error`. If OpenRouter can return the same error shape for tool-enabled calls, this path will still silently degrade into an empty text/`StopReason.OTHER` result. Not necessarily part of this PR’s scope, but worth aligning to avoid inconsistent behavior between plain chat/review and tool-call flows.

---

PR #368, opened 2026-09-01
feat(git): add secure Gitea SSH transport

CaeruleusAqua:
## What has been done?

- Added opt-in Gitea SSH transport with host-key confirmation, managed remote-key lifecycle, and isolated SSH credentials for Git commands.
- Deepened the repository checkout seam so providers resolve credential-free repository remotes while WorkspaceService resolves remotes and credentials exactly once.
- Added stale-form protection, optimistic versions, pessimistic deletion/assignment fencing, retryable cleanup, and safe HTTP fallback during deletion.
- Hardened SSH remote parsing and known_hosts endpoint binding, including fork-safe PR checkout fallback.
- Added the Gitea SSH UI, localized messages, documentation, and H2/PostgreSQL migrations V47 through V51.

## Why?

Gitea installations commonly expose Git over SSH, but the bot previously supported HTTP credentials only. This adds SSH without embedding secrets in clone URLs or workspaces and makes key setup, replacement, and deletion safe across failures, stale forms, retries, and concurrent application instances.

## What has been tested?

- Maven clean verify in a Linux Maven 21 container: 1,674 tests, 0 failures, 0 errors, 0 skipped.
- Focused repository, SSH, workspace, integration lifecycle, concurrency, bot, migration, localization, and E2E promotion tests.
- H2 Flyway migrations through V51.
- PostgreSQL 17.10 Flyway smoke test from V50 to V51 with an existing integration row and verified defaults.
- docker compose config validation and git diff whitespace checks.
- No live Gitea/OpenSSH clone-fetch-push roundtrip was performed.

## How AI was used?

AI assisted with implementation, regression-test generation, plan-to-code completion audits, and independent standards/spec reviews. The final code was verified with the full automated suite and both Flyway dialects.

PR #370, opened 2026-09-02
refactor(repository): centralize checkout resolution

CaeruleusAqua:
## What has been done?

- Added provider-owned resolution of complete, credential-free repository remotes.
- Changed WorkspaceService to resolve the remote and credentials exactly once before allocating a workspace.
- Updated every production checkout caller and preserved pull-request ref fallback for fork-based unit-test and E2E promotion workflows.
- Added focused remote-resolution, failure-path, exact-once, and caller tests.

## Why?

This extracts the repository checkout prerequisite from #368 into a reviewable change. Keeping provider-specific remote resolution behind RepositoryApiClient also prepares SSH transport without exposing credentials in clone URLs or coupling workflows to provider URL formats.

## What has been tested?

- WorkspaceServiceTest in the Maven Java 21 container.
- Full mvn -q clean verify in the Maven Java 21 container: 1,546 tests passed.

## How AI was used?

AI helped reconstruct this focused change from #368, update affected callers and tests, and review the final diff for scope, repository standards, and regressions.

### DIFF
--- a/src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java
+++ b/src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Locale;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -375,19 +376,17 @@ public class WorkspaceService {
      * private temporary parent, so the token is never stored inside the cloned
      * repository. Returns {@code null} for local paths or blank tokens.
      */
-    Path createCredentialsFile(String repositoryRemote, String token, Path workspaceDir)
-            throws IOException {
-        return createCredentialsFile(repositoryRemote, null, token, workspaceDir);
-    }
-
     Path createCredentialsFile(String repositoryRemote, String username, String token,
                                Path workspaceDir) throws IOException {
-        if (token == null || token.isBlank()
-                || repositoryRemote.startsWith("file://") || repositoryRemote.startsWith("/")) {
+        if (token == null || token.isBlank()) {
             return null;
         }
-        String protocol = repositoryRemote.startsWith("https://") ? "https" : "http";
-        String baseUrl = repositoryRemote.replaceFirst("https?://", "");
+        String lowerRemote = repositoryRemote.toLowerCase(Locale.ROOT);
+        if (lowerRemote.startsWith("file://") || repositoryRemote.startsWith("/")) {
+            return null;
+        }
+        String protocol = lowerRemote.startsWith("https://") ? "https://" : "http://";
+        String baseUrl = repositoryRemote.substring(protocol.length());
         if (baseUrl.endsWith("/")) {
             baseUrl = baseUrl.substring(0, baseUrl.length() - 1);
         }
@@ -401,7 +400,7 @@ public class WorkspaceService {
         restrictToOwner(credentialsFile, false);
         String credentialUsername = username == null || username.isBlank() ? "oauth2" : username;
         Files.writeString(credentialsFile,
-                protocol + "://" + credentialUsername + ":" + token + "@" + host + "\n");
+                protocol + credentialUsername + ":" + token + "@" + host + "\n");
         return credentialsFile;
     }
 
--- a/src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java
+++ b/src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java
@@ -1,6 +1,8 @@
 package org.remus.giteabot.agent.validation;
 
 import lombok.extern.slf4j.Slf4j;
+import org.remus.giteabot.repository.RepositoryApiClient;
+import org.remus.giteabot.repository.model.RepositoryCredentials;
 import org.remus.giteabot.util.ProcessSupport;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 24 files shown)


max-california (dismissed):
## 🤖 Agentic Code Review

Summary: the refactor looks sound overall and the test coverage is good, but I found one correctness regression in checkout resolution that will reject valid configured clone URLs.

## BLOCKER

1. **Valid uppercase HTTP(S) clone URLs are now rejected**
   - **Where:** `src/main/java/org/remus/giteabot/repository/RepositoryApiClient.java:47`
   - `getRepositoryRemote()` now validates the scheme with a case-sensitive check:
     ```java
     "http".equals(base.getScheme()) || "https".equals(base.getScheme())
     ```
     But URI schemes are case-insensitive. A configured value like `HTTPS://git.example.com` is valid and previously would have worked; now it throws `IllegalStateException` and aborts workspace preparation.
   - The new `GiteaApiClientTest` even codifies this behavior as expected, which suggests the regression may be unintentional.
   - Recommendation: normalize the scheme before comparison (e.g. `toLowerCase(Locale.ROOT)`) or compare case-insensitively, while still rejecting non-HTTP(S) schemes and credential-bearing URLs.

## LOW

1. **Test name/expectation appears to bless a non-canonical but valid URI as invalid**
   - **Where:** `src/test/java/org/remus/giteabot/gitea/GiteaApiClientTest.java`
   - `getRepositoryRemote_rejectsNonCanonicalHttpScheme` treats uppercase `HTTPS://...` as an invalid scheme. That's not non-HTTP; it's a valid HTTP URI with different casing.
   - If the intent is to reject malformed or unsupported schemes, a better test would use something actually unsupported like `ssh://...` or `git@...`.

---

tmseidel:
@max-california This is no regression, in the former code of `org.remus.giteabot.agent.validation.WorkspaceService` in method  `buildCloneUrl` the `HTTPS` url were not rejected but the method returned a `http` which is also not correct. So yes, protocol-names should be handled case-insenstively but this is not a regression to the former implementation

src/main/java/org/remus/giteabot/repository/RepositoryApiClient.java
```diff
@@ -29,9 +30,32 @@ default String getBaseUrl() {
         return getCredentials().baseUrl();
     }
 
-    /** Returns the web/clone URL of the repository provider (e.g. {@code https://github.com}). */
-    default String getCloneUrl() {
-        return getCredentials().cloneUrl();
+    /** Returns the credential-free HTTP clone base URL (e.g. {@code https://github.com}). */
+    String getCloneUrl();
```
tmseidel: What's the purpose of removing the default implementation and duplicating the code in every Implementation?

src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java
```diff
@@ -352,37 +370,24 @@ void cleanupWorkspace(WorkspaceSetup setup) {
...
+    Path createCredentialsFile(String repositoryRemote, String username, String token,
+                               Path workspaceDir) throws IOException {
         if (token == null || token.isBlank()
-                || cloneBaseUrl.startsWith("file://") || cloneBaseUrl.startsWith("/")) {
+                || repositoryRemote.startsWith("file://") || repositoryRemote.startsWith("/")) {
             return null;
         }
-        String protocol = cloneBaseUrl.startsWith("https://") ? "https" : "http";
-        String baseUrl = cloneBaseUrl.replaceFirst("https?://", "");
+        String protocol = repositoryRemote.startsWith("https://") ? "https" : "http";
```
tmseidel: we should handle this one also case-insenetively, so toLowerCase and in the nextline not `replaceFirst` but `substring`, which should be more robust and also faster (no regexp)
```java
String protocol = cloneBaseUrl.toLowerCase(Locale.ROOT).startsWith("https://") ? "https://" : "http://";
String baseUrl = cloneBaseUrl.substring(protocol.length());
```
and later 
````java
Files.writeString(credentialsFile,
                protocol + credentialUsername + ":" + token + "@" + host + "\n");
``

File: src/main/java/org/remus/giteabot/repository/RepositoryApiClient.java [hunk 3]
```diff
@@ -29,9 +30,32 @@ default String getBaseUrl() {
...
+
+    /**
+     * Resolves the complete, credential-free HTTP Git remote for one repository.
+     */
+    default String getRepositoryRemote(String owner, String repo) {
```
tmseidel, line 42: Wouldn't it make sense to return an `URL` here instead of a `String`, so that we can remove the partly duplicated parsing-code in `createCredentialsFile` of the class `WorkspaceService`?

tmseidel (changes requested):

src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java
```diff
@@ -352,37 +370,24 @@ void cleanupWorkspace(WorkspaceSetup setup) {
...
-        return String.format("%s://%s/%s/%s.git", protocol, baseUrl, owner, repo);
-    }
-
     /**
      * Writes a git credential-store file <em>outside</em> the workspace in its
      * private temporary parent, so the token is never stored inside the cloned
      * repository. Returns {@code null} for local paths or blank tokens.
      */
-    Path createCredentialsFile(String cloneBaseUrl, String token, Path workspaceDir)
+    Path createCredentialsFile(String repositoryRemote, String token, Path workspaceDir)
```
tmseidel: Can you remove this method? - It's only referenced by tests and we don't need to be backward-compatible.

src/main/java/org/remus/giteabot/repository/RepositoryApiClient.java
CaeruleusAqua: I kept String intentionally because this PR is the base for the SSH follow-up. Gitea returns both ssh:// URLs and SCP-style remotes such as git@host:owner/repo.git; the latter cannot be represented by java.net.URL. Keeping the provider-owned remote opaque avoids changing the API again in the next PR.

src/main/java/org/remus/giteabot/repository/RepositoryApiClient.java
CaeruleusAqua: Restored the default implementation in 5413886 and removed the duplicate overrides from all four provider clients and the test stub.

src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java
CaeruleusAqua: Updated in 5413886: scheme detection now uses Locale.ROOT, the credential URL uses a lowercase protocol prefix, and substring replaces the regex. The test covers uppercase HTTPS.

src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java
CaeruleusAqua: Removed in 5413886. Tests now call the single username-aware method directly.

CaeruleusAqua:
Review feedback is addressed in 5413886:

- restored the default getCloneUrl implementation and removed duplicate overrides
- accepted and canonicalized case-insensitive HTTP(S) schemes
- replaced regex parsing in credential-file creation and removed its test-only overload
- kept repository remotes as String for the SSH follow-up's SCP-style git@host:path values

Focused tests and a full mvn -q clean verify pass locally. @tmseidel @max-california please re-review.

max-california:
## 🤖 Follow-up

Re-reviewed: this looks good to me overall. The refactor is coherent, and the important regression risk I’d have watched for — losing the fork-safe PR ref fallback — appears intentionally preserved, including for unit-test/E2E flows.

## Summary
- `WorkspaceService` now resolves remote + credentials exactly once up front, which is cleaner and avoids passing clone/token pairs around callers.
- `RepositoryApiClient.getRepositoryRemote()` adds good validation for credential-free HTTP(S) clone bases.
- Callers were updated consistently.
- Tests cover the main new behaviors: exact-once resolution, failure-before-allocation, remote validation, and PR fallback cases.

## Findings
### No blocking issues found

### Minor observation
- `WorkspaceService.createCredentialsFile()` still treats any non-`file://` / non-absolute-path remote as HTTP and derives either `https://` or `http://` from the prefix. That’s fine with the current contract because `getRepositoryRemote()` now guarantees HTTP(S) for provider-owned remotes, and local path tests still pass through. So I don’t see a bug here, just a coupling worth keeping in mind if SSH remotes are introduced later — this helper will need to reject or branch on those explicitly.

Given the current PR scope and tests, I’d approve this re-review.

---

tmseidel (approved):
LGTM. Thanks a lot for your work.. 🚀

PR #372, opened 2026-09-03
fix(security): clarify optional credential encryption

CaeruleusAqua:
## What has been done?

- Updated UI help in all seven locales to state that credential encryption requires APP_ENCRYPTION_KEY.
- Removed encryption wording from keep-existing-value hints and kept HTML out of text-only messages.
- Corrected the same claims in templates, deployment and user documentation, READMEs, pitch material, and LLM entry points.
- Added a locale regression test for text-only help markup.

## Why?

EncryptionService intentionally stores values as plain text when APP_ENCRYPTION_KEY is absent, but several user-facing messages claimed encryption was unconditional. Operators need accurate storage and deployment guidance.

## What has been tested?

- I18nConfigTest and EncryptionServiceTest in the Maven Java 21 container.
- Full mvn -q clean verify in the Maven Java 21 container.
- Verified all seven bundles contain the conditional wording and no text-only markup.
- Verified no SSH-specific behavior or documentation was added.

## How AI was used?

AI helped audit user-facing encryption claims, extract the independent correction from superseded #368, preserve locale and rendering conventions, and review the final scope.


tmseidel (approved):
LGTM

max-california (approved):
## 🤖 Agentic Code Review

Summary: this looks like a solid, correctness-focused documentation/UI fix. I checked the runtime behavior in `EncryptionService` and the new wording now matches the implementation: without `APP_ENCRYPTION_KEY`, secrets are stored as plain text, and the added i18n regression test usefully guards against reintroducing HTML into text-only help strings. I did not find any blocking correctness or security regressions in the changed areas.

## Findings

_No findings._

---

PR #373, opened 2026-09-03
fix(e2e): isolate suite promotion failures

CaeruleusAqua:
## What has been done?

- Converted unexpected runtime failures across suite promotion into a failed promotion outcome.
- Added a service regression test for provider failures.
- Added a workflow regression test proving that failed promotion does not replace an already successful E2E result.

## Why?

Suite promotion is a follow-up action and promises to report failures without rolling back the original workflow. An escaped runtime exception could instead make an otherwise successful E2E run fail.

## What has been tested?

- SuitePromotionServiceTest and E2ETestWorkflowTest in the Maven Java 21 container.
- Full mvn -q clean verify in the Maven Java 21 container.

## How AI was used?

AI helped isolate this fix from superseded #368, trace the workflow status path, add focused regression tests, and review the final three-file diff.

### DIFF
--- a/src/main/java/org/remus/giteabot/prworkflow/e2e/promotion/SuitePromotionService.java
+++ b/src/main/java/org/remus/giteabot/prworkflow/e2e/promotion/SuitePromotionService.java
@@ -77,6 +77,16 @@ public class SuitePromotionService {
 
     public Outcome promote(Bot bot, PrWorkflowRun run, PrTestSuite suite,
                            String repoOwner, String repoName, String featureBranch) {
+        try {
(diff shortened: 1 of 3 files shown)


max-california (approved):
## 🤖 Agentic Code Review

Summary: this change looks correct and aligned with the workflow contract. Wrapping unexpected promotion-time runtime exceptions into `Outcome.failure(...)` prevents a follow-up action from flipping an otherwise successful E2E run to failed, and the added tests cover both the service-level regression and the workflow-level status behavior.

No blocking issues found.

---

tmseidel (approved):
LGTM

PR #376, opened 2026-09-04
feat(mcp): Allow referencing ENV variables in mcp config 

### LINKED ISSUE #360: feature: Prevent MCP Authorization Keys from being shown in plain text
Today I added context7 to my configuration. This mcp requires an api key.
The only way (that I could find) to configure the api key was to input it as plain text into the mcp config.

Idea:
Either:
* add the possibility to expand environment variables in the mcp config. 
* Or add hashicorp vault support, and support expansion of vault paths/keys in the mcp config. 

Any of these two would suffice as a first step.

Postremus:
### What was done and why?

Headers and authorization tokens of mcp servers can now be referenced from ENV variables.
Syntax is ${env:NAME_OF_MY_ENV}. The parser for this syntax can later easily be extended for other secret source types (e.g. vault).
This is guarded by a whitelist, in which env variables need to be named before they can be referenced.
Also a blacklist is present, preventing access to application secrets (like e.g. database password)
Secret references can be escaped, in case the reference need to be ever added in plain text: $${env:NAME_OF_MY_ENV} would result in plain text ${env:NAME_OF_MY_ENV}.

I deliberatly left out:
* showing any kind of validation message in the mcp ui when an unresolvable secret reference is used.
* or supporting dynamic reload of secret values. For example when an env var changes.

These parts can easily be done in follow up prs.

### AI Usage
I designed and implemented the `org.remus.giteabot.secret` package, and the first few tests for the SecretTemplateParser, by hand. (Since that is the fun part).
Opus integrated this work with the mcp config, let him write tests and the doc.
I manually tested the changes by hand to make sure the secret references are actually resolved.
I then did a few review rounds with opus, and let him split up the work into seperate commits.

closes #360


max-california (dismissed):
## 🤖 Agentic Code Review

Summary: the feature is well-scoped and the late resolution approach is sensible, but I found one behavioral regression in how unresolved references are handled. The new code can silently send placeholder strings like `${env:...}` as live credentials/headers, which is risky and differs from the “unresolvable secret reference” caveat in the PR description.

## BLOCKER

1. **Unresolved secret references are still sent over the wire as credential/header values**
   - **Files:** `src/main/java/org/remus/giteabot/secret/SecretTemplate.java`, `src/main/java/org/remus/giteabot/mcp/McpOrchestrationService.java`
   - **Why it matters:** `SecretTemplate.expose()` currently falls back to `sr.raw()` both when the source type is unknown and when the key cannot be resolved. `McpOrchestrationService.applyHeaders()` then unconditionally uses that returned value for request headers and authorization. This means a config like `"authorization_token": "${env:MISSING}"` becomes `Authorization: Bearer ${env:MISSING}` on outbound requests.
   - **Risk:** This is easy to misconfigure in production and hard to spot. It can cause confusing auth failures, leak internal placeholder syntax to third-party MCP servers, and makes “unresolvable secret reference” indistinguishable from intentional literal text.
   - **Suggested fix:** Treat unresolved references as resolution failures at request-build time instead of substituting the raw template. A few acceptable options:
     - make `SecretTemplate.expose()` throw `KeyResolveException` for unresolved `SecretReference` segments;
     - or add a stricter method used by MCP auth/header paths, e.g. `exposeOrThrow()`;
     - or have `applyHeaders()` detect whether a template still contains unresolved references and skip/fail the request with a clear diagnostic.
   - The current test `applyHeaders_sendsRawReferenceWhenSecretCannotBeResolved` codifies the problematic behavior and should be changed accordingly.

## LOW

1. **The parser only accepts lowercase secret source types, which may be stricter than needed**
   - **File:** `src/main/java/org/remus/giteabot/secret/SecretTemplateParser.java`
   - `(?<type>[a-z]+)` rejects `${ENV:KEY}` and any future source names containing digits/underscores. That may be intentional, but it’s a fairly narrow syntax contract for something described as extensible.
   - If this is deliberate, consider documenting the restriction clearly in `doc/SECRET_REFERENCES.md`; otherwise loosen the pattern now before users depend on it.

---

src/main/java/org/remus/giteabot/mcp/McpConfigurationParser.java
```diff
@@ -14,6 +16,12 @@ public class McpConfigurationParser {
 
     private final ObjectMapper objectMapper = new ObjectMapper();
 
+    private final SecretTemplateParser secretTemplateParser;
+
+    public McpConfigurationParser(SecretTemplateParser secretTemplateParser) {
```
tmseidel: Could you use Lombok for the constructor? - And add a Service annotation to get this wired automatically

src/main/java/org/remus/giteabot/mcp/McpOrchestrationService.java
```diff
@@ -47,12 +48,18 @@ public class McpOrchestrationService {
...
 
-    private final McpConfigurationParser configurationParser = new McpConfigurationParser();
-    private final McpServerDiscovery serverDiscovery = new McpServerDiscovery(configurationParser);
     private final ObjectMapper objectMapper = new ObjectMapper();
     private final Map<CacheKey, McpToolCatalog> toolCache = new ConcurrentHashMap<>();
+    private final McpConfigurationParser configurationParser;
+    private final McpServerDiscovery serverDiscovery;
     private final EncryptionService encryptionService;
 
+    public McpOrchestrationService(EncryptionService encryptionService, SecretTemplateParser secretTemplateParser) {
```
tmseidel: Could you use here Lombok as well, for that you also have to springify org.remus.giteabot.mcp.McpServerDiscovery and Lombok (I must have missed that in the past :( )

src/main/java/org/remus/giteabot/secret/SecretTemplate.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public String expose() {
+        return segments.stream().map(segment -> {
+            switch (segment) {
+                case Segment.SecretReference sr -> {
+                    Optional<SecretSource> optionalSecretSource = secretSourceRegistry.retrieve(sr.type());
+
+                    if (optionalSecretSource.isEmpty()) {
+                        log.error("Could not find secret source for {}", sr.raw());
+                        return sr.raw();
```
tmseidel, line 27: LGTM for the first step, with the error log entry debugging should be easy. 👍

tmseidel (changes requested):
Just the springification and the Lomboks  - The rest looks really great and well structured.

src/main/java/org/remus/giteabot/mcp/McpOrchestrationService.java
Postremus: done

src/main/java/org/remus/giteabot/mcp/McpConfigurationParser.java
Postremus: done, although I decided to use @Component since the MCPOrchestrationService is the entry point of this package and this parser class is just a helper.

tmseidel (approved):
Nice...

PR #378, opened 2026-09-06
feat(git): add manual Gitea SSH transport

CaeruleusAqua:
## What has been done?

- Added GitTransport (HTTP/SSH) with per-integration manual SSH configuration: encrypted private key, verified known_hosts, clear/keep semantics, and Gitea-only enforcement.
- Resolved repository-specific Gitea ssh_url values with SshEndpoint validation and known_hosts endpoint matching.
- Executed Git through isolated private-key and known_hosts files with strict host-key checking, no identity fallback, and fail-closed secret-file cleanup.
- Routed the branch-switcher fetch through WorkspaceService so SSH authentication is preserved.
- Added manual SSH configuration UI (transport selector, one-way key/known_hosts fields, encryption warnings) in all seven locales.
- Added H2 and PostgreSQL V47 migration and installed openssh-client in the Docker image; documented the manual operator setup.

## Why?

Gitea instances with [repository] DISABLE_HTTP_GIT = true cannot be served by the HTTP transport. This is the manual-transport foundation from the #368 split: operators register a dedicated key themselves while the bot keeps tokens out of clone URLs, pins host keys, and never stores private keys without APP_ENCRYPTION_KEY. Automatic key management follows in a separate PR.

## What has been tested?

- Focused tests: WorkspaceServiceTest, GiteaApiClientTest, SshEndpointTest, GiteaClientFactoryTest, GitIntegrationServiceTest, GitIntegrationControllerTest, GitIntegrationSshMigrationTest (V46 to V47), ToolExecutionServiceTest/CtagsTest in the Maven Java 21 container.
- Full mvn clean verify in the Maven Java 21 container: 1,622 tests passed.

## How AI was used?

AI helped split this scope out of superseded #368, port and reconcile it with the merged checkout-resolution PR, add redaction and key-replacement regression tests, and review the final diff for scope and standards.


max-california (changes requested):
## 🤖 Agentic Code Review

Summary: overall this is a thoughtful, security-conscious SSH transport addition. The separation between API credentials and Git transport, the fail-closed cleanup, and the Gitea-side `ssh_url` validation all look solid. I found one correctness issue around the UI/service interaction when an operator edits the Gitea endpoint while already using SSH.

## BLOCKER

1. **Editing a Gitea integration URL can silently force the transport back to HTTP and drop stored SSH credentials**
   - **Where:** `src/main/resources/templates/git-integrations/form.html` around the new `reconcileEndpointEdit()` / `switchToHttpAndClearSsh()` logic, and `src/main/java/org/remus/giteabot/admin/GitIntegrationService.java` lines 79–82.
   - **Why this is a bug:** On edit, if the original integration uses SSH and the operator changes the Gitea URL, the JS immediately does:
     - `transportInput.value = 'HTTP'`
     - `clearSshCredentials = true`
   - Then on submit, the service sees non-SSH transport and unconditionally clears `sshPrivateKey` / `sshKnownHosts`.
   - The problem is that `GitIntegrationService.validate(...)` only requires a new token when the endpoint changes; it does **not** require the operator to consciously confirm the transport downgrade / SSH credential deletion. So an admin who just updates the URL and token can unintentionally convert an SSH integration into HTTP and erase the stored SSH setup.
   - **Why this matters:** It’s a production-behavior change, not just UX: instances using `DISABLE_HTTP_GIT=true` will stop working after what appears to be a routine URL edit.
   - **Suggestion:** Don’t auto-switch transport to HTTP on endpoint edits. Instead either:
     - keep SSH selected and require the operator to provide replacement `known_hosts` (and private key only if they want to rotate it), or
     - block save with a clear validation error until the operator explicitly changes transport or explicitly clears/replaces SSH credentials.

## LOW

1. **`RepositoryApiClient` docs are now slightly misleading after adding SSH transport**
   - **Where:** `src/main/java/org/remus/giteabot/repository/RepositoryApiClient.java`, comment on `getCloneUrl()` / `getCredentials()`.
   - **Why:** The comments still describe `cloneUrl` as HTTP-specific and `getCredentials()` as basically base URL / clone URL / username / token, but `RepositoryCredentials` now also carries transport and SSH material.
   - **Suggestion:** Refresh those method Javadocs to match the new model so future maintainers don’t assume clone operations are always HTTP-based.

---

max-california (dismissed):
## 🤖 Agentic Code Review

Summary: the overall design looks solid — especially the fail-closed secret-file handling, SSH endpoint validation, and routing remote git operations through `WorkspaceService`. I found one correctness issue in the admin save flow that can unexpectedly clear SSH credentials on unrelated edits, plus one maintainability/testing gap worth addressing.

### BLOCKER
- **Editing an existing SSH integration can silently erase `known_hosts` when the user updates only the private key.**  
  In `src/main/java/org/remus/giteabot/admin/GitIntegrationController.java:84-90`, the controller only copies `sshKnownHosts` onto the bound entity when the request parameter is non-blank. For edit forms, the `<textarea name="sshKnownHosts">` still submits an empty string when left untouched, so `integration.getSshKnownHosts()` remains blank. Then in `GitIntegrationService.save(...)`, if a new private key is supplied, `hasNewPrivateKey` is true and `hasNewKnownHosts` is false, which currently triggers `IllegalArgumentException("SSH private key and known_hosts are required...")` (`src/main/java/org/remus/giteabot/admin/GitIntegrationService.java:153-155`).  
  More importantly, if this validation is ever relaxed later, the current controller/service split makes it easy to accidentally overwrite the stored `known_hosts` with blank input. The controller should preserve the “blank means keep existing value” invariant consistently for both SSH fields, or the service should explicitly merge one-way fields before validation. This is user-facing and blocks a common credential-rotation workflow.

### MEDIUM
- **No regression test covers the “replace private key while retaining stored known_hosts” edit path.**  
  The new service tests cover blank-both-keeps, replace-without-known_hosts-rejects, and endpoint-change-requires-known_hosts, but not the expected admin edit case where only the key is rotated while the endpoint trust stays unchanged. Given the controller/service interaction above, add an integration/controller test around `POST /git-integrations/save` for an existing SSH integration to assert that leaving `sshKnownHosts` blank preserves the stored value and allows a key-only rotation. Relevant files: `src/test/java/org/remus/giteabot/admin/GitIntegrationControllerTest.java`, `src/test/java/org/remus/giteabot/admin/GitIntegrationServiceTest.java`.

---

src/main/resources/db/migration/h2/V47__git_integration_ssh_transport.sql
```diff
@@ -0,0 +1,4 @@
+-- V47: Configure repository Git transport and SSH credentials per integration.
+ALTER TABLE git_integrations ADD COLUMN transport VARCHAR(16) NOT NULL DEFAULT 'HTTP';
```
tmseidel: Idempotency is required here

src/main/resources/db/migration/postgresql/V47__git_integration_ssh_transport.sql
```diff
@@ -0,0 +1,4 @@
+-- V47: Configure repository Git transport and SSH credentials per integration.
+ALTER TABLE git_integrations ADD COLUMN transport VARCHAR(16) NOT NULL DEFAULT 'HTTP';
```
tmseidel: Idempotency is required here

tmseidel (changes requested):
just the sql-files.

tmseidel (approved):
Nice feature overall... LGTM

PR #379, opened 2026-09-07
feat(usage): add server-rendered pagination controls

### LINKED ISSUE #366: feature: Extensions for the paginated lists
Right now the paginated lists in the usage log only have buttons to jump to the next or previous page:

Image

So if I want to look at the logs of a workflow spanning 50+ rounds, I have to click around a lot.

Would it be possible
* to add buttons to jump to the first / last page?
* and Add a dropdown (or similar) to choose the amount of elements on each page of the list

Maybe an input to jump to a specific page could be added - or show the next few page numbers. For example, what I had in mind:
Image

This also affects other paginated list then only the usage log. E.g. mcp selection, tool selection, etc. Only to a lesser effect on those lists, since fewer elements are ever present in them.

tmseidel:
# What has been done?

Added configurable, server-rendered pagination controls to the usage log page. Both the usage and errors tables now expose the same Bootstrap pagination UI used consistently across the application: first/last, previous/next, a sliding window of page numbers, a jump-to-page form, and a rows-per-page dropdown. The AiUsageService now accepts a requested page size and falls back to the configured default for invalid values.

The mentioned pagination of MCP tools or Tools selection in the system setting (#366) is out of scope since these lists work completely differently and need a separate pagination logic. Since these lists won't be so long by the business-cases we keep the current logic (pure client-side pagination)

This is a screenshot of the result:

image

# Why?

The usage page previously only had basic previous/next pagination and a fixed page size. This change gives users the same rich navigation experience already established elsewhere, lets them choose how many rows to display, and prevents navigation to out-of-range pages.

# What has been tested?

- AiUsageServiceTest — verified requested page size is honored and invalid sizes fall back to the default.
- UsageControllerTest — verified the rendered usage view contains the new pagination controls, including page-number links and the rows-per-page dropdown.
- Ran mvn test -Dtest=AiUsageServiceTest,UsageControllerTest -q successfully.
- manual UI Tests

# How AI was used?

An AI coding assistant was used to analyze the existing usage page, extend the service and controller for page size, rewrite the Thymeleaf pagination markup, add the new i18n keys across languages, and update the tests. All changes were reviewed against the `develop` branch and verified by the test suite. All changes and smaller bugs of the AI reviewed or corrected by me


max-california (changes requested):
## 🤖 Agentic Code Review

Summary: the pagination UI is a useful improvement and the core server-side plumbing looks consistent, but I found one correctness issue in the jump-to-page handling that will send users to the wrong page and can also throw on out-of-range input. Everything else I checked looked aligned with the existing code.

## BLOCKER

1. **Jump-to-page submits a 1-based page number, but the controller/service expect 0-based indices**
   - **Files:** `src/main/resources/templates/usage/list.html`, `src/main/java/org/remus/giteabot/aiusage/UsageController.java`
   - In both jump forms, the input is rendered as:
     - `name="usagePage" type="number" min="1"`
     - `name="errorPage" type="number" min="1"`
   - However, `UsageController` passes those request params directly to Spring Data after only `Math.max(0, ...)`, and `PageRequest.of(...)` is 0-based.
   - Result:
     - Entering page `1` takes the user to the **second** page, not the first.
     - Entering the visible last page number (e.g. `5`) attempts to load zero-based page `5`, i.e. the **sixth** page. If only five pages exist, repository behavior can degrade into an empty result, and the template then computes links like `usagePage.totalPages - 1`, which can become negative and eventually feed `PageRequest.of(-1, ...)` on the next request.
   - This conflicts with the PR description’s goal of preventing out-of-range navigation.
   - **Suggested fix:** either convert submitted values to zero-based in the controller (`requestedPage - 1`, clamped into `[0, totalPages-1]`), or rename/add separate parameters for 1-based UI values and normalize them before querying.

## MEDIUM

None.

## LOW

None.

---

max-california (changes requested):
## 🤖 Agentic Code Review

Summary: The pagination redesign looks largely correct, and I didn’t find any obvious correctness or security regressions in the 1-based/0-based page conversion or sort-column sanitization. The main concern is that the new `pageSize` handling is effectively unbounded server-side, which can allow very large requests despite the UI only offering small preset values.

## BLOCKER

- **Unbounded `pageSize` can trigger excessively large DB queries and page materialization**
  - **Files:** `src/main/java/org/remus/giteabot/aiusage/UsageController.java:42,46`, `src/main/java/org/remus/giteabot/aiusage/AiUsageService.java:123`
  - `usageSize` / `errorSize` come straight from request parameters and `effectivePageSize()` only rejects non-positive values. A user can request `/usage?usageSize=1000000` and the service will pass that directly into `PageRequest.of(...)`.
  - Because this is a server-rendered admin page, that means a single request can force a very large result page to be loaded from the database and rendered, which is a performance and availability risk.
  - Suggestion: clamp to an allowlist or max value in the service layer, e.g. only `{10,20,50,100}` or at least `min(requested, 100)` with fallback to `PAGE_SIZE`.

## MEDIUM

- None.

## LOW

- **No controller/view test covers hostile or oversized page-size inputs**
  - **Files:** `src/test/java/org/remus/giteabot/aiusage/AiUsageServiceTest.java`, `src/test/java/org/remus/giteabot/aiusage/UsageControllerTest.java`
  - The tests cover `0` falling back to default and valid requested sizes being honored, but they don’t cover the important boundary case introduced by this PR: very large positive values.
  - Once you clamp or allowlist page sizes, please add a test for something like `usageSize=99999` / `errorSize=99999` to lock the behavior in.

---

max-california (changes requested):
## 🤖 Agentic Code Review

Summary: the feature looks solid overall and the service-side page-size clamping is a good safety addition. I found one correctness issue in the new pagination flow: when a user changes the rows-per-page on a later page, the controller can request a page index that no longer exists, which produces an empty table instead of keeping the user on a valid page.

## BLOCKER

1. **Changing page size can navigate to an out-of-range page and show an empty result set**
   - **Where:** `src/main/resources/templates/usage/list.html:200`, `src/main/resources/templates/usage/list.html:222`, `src/main/resources/templates/usage/list.html:369`, `src/main/resources/templates/usage/list.html:391`, and `src/main/java/org/remus/giteabot/aiusage/UsageController.java:55-58`
   - **Why it matters:** The PR description explicitly says it “prevents navigation to out-of-range pages”, but the rows-per-page forms preserve the current page number (`usagePage.number + 1` / `errorPage.number + 1`) when the page size changes. If the user is on a high page number and switches to a larger page size, that page may no longer exist. The controller currently just forwards the resulting 0-based index to Spring Data, so the page query returns an empty page even though matching records still exist.
   - **Example:** 81 usage entries, user on page 5 at 20 rows/page, then switches to 100 rows/page. The form submits `usagePage=5`, controller converts to index 4, but only page 1 exists at size 100. The table renders empty.
   - **Suggested fix:** Reset to page 1 when page size changes, or clamp the requested page after loading total pages (e.g. re-query if requested page exceeds `totalPages - 1`). Resetting to page 1 in the size-change forms is the simplest and matches typical UX.

## LOW

1. **Missing regression test for the out-of-range page-size change case**
   - **Where:** `src/test/java/org/remus/giteabot/aiusage/UsageControllerTest.java`
   - **Why:** The new tests cover 1-based/0-based conversion and rendering, but not the key edge case above. A focused test around changing `usageSize`/`errorSize` while preserving an invalid page would make this behavior explicit and prevent regressions once fixed.

---

max-california (approved):
## 🤖 Agentic Code Review

Summary: the feature is generally well-implemented: service-side page-size clamping is a good safeguard, and the controller/template updates consistently preserve the other table’s state. I found one correctness issue in the new pagination UI that can render invalid page links near the beginning of the result set.

## MEDIUM

1. **Previous-page links generate page `0` in the UI on the second page**
   - **Where:** `src/main/resources/templates/usage/list.html` around the new previous-link hrefs for both tables (`usagePage=${usagePage.number}` and `errorPage=${errorPage.number}`)
   - **Why it matters:** after switching the request contract to 1-based page numbers, the “Previous” link should point to page `current - 1` in 1-based terms. On the second page, `usagePage.number` is `1` (0-based), so the generated query parameter becomes `usagePage=1`, which happens to be correct. But on the first page, the disabled link still renders `usagePage=0`, and likewise for `errorPage=0`. Even if the browser usually won’t follow a disabled Bootstrap link, the markup now contains out-of-range page parameters that contradict the stated “prevents navigation to out-of-range pages” goal and can be exposed to crawlers, accessibility tools, or users opening the link target directly.
   - **Suggested fix:** generate bounded previous/next targets in 1-based terms, e.g. `T(java.lang.Math).max(1, usagePage.number)` for previous and `T(java.lang.Math).min(usagePage.totalPages, usagePage.number + 2)` for next; same for the error table. Alternatively, avoid rendering an `<a>` at all for disabled items and render a `<span class="page-link">...`.

## LOW

1. **Controller exposes raw requested page sizes back to the view instead of the effective/clamped value**
   - **Where:** `src/main/java/org/remus/giteabot/aiusage/UsageController.java:64-65`
   - **Why it matters:** if a user requests `usageSize=999999`, the service correctly clamps to 100, but the model still contains `usageSize=999999`. That can leave the rows-per-page `<select>` with no option selected and makes the rendered state diverge from the actual query.
   - **Suggested fix:** either normalize sizes in the controller before adding them to the model, or expose a shared normalization helper from `AiUsageService` and reuse it there.

2. **Missing test coverage for invalid/clamped page-size rendering**
   - **Where:** `src/test/java/org/remus/giteabot/aiusage/UsageControllerTest.java`
   - **Why it matters:** there are service tests for clamping, but no controller/view test that verifies an invalid `usageSize`/`errorSize` request still renders the correct selected option/state in the HTML. That would have caught the model/view mismatch above.
   - **Suggested fix:** add a controller test with oversized or non-positive size params and assert the rendered select shows the effective size.

---

PR #382, opened 2026-09-09
feat(git): manage Gitea SSH keys with confirmed host trust

CaeruleusAqua:
## What has been done?

- Added SSH host-key preview and explicit fingerprint confirmation, with a fresh scan checked before setup.
- Generate dedicated Ed25519 keys locally, register only public keys with Gitea, and encrypt stored private keys.
- Track remote key ID, owner ID, and unique title, including recovery after ambiguous registration responses.
- Commit HTTP-only state and retain tracking before remote cleanup; clear tracking only after confirmed cleanup. Validate form constraints before revoking working keys and suppress raw upstream errors in logs/flash messages.
- Preserve the requested SSH transport: incomplete replacements fail before cleanup rather than silently saving HTTP. Explicit HTTP selection and complete manual SSH replacement remain supported.
- Added automatic setup UI, seven-locale messages, provider APIs, and operator documentation while preserving manual-key rotation and endpoint-edit fixes.
- Add one idempotent V50__git_integration_managed_ssh_keys.sql per H2/PostgreSQL dialect for all three tracking columns. V48/V49 remain available for the upstream model-flavor and branch-filter migrations.

## Why?

This is part 3 of the replacement for superseded #368, built on merged #378. Operators can configure managed Gitea SSH credentials after independently verifying host fingerprints, without manually generating and registering a key.

Database-backed lifecycle concurrency fencing, browser version checks, and deletion_pending orchestration remain reserved for part 4. This PR establishes sequential setup/cleanup behavior and committed recovery states.

## What has been tested?

- Full Maven Java 21 clean verify: 1,831 tests passed with zero failures/errors/skips, using temporary /workspace/data storage to isolate persistent preview database history.
- JavaScript form regression: node src/test/js/git-integration-form.cjs.
- Real Ed25519 generation, mocked Gitea API registration/cleanup, ownership and confirmation checks, and secret-redaction tests.
- Spring/H2 tests verify committed state in separate transactions before remote side effects and after injected failures, including rejected invalid/incomplete replacements, preserved transport choice, and cleanup retry.
- Native PostgreSQL 16.15 with Flyway 11.14.1: upstream c3a455a migrations through V49 followed by consolidated V50, tested across eight cases covering every existing-column combination. Pre-upgrade credentials and tracking were preserved, V50 SQL replayed twice per case, upstream history unchanged, and Flyway validation/repeated migration succeeded.
- H2 migration and idempotency regressions; independent implementation review and git diff --check.
- No live Gitea or browser end-to-end run; concurrent lifecycle operations are not fenced in this PR.

## How AI was used?

AI reconstructed the managed-key scope from the archived implementation without cherry-picking interleaved commits, preserved merged review fixes, added failure-state and migration regressions, and independently reviewed the implementation. Review findings were addressed, including requested-transport preservation and consolidation at V50 to avoid upstream migration-version collisions.


max-california (changes requested):
## 🤖 Agentic Code Review

Summary: the overall design looks careful, especially the staged persistence around remote side effects and the explicit host-key confirmation flow. I found one correctness issue in the controller’s managed-key cleanup path that can silently drop a user’s intended transport choice.

## BLOCKER

1. **Managed-key cleanup forces SSH selections back to HTTP, so operators cannot switch from a managed key to a manual SSH key in one save**
   - **Where:** `src/main/java/org/remus/giteabot/admin/GitIntegrationController.java:146-149`, with the follow-on save at `:164`
   - **Why it matters:** In the managed-key cleanup branch, the code treats `privateKeyChanged == false` as “disable SSH” and unconditionally does:
     ```java
     integration.setTransport(GitTransport.HTTP);
     integration.setSshPrivateKey(null);
     integration.setSshKnownHosts(null);
     ```
     That includes the case where the user submitted `transport=SSH`, checked “clear managed credentials”, and provided a new `known_hosts` but no new private key yet (or intends to preserve an already-submitted SSH selection while changing other fields). After cleanup, `gitIntegrationService.save(...)` persists HTTP, not the SSH choice the form submitted.
   - **Concrete impact:** A valid manual-SSH migration path from a managed key can be silently converted into HTTP without an explicit user choice. This is especially risky because the request still returns success.
   - **Suggested fix:** Only coerce to HTTP when the user explicitly leaves SSH transport or when you intentionally want cleanup-only behavior. If the submitted transport remains SSH, preserve it and let `validateSave`/`save` enforce whether manual SSH credentials are complete. In other words, separate “remove managed key tracking first” from “change requested transport to HTTP”.

## MEDIUM

None.

## LOW

None.

---

max-california (approved):
## 🤖 Agentic Code Review

## Summary

The managed-key lifecycle is thoughtfully staged, with strong recovery handling, host-key revalidation, private-key encryption, secret-safe errors, and good regression coverage. One lifecycle edge case can leave an orphaned remote key.

## Findings

### BLOCKER

- **Remote key can be orphaned if the cleanup-tracking transaction fails after deletion**  
  `src/main/java/org/remus/giteabot/admin/GiteaSshSetupService.java:56-57`

  During replacement, `deleteTrackedRemoteKeys(...)` first removes the old key from Gitea, then `finishManagedSshKeyRemoval(...)` clears local tracking. If that database operation fails, a retry enters cleanup with the old tracked ID and title. Since neither exists remotely, cleanup succeeds—but tracking is cleared and setup proceeds to create a new key.

  However, the first setup invocation may continue if the transaction commit fails only when the proxied `finishManagedSshKeyRemoval` method returns, and the exception correctly aborts setup. More importantly, similar controller flows explicitly call `finishManagedSshKeyRemoval` after remote deletion; if that call fails, the retained marker is recoverable.

  The concrete orphan case is later in this method: after `createSshKey(...)`, if `configureGeneratedSsh(...)` fails, rollback deletes the key and then calls `finishManagedSshKeyRemoval(...)`. If key deletion succeeds but clearing tracking fails, the catch treats rollback as failed and retains the marker, which is safe. I do not find a confirmed defect in these paths after accounting for transaction-proxy commit timing.

  **Withdrawn as a defect:** the staged behavior is consistent and retryable.

No actionable correctness, security, or maintainability issues were found in the reviewed implementation.

---

tmseidel (approved):
LGTM

PR #383, opened 2026-09-09
feat(ai): stream responses from Ollama and llama.cpp to fix read time…

### LINKED ISSUE #381: Stream Ollama and llama.cpp responses to eliminate whole-request read timeouts on local models
# Problem
Agentic PR reviews against local Ollama intermittently fail with:
```
org.springframework.web.client.ResourceAccessException: I/O error on POST request for
"http://*****:11434/api/chat": Read timed out
  at o.r.g.ai.ollama.OllamaClient.executeRequest(OllamaClient.java:274)
  at o.r.g.ai.ollama.OllamaClient.chatWithTools(OllamaClient.java:109)
  at o.r.g.agent.loop.AgentLoop.callAiWithRetry(AgentLoop.java:348)
Caused by: java.net.SocketTimeoutException: Read timed out
```
Root cause (confirmed by reproduction, see Test Plan): OllamaClient and LlamaCppClient both send stream: false The server then emits zero bytes — not even response headers — until the entire generation completes. The configured read timeout (spring.http.clients.read-timeout, default 120s, user-configured up to 600s via HTTP_CLIENT_READ_TIMEOUT) therefore acts as a whole-request budget covering model load + prompt eval + generation.

The reporter's own measurement shows how quickly that budget is exhausted on a slow host:

total_duration: 111.9s = load_duration: 91.4s + prompt_eval: 0.3s + eval: 20.3s   (16 prompt tokens, 491 eval tokens)

In the agent loop, later rounds carry the full conversation history (PR diff + tool results), so prompt eval alone can take minutes — and 91s of model load happens whenever the model was evicted between rounds. The result is exactly the reported symptom: intermittent failures "only when the AI needs more time to review a PR." Raising the timeout further is an unbounded guessing game (each retry in AgentLoop.callAiWithRetry also multiplies the wait), and a genuinely hung server is indistinguishable from a slow one until the timeout fires.

The timeout plumbing itself is verified working end-to-end (property → Boot 4 HttpClientSettings → Apache HC5 socket timeout); no config fix can address the architectural issue.

# Proposal
Switch both local-provider clients to streaming requests and reassemble the full response client-side from the chunks. Streaming keeps bytes flowing during model load, prompt eval, and generation, so the read timeout becomes a per-chunk stall detector: slow-but-alive models never trip it at any reasonable timeout value, while a truly wedged server still fails fast.

Both streaming protocols are line-oriented, so one shared transport helper covers both:
```mermaid
sequenceDiagram
    participant L as AgentLoop
    participant C as OllamaClient / LlamaCppClient
    participant T as StreamingTransport (new, ai/ pkg)
    participant S as Ollama / llama.cpp

    L->>C: chatWithTools / submitReviewPrompt
    C->>T: executeStreaming(request, chunkConsumer)
    T->>S: POST /api/chat (stream:true) / /completion (stream:true)
    S-->>T: chunk 1 (model load done / first tokens)
    Note over T,S: each chunk resets the socket read timeout
    S-->>T: chunk 2..n (content deltas)
    S-->>T: final chunk (done:true / stop:true + usage counters)
    T-->>C: (consumer merged chunks incrementally)
    C->>C: reassemble into existing OllamaResponse / LlamaCppResponse
    C-->>L: ChatTurn / String (unchanged contract)
```

# Technical Implementation

1. Shared streaming transport (new class, org.remus.giteabot.ai)

StreamingLineReader (or similar) wrapping the Apache HC5-backed RestClient exchange:

- POST the request body, read the response as a raw InputStream (RestClient .exchange(...) with access to ClientHttpResponse.getBody() — NOT .body(Clazz)).
- Read line-by-line with a bounded line buffer; hand each non-blank line to a Consumer<String> callback.
- Both target protocols deliver one complete JSON object per unit:
  - Ollama: NDJSON — one OllamaResponse-shaped JSON object per line.
  - llama.cpp /completion with stream:true: SSE — data: {json}\n\n per chunk; strip the data:  prefix (same SSE-framing mode the helper can later reuse for OpenAI).
- Errors mid-stream (connection drop, malformed line) must surface as ResourceAccessException so AgentLoop.callAiWithRetry treats them like today's transient I/O failures.

2. OllamaClient changes

- Set .stream(true) in chatWithTools (OllamaClient.java:101) and doRequest (:250).
- executeRequest → stream variant: parse each line as OllamaResponse; accumulate message.content (concatenate); take toolCalls, doneReason, promptEvalCount, evalCount from the done:true chunk (usage counters only appear there — audit/usage reporting via reportUsage must use the final chunk, same totals as today).
- Reassembled OllamaResponse feeds the unchanged interpret(...) / extractText(...) paths — no changes to AgentLoop, AuditingAiClient, or callers.

3. LlamaCppClient changes

- Set .stream(true) in doRequest (LlamaCppClient.java:137).
- Merge chunk content fields; take tokensEvaluated/tokensPredicted, stoppedLimit, timings from the final chunk (stop:true).
- Preserve existing behavior: GBNF grammar (AGENT_JSON_GRAMMAR), cachePrompt(true), stop sequences, and the truncated-response warning at :188 must work identically on the merged result. (llama.cpp applies grammar constraints identically in streaming mode; chunk text is unaffected.)

4. Out of scope (explicitly)

- Anthropic, OpenAI, Google clients: hosted APIs respond in seconds; their read-timeout budget is not a reported problem. Anthropic streaming additionally requires SSE event reassembly (message_start/content_block_delta/message_delta) with usage split across events and prompt-caching interaction — separate follow-up issue if ever needed.
- No config surface changes: spring.http.clients.read-timeout keeps its meaning, now per-chunk instead of per-request.
- No UI-visible streaming (responses are still consumed whole by the agent loop).

# Acceptance Criteria

1. OllamaClient sends "stream": true on every /api/chat call (review, chat, chatWithTools); the merged response produces byte-identical text, tool calls, stop reason, and token counts compared to a non-streamed response for the same generation.
2. LlamaCppClient sends "stream": true on every /completion call; merged content, GBNF grammar output validity, usage counters, and the stoppedLimit truncation warning behave as before.
3. A request whose server-side processing exceeds spring.http.clients.read-timeout succeeds as long as chunks keep arriving (integration test: server delays 10s before first chunk + 1s between chunks, timeout 2s → request completes).
4. A request where the server goes completely silent mid-stream still fails with ResourceAccessException/SocketTimeoutException within ~the read timeout (existing OllamaReadTimeoutIT-style test, adapted for streaming; must keep passing for the pre-first-chunk stall case).
5. Usage audit records (ai_usage_log) show identical input_tokens/output_tokens totals as non-streamed calls — no double-counting from summing per-chunk usage.
6. mvn verify green, including ArchitectureTest (new helper in ai/ package — no layering violations; no new dependencies).

# Edge Cases & Handling

Case: Server silent before first chunk (model loading > timeout)
Handling: Read timeout still applies — correct; Ollama emits no chunks during load. Mitigated in practice because load happens once; document
  that load time still counts
────────────────────────────────────────
Case: Connection dropped mid-stream
Handling: ResourceAccessException → existing retry in AgentLoop.callAiWithRetry
────────────────────────────────────────
Case: Malformed NDJSON/SSE line
Handling: Fail the request as I/O error (do not silently skip lines)
────────────────────────────────────────
Case: Empty stream (0 chunks)
Handling: Same as today's empty-response path: log warn + "Unable to generate..." fallback
────────────────────────────────────────
Case: done:true chunk missing usage counters
Handling: Skip reportUsage exactly as today's null-check does
────────────────────────────────────────
Case: Tool calls split across chunks
Handling: Take the final chunk's tool_calls (Ollama emits them complete); assert in test
────────────────────────────────────────
Case: Old Ollama/llama.cpp server without streaming support
Handling: Both endpoints have supported stream:true for years; no version gate planned — note in docs

# Test Plan

New/extended tests using the JDK com.sun.net.httpserver.HttpServer (project convention — no WireMock):

1. Ollama NDJSON merge test: stub server emits the reporter's real chunk sequence (see issue body), assert merged OllamaResponse has full content "Orange wavelengths dominate."-style text, evalCount=491, promptEvalCount=16.
2. Slow-but-alive test (AC 3): 10s initial delay + per-chunk delay, 2s read timeout → succeeds.
3. Silent-mid-stream test (AC 4): send 2 chunks then never close → times out ≈ read timeout.
4. Pre-first-chunk stall: existing OllamaReadTimeoutIT behavior preserved.
5. llama.cpp SSE merge test: data: {...} chunks incl. final stop:true chunk with tokens_evaluated/tokens_predicted; assert grammar-constrained JSON content merges intact.
6. Tool-call streaming test: chunk sequence ending in tool_calls → ChatTurn with StopReason.TOOL_USE, synthesized call ids unchanged.
7. Usage parity test: recorded audit usage equals final-chunk counters, not a sum.

Reference chunk stream (from the incident report) for test 1:

```json
{"model":"qwen3.8:latest","created_at":"...","message":{"role":"assistant","content":"ange"},"done":false}
{"model":"qwen3.8:latest","created_at":"...","message":{"role":"assistant","content":" wavelengths"},"done":false}
...
{"model":"qwen3.8:latest","created_at":"...","message":{"role":"assistant","content":""},"done":true,"done_reason":"stop","total_duration":111949917810,"load_duration":91368450406,"prompt_eval_count":16,"prompt_eval_duration":281125000,"eval_count":491,"eval_duration":20297458000}

```

tmseidel:
# What has been done?

Bug fix: local AI providers (Ollama, llama.cpp) now stream responses instead of requesting non-streamed ones, fixing intermittent ResourceAccessException: Read timed out failures on long PR reviews.

- New org.remus.giteabot.ai.StreamingLineReader — shared helper that POSTs via RestClient.exchange(...) and delivers the response line-by-line to a consumer. 4xx/5xx still produce the exact HttpClientErrorException via createException(), so prompt-too-long handling and retries are unchanged. No new dependencies.
- OllamaClient — sends stream: true on all /api/chat calls (review, chat, chatWithTools) and merges the NDJSON chunks: content concatenated, done_reason/prompt_eval_count/eval_count/tool_calls taken from the single done:true chunk.
- LlamaCppClient — sends stream: true on all /completion calls and merges the SSE (data: {json}) chunks, stripping framing and tolerating data: [DONE]; counters and stopped_limit from the stop:true chunk. GBNF grammar, cache_prompt, stop sequences and the truncation warning are untouched.
- 0-chunk streams still hit the existing "empty response" fallback; malformed lines fail the request as an I/O error (retried by AgentLoop.callAiWithRetry).
- systemtest/docker-compose-ollama.yml — OLLAMA_KEEP_ALIVE=-1 (keep model resident) and default pull model changed llama3.2:1b → qwen2.5-coder:7b.

# Why?

A non-streamed request makes the server emit zero bytes until the entire generation completes (model load + prompt eval + tokens — e.g. 111 s in the incident report). The configured spring.http.clients.read-timeout therefore acted as a whole-request budget, so slow local models failed intermittently exactly "when the AI needs more time to review a PR" — and raising the timeout is an unbounded guessing game that also multiplies retry waits, with a hung server indistinguishable from a slow one.

With streaming, bytes flow continuously (each chunk resets the socket read), so the read timeout becomes a per-chunk stall detector: slow-but-alive models never trip it, while a truly wedged server still fails fast. Model-load time still counts (no chunks are emitted during load) — a documented, accepted limitation.

# What has been tested?

New tests (all drive the real clients against a JDK HttpServer stub with a real 2 s Apache-HC5 read timeout — no WireMock, no Spring context, matching the project's existing OllamaReadTimeoutIT convention):

Manual system-tests were executed with an Agent-based review worfklow including the decision making. Hardware: Laptop with 32GB RAM and 4GB dedicated VRAM
1. Ollama with `qwen2.5-coder:7b` and the environemt-variable `OLLAMA_KEEP_ALIVE=-1` — this prevents Ollama from evicting a loaded model.
2. llama.cpp with qwen2.5-coder-7b-instruct-q4_k_m.gguf

# How AI was used?

Initial implemenation by an AI coding agent (Hermes Agent, Nous Research, running a local Qwen model) acting as a human-in-the-loop pair-programming session — the user directed scope (the issue) and decisions; the agent did the rest: human fixed code-smells and provided the `/spring-developer`skill

### DIFF
--- a/src/main/java/org/remus/giteabot/agent/loop/AgentLoop.java
+++ b/src/main/java/org/remus/giteabot/agent/loop/AgentLoop.java
@@ -9,8 +9,10 @@ import org.remus.giteabot.ai.AiMessage;
 import org.remus.giteabot.ai.ChatTurn;
 import org.remus.giteabot.ai.ToolDescriptor;
 import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.ResourceAccessException;
 
 import java.net.SocketException;
+import java.net.SocketTimeoutException;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
@@ -332,8 +334,12 @@ public final class AgentLoop {
      * When the provider rejects the request because the prompt exceeds its
      * context window, this method aggressively compacts the in-memory history
      * (keeping only the last 2 compaction units) and retries once.
-     * Transient socket write failures, such as a broken pipe or connection
-     * reset, are also retried once without modifying the conversation.
+     * Transient network failures are also retried once without modifying the
+     * conversation: socket write failures (broken pipe, connection reset,
+     * connection aborted) and streaming transport failures wrapped in
+     * {@link ResourceAccessException} (read timeout, mid-stream connection
+     * drop, malformed stream line — see {@link
+     * org.remus.giteabot.ai.StreamingLineReader}).
      *
      * <p>All other failures, and any failure on the second attempt, are
      * re-thrown to the caller.</p>
@@ -358,9 +364,9 @@ public final class AgentLoop {
             } catch (RuntimeException e) {
                 boolean promptTooLong = e instanceof HttpClientErrorException clientError
                         && aiClient.isPromptTooLongError(clientError);
-                boolean transientNetworkWriteFailure = isTransientNetworkWriteFailure(e);
+                boolean transientNetworkFailure = isTransientNetworkFailure(e);
 
-                if (attempt == maxAttempts || (!promptTooLong && !transientNetworkWriteFailure)) {
+                if (attempt == maxAttempts || (!promptTooLong && !transientNetworkFailure)) {
                     throw e;
                 }
 
@@ -371,7 +377,7 @@ public final class AgentLoop {
                     // Aggressive compaction: keep only the last 2 compaction units
                     compactor.compactAggressively(history);
                 } else {
-                    log.warn("AgentLoop: AI call failed with a transient network write error on attempt {}/{}. "
+                    log.warn("AgentLoop: AI call failed with a transient network error on attempt {}/{}. "
                             + "Retrying without changing history. Error: {}",
                             attempt, maxAttempts, e.getMessage());
                 }
@@ -381,18 +387,32 @@ public final class AgentLoop {
         throw new IllegalStateException("callAiWithRetry: exceeded max retries");
     }
 
-    static boolean isTransientNetworkWriteFailure(Throwable error) {
+    static boolean isTransientNetworkFailure(Throwable error) {
         Throwable current = error;
         while (current != null) {
-            if (current instanceof SocketException) {
-                String message = current.getMessage();
-                if (message != null) {
-                    String normalized = message.toLowerCase(Locale.ROOT);
-                    if (normalized.contains("broken pipe")
-                            || normalized.contains("connection reset")
-                            || normalized.contains("connection aborted")) {
-                        return true;
+            // Streaming transport failures surface as ResourceAccessException
+            // (read timeout / mid-stream drop / malformed line — see
+            // StreamingLineReader); a retry is the intended recovery, so treat
+            // the whole family as transient rather than matching the message.
+            switch (current) {
+                case ResourceAccessException resourceAccessException -> {
+                    return true;
+                }
+                case SocketTimeoutException socketTimeoutException -> {
+                    return true;
+                }
+                case SocketException socketException -> {
+                    String message = current.getMessage();
+                    if (message != null) {
+                        String normalized = message.toLowerCase(Locale.ROOT);
+                        if (normalized.contains("broken pipe")
+                                || normalized.contains("connection reset")
+                                || normalized.contains("connection aborted")) {
+                            return true;
+                        }
                     }
+                }
+                default -> {
                 }
             }
 
--- a/src/main/java/org/remus/giteabot/ai/AiClient.java
+++ b/src/main/java/org/remus/giteabot/ai/AiClient.java
@@ -98,6 +98,9 @@ public interface AiClient {
      */
     default boolean isPromptTooLongError(HttpClientErrorException e) {
         String body = e.getResponseBodyAsString();
+        if (body == null) {
+            return false;
+        }
         String normalized = body.toLowerCase(Locale.ROOT);
         String status = String.valueOf(e.getStatusCode().value());
         return normalized.contains("prompt is too long")
--- a/src/main/java/org/remus/giteabot/ai/StreamingLineReader.java
+++ b/src/main/java/org/remus/giteabot/ai/StreamingLineReader.java
@@ -1,0 +1,109 @@
+package org.remus.giteabot.ai;
+
+import org.springframework.http.HttpStatusCode;
+import org.springframework.web.client.ResourceAccessException;
+import org.springframework.web.client.RestClient;
+
+import java.io.BufferedReader;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.function.Consumer;
+
+/**
+ * Shared line-oriented streaming transport for local-provider AI clients
+ * (Ollama, llama.cpp).
+ *
+ * <p>Posts a request body and reads the streaming response line by line,
+ * handing every non-blank line to a consumer. Reading line by line keeps
+ * {@code spring.http.clients.read-timeout} acting as a <em>per-chunk stall
+ * detector</em> rather than a whole-request budget: each socket read is bounded
+ * by the timeout, so a slow-but-alive model (chunks keep arriving within the
+ * window) never trips it, while a genuinely wedged server (no byte for the whole
+ * window) still fails fast. This fixes the intermittent "Read timed out"
+ * failures on local Ollama / llama.cpp, where a non-streamed request blocks the
+ * socket until the entire generation — model load + prompt eval + token
+ * generation — completes.</p>
+ *
+ * <p>The helper is format-agnostic: it only strips line terminators and skips
+ * blank lines. The consumer is responsible for any framing it needs — Ollama
+ * emits NDJSON (one JSON object per line) and llama.cpp's {@code /completion}
+ * with {@code stream:true} emits SSE ({@code data: {json}} per chunk) — and for
+ * parsing / reassembling the chunks into the provider-specific response DTO.</p>
+ *
+ * <p>Notes:</p>
+ * <ul>
+ *   <li>Model-load time still counts against the read timeout: Ollama emits no
+ *       chunks while loading, so a load longer than the timeout still trips it.
+ *       That is expected and documented; streaming removes the timeout pressure
+ *       for the (much larger) prompt-eval + generation phases.</li>
+ *   <li>I/O failures mid-stream (connection drop, read timeout) surface as
+ *       {@link ResourceAccessException}, which {@code AgentLoop.callAiWithRetry}
+ *       already treats as transient.</li>
+ *   <li>Malformed lines are surfaced to the consumer; the client fails the
+ *       request (does not silently skip) when a line does not parse.</li>
+ *   <li>No new dependencies: only the auto-configured {@link RestClient} and JDK.</li>
+ * </ul>
+ */
+public final class StreamingLineReader {
+
+    private StreamingLineReader() {
+        // utility class
+    }
+
+    /**
+     * POSTs {@code body} to {@code uri} (relative to the client's base URL) and
+     * streams the response, invoking {@code lineConsumer} for each non-blank
+     * line of the body.
+     *
+     * <p>The response is closed by the underlying {@link RestClient}
+     * {@code exchange} call after this method returns (or after an exception
+     * propagates), so callers do not need to close anything. The stream is
+     * drained fully (to EOF) before returning, which is also what keeps the
+     * per-read socket timeout active for the entire generation.</p>
+     *
+     * @param restClient   the configured client (base URL, headers, read timeout)
+     * @param uri          the endpoint, relative to the client's base URL
+     * @param body         the request payload
+     * @param lineConsumer receives each non-blank response line (may throw to
+     *                     fail the request, e.g. on a malformed line)
+     * @throws ResourceAccessException on I/O errors, including a read timeout
+     *         before or mid-stream
+     * @throws org.springframework.web.client.HttpClientErrorException on 4xx
+     * @throws org.springframework.web.client.HttpServerErrorException on 5xx
+     */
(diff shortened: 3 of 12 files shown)


max-california:
⚠️ **AI Agent (Agentic Review)**: The review could not be completed because of an error. This is usually a transient issue with the AI provider. Please try again later.

Error details:
```
400 Bad Request: "{<EOL>  "error": {<EOL>    "message": "Function tools with reasoning_effort are not supported for gpt-5.6-sol in /v1/chat/completions. To use function tools, use /v1/responses or set reasoning_effort to 'none'.",<EOL>    "type": "invalid_request_error",<EOL>    "param": "reasoning_effort",<EOL>    "code": null<EOL>  }<EOL>}"
```

_Check the bot logs for the full stack trace._

max-california (changes requested):
## 🤖 Agentic Code Review

Summary: the streaming transport and provider-specific reassembly look sound overall, and the new tests cover the happy path plus the key timeout behavior well. I found one correctness regression in Ollama metadata preservation and one maintainability/robustness issue around null handling in prompt-too-long detection.

## BLOCKER

1. **Ollama merged responses drop model metadata for non-`done:true` streams**
   - **Where:** `src/main/java/org/remus/giteabot/ai/ollama/OllamaClient.java`, around `executeRequest()` lines 319–327
   - **Why it matters:** The old non-streaming path deserialized a single `OllamaResponse`, so `model` was available whenever the server sent it. In the new streaming path, `merged.setModel(...)` only happens when a `done:true` chunk is seen. If the stream ends without a final done chunk but still contains usable content (e.g. provider/proxy truncation, partial stream before disconnect, or any server quirk), `executeRequest()` now returns a non-null merged response with content but **without model**, even if earlier chunks contained it.
   - **Impact:** This changes raw/audit payload shape and can silently lose provider metadata in degraded-but-successful cases. The llama.cpp path already avoids this by merging from `lastRef` as fallback.
   - **Suggestion:** Mirror the llama.cpp approach by tracking the last chunk and using it as the source for metadata when no final chunk exists.

## MEDIUM

1. **`isPromptTooLongError()` now assumes `getResponseBodyAsString()` is never null**
   - **Where:**  
     - `src/main/java/org/remus/giteabot/ai/ollama/OllamaClient.java` lines 118–121  
     - `src/main/java/org/remus/giteabot/ai/llamacpp/LlamaCppClient.java` lines 87–94
   - **Why it matters:** `LlamaCppClient` explicitly removed a null guard in this PR, and `OllamaClient` also lowercases the body unconditionally. If Spring returns an exception with no response body, `body.toLowerCase(...)` will throw `NullPointerException` during retry classification, masking the original HTTP error.
   - **Suggestion:** Restore the previous defensive pattern:
     ```java
     if (body == null) {
         return false;
     }
     ```
   - This is small, but it avoids turning an ordinary provider error into an internal classification failure.

## LOW

1. **Missing regression test for Ollama partial-stream metadata fallback**
   - **Where:** `src/test/java/org/remus/giteabot/ai/ollama/OllamaClientStreamingTest.java`
   - **Suggestion:** Add a case where content arrives but no `done:true` chunk is emitted, and assert whatever fallback behavior you intend for metadata/counters. That would lock in the degraded-stream semantics and catch the metadata-loss issue above.

---

max-california (changes requested):
## 🤖 Agentic Code Review

## Summary

The streaming approach looks sound overall and the tests cover the key timeout/merge behaviors well. I found one correctness issue in the retry path: malformed streamed chunks are now wrapped as `ResourceAccessException`, but the existing retry logic does not actually treat `ResourceAccessException` as transient, despite the PR description/comments saying it will be retried.

## Findings

### BLOCKER

1. **Malformed/timeout stream failures are not retried by `AgentLoop.callAiWithRetry`**
   - **Where:** `src/main/java/org/remus/giteabot/ai/StreamingLineReader.java`, `src/main/java/org/remus/giteabot/ai/ollama/OllamaClient.java`, `src/main/java/org/remus/giteabot/ai/llamacpp/LlamaCppClient.java`, `src/main/java/org/remus/giteabot/agent/loop/AgentLoop.java:358-376`
   - **Why it matters:** Both streaming clients convert malformed chunk parse failures into `ResourceAccessException`, and real read timeouts during streaming will also surface as `ResourceAccessException`. However, `callAiWithRetry()` only retries prompt-too-long errors and `SocketException`-based "broken pipe/connection reset/connection aborted" write failures. A `ResourceAccessException` from a read timeout or malformed stream will fail immediately on the first attempt.
   - **Impact:** This contradicts both the PR description ("malformed lines fail the request as an I/O error (retried by AgentLoop.callAiWithRetry)") and the inline comments in the new code. More importantly, it weakens resilience for exactly the streaming/network faults this PR is touching.
   - **Suggestion:** Extend retry detection in `AgentLoop` to treat relevant `ResourceAccessException`s as transient (at least read timeouts / socket timeouts, and possibly the explicitly wrapped malformed-stream errors if that is the intended behavior), or stop wrapping malformed parse errors as `ResourceAccessException` if they are meant to be terminal.

### LOW

1. **`OllamaReadTimeoutIT` response body is malformed JSON**
   - **Where:** `src/test/java/org/remus/giteabot/admin/OllamaReadTimeoutIT.java:61-66`
   - **Why:** The stub body is `{"message":{"role":"assistant","content":"ok"},` with no closing braces. The test currently only asserts timeout behavior, so it probably still passes, but the fixture is misleading and could confuse future maintenance/debugging.
   - **Suggestion:** Return a complete JSON document, even if the test does not consume it.

---

max-california (approved):
## 🤖 Agentic Code Review

Summary: the streaming approach looks sound and the coverage is strong. I didn’t find a correctness issue in the Ollama/llama.cpp response reassembly itself. My main concern is that the retry classifier in `AgentLoop` was broadened enough to also retry deterministic local parsing bugs, which can hide real defects and duplicate side effects without improving recovery.

## MEDIUM

### `AgentLoop` now retries malformed streamed payloads as if they were transient transport errors
In `src/main/java/org/remus/giteabot/agent/loop/AgentLoop.java:393` and in both streaming clients (`src/main/java/org/remus/giteabot/ai/ollama/OllamaClient.java:297`, `src/main/java/org/remus/giteabot/ai/llamacpp/LlamaCppClient.java:196`), malformed provider lines are wrapped in `ResourceAccessException`, and `isTransientNetworkFailure(...)` now treats **all** `ResourceAccessException`s as retryable.

That means a deterministic parse failure caused by:
- a provider format change,
- an incompatible proxy injecting framing,
- or a bug in our SSE/NDJSON handling

will now always get retried once as though it were a socket timeout/reset. Retrying won’t fix those cases, and for tool-using turns it can duplicate external side effects if the provider/tool invocation had already progressed before the malformed chunk arrived.

I’d suggest narrowing the retry classification so only genuine I/O transport failures are retried (e.g. `SocketTimeoutException`, `SocketException`, or a `ResourceAccessException` whose cause chain is I/O-related), while malformed-stream parsing should surface immediately as a non-retryable client bug/protocol mismatch. An alternative is to use a distinct exception type for “bad streamed payload” and exclude it from `isTransientNetworkFailure()`.

---

PR #384, opened 2026-09-09
feat(ai): add per-integration model flavor presets

tmseidel:
# What has been done?

New feature: a per-integration Model Flavor setting for AI Integrations (admin UI + provider layer + DB).

- Generic provider extension point: ModelFlavor record (provider type, id, label, description) + AiProviderMetadata.getFlavors() (default: none) + registry accessor, so flavor ids stay unique across providers.
- OpenAI integration offers two flavors: standard (default, no extra request fields) and no_reasoning (sends reasoning_effort: "none" in the chat-completions request; field omitted otherwise via @JsonInclude(NON_NULL)).
- New ai_integrations.model_flavor column (Flyway V48, PostgreSQL + H2, idempotent, default standard).
- Admin form: Model flavor field mirroring the Model field's layout (text input + "Select" dropdown, full column width, hint listing all available flavors); only shown when the provider offers flavors. i18n keys in all 7 locales.
- Unknown or blank flavor ids degrade to standard with a warn log — a stale value can never break a workflow.
- Docs updated: doc/USER_GUIDE.md (form field list, OpenAI provider notes, OpenAI-compatible field table, new troubleshooting bullet for the exact 400 error) and doc/AGENT.md (new "Model flavors" section).

# Why?

Switching an integration from gpt-5.4 to gpt-5.6-sol broke the PR workflow: the gateway in front of the model injects a default reasoning_effort, and newer reasoning models refuse native function tools on /v1/chat/completions unless reasoning_effort is explicitly none — the provider answers with 400 "Function tools with reasoning_effort are not supported…". Rather than hardcoding an error-string retry in OpenAiClient, this adds a small, operator-visible knob: the admin can flip the affected integration to "No reasoning effort" and the workflow works again. The flavor mechanism is generic, so future per-model request tweaks don't need new columns or special-casing.

# What has been tested?

- OpenAiClientTest (+103 lines): request payload assertions for both flavors (field present with "none" / absent for standard), flavor resolution, and unknown/blank fallback to standard.
- AiProviderRegistryTest (+19 lines): the flavor registry surface (OpenAI exposes two provider-stamped flavors, other providers none).
- AiIntegrationControllerTest (4/4): renders /new and /edit through Thymeleaf, so the new form markup parses and binds cleanly.
- Full suite: 1692 tests, 0 failures, 0 errors (re-run after the final template/i18n changes).
- Live manual test of a real PR review against the gpt-5.6-sol gateway  the 400 was gone 

# How AI was used?

All of the work above was implemented with an AI coding agent (Hermes Agent, local LLM — Qwen3.8-27B): code, tests, form markup/JS, i18n, DB migration, and documentation. The user provided the production 400 log and steered the design (generic flavor field instead of a hardcoded retry; model-field layout for the dropdown). All claims in this description are backed by the test runs and diffs above


max-california (changes requested):
## 🤖 Agentic Code Review

Summary: the provider-side flavor plumbing and OpenAI request changes look coherent, and the fallback-to-standard behavior is a sensible safety net. I found one correctness issue in the controller tests/mvc wiring and one smaller maintainability issue in the new request-field docs.

## BLOCKER

1. **`AiIntegrationControllerTest` does not stub the new `providerRegistry.getFlavors()` call, so `/new` and `/edit` rendering can now fail with a null dereference in the template script**
   - **Where:** `src/main/java/org/remus/giteabot/admin/AiIntegrationController.java:62`, `src/main/resources/templates/ai-integrations/form.html:189-246`, `src/test/java/org/remus/giteabot/admin/AiIntegrationControllerTest.java:57-72`, `124-143`
   - **Why:** the controller now always adds `providerFlavors`, but the WebMvc test only stubs `getProviderTypes/getDisplayNames/getDefaultApiUrls/getSuggestedModels/getApiKeyRequirements`. For a Mockito bean, unstubbed `getFlavors()` returns `null`, which means the inline JS emits `const providerFlavors = null;` and `providerFlavors[provider]` will throw at runtime when `updateProviderDefaults()` runs.
   - **Impact:** the admin new/edit page can break in tests, and more importantly this is brittle enough that any similar mock-based rendering path will miss a required model attribute.
   - **Suggestion:** stub `when(providerRegistry.getFlavors()).thenReturn(Map.of(...))` in the existing controller tests, ideally returning empty lists for providers without flavors.

## MEDIUM

1. **`OpenAiRequest`’s new field comment references a “targeted retry” that no longer exists**
   - **Where:** `src/main/java/org/remus/giteabot/ai/openai/OpenAiRequest.java:24-30`
   - **Why:** the comment says “See the targeted retry in `OpenAiClient#executeRequest(OpenAiRequest)`”, but `executeRequest()` is now a straight POST with no retry logic.
   - **Impact:** misleading comments around provider error handling make future maintenance harder, especially since the PR description explicitly says this feature replaces a hardcoded retry approach.
   - **Suggestion:** update the Javadoc to describe the flavor-based behavior directly and remove the retry reference.

---

max-california (approved):
## 🤖 Agentic Code Review

Summary: the feature looks coherent end-to-end and the OpenAI flavor wiring appears correct. I didn’t find any correctness or security regressions in the provider/client path or migration. I only have one maintainability/UI issue around stale form state when switching providers.

## MEDIUM

1. **Hidden `modelFlavor` value is still submitted after switching to a provider that does not support flavors**
   - **File:** `src/main/resources/templates/ai-integrations/form.html`
   - **Lines:** around 251–275
   - When the selected provider has no flavors, the script hides the field/button, but it does not clear `modelFlavorInput.value`. That means an operator can:
     1. select OpenAI,
     2. pick `no_reasoning`,
     3. switch to another provider with no flavors,
     4. save,
     5. persist `model_flavor=no_reasoning` on a non-OpenAI integration.
   - This does not break runtime today because other providers ignore the column, but it leaves stale provider-specific state in the DB and creates surprising behavior if the integration is later switched back or if another provider introduces a conflicting flavor id.
   - Suggestion: in the `else` branch, explicitly reset the input to `"standard"` (or blank if you want the server-side defaulting path), e.g.:
     ```js
     modelFlavorInput.value = 'standard';
     modelFlavorHint.textContent = '';
     ```
     This keeps persisted state aligned with the currently selected provider.

## LOW

1. **The “globally unique” flavor-id comments are misleading**
   - **Files:** `src/main/java/org/remus/giteabot/ai/openai/OpenAiProviderMetadata.java`, `src/main/java/org/remus/giteabot/ai/openai/OpenAiFlavor.java`, `src/main/java/org/remus/giteabot/ai/ModelFlavor.java`
   - Several comments say the flavor id is “globally unique”, but the persisted value is still just the bare `id` (`standard`, `no_reasoning`), and the UI also writes only `flavor.id` back into the text field.
   - In practice uniqueness comes from the `(providerType, id)` pair in metadata, not from the stored id itself.
   - Recommend rewording those comments to avoid implying that the DB column stores namespaced ids.

---

PR #385, opened 2026-09-10
feat(bot): Branch filter

### LINKED ISSUE #375: Branch filter on PR-Workflow operations
Originates from #374

## Quality assessment

The issue intent is now clear enough to implement: add a persisted branch/ref allowlist to Bot configuration and enforce it before starting PR workflows from incoming webhook events. The original text had wording issues and some ambiguity around exact matching semantics, but the expected behavior is sufficiently defined to produce an actionable draft with a few non-critical open questions.

## Summary
Add a configurable branch/ref allowlist for PR-Workflow execution on the bot configuration page. When a webhook event would normally trigger a PR workflow, the workflow must only start if the PR branch/ref matches the bot's configured `branchFilter`.

## Problem
Currently, an assigned PR workflow starts whenever the triggering conditions are met (for example, bot assignment or `Run workflow when PR is opened`). This does not allow teams using branch-based workflows (for example, git-flow) to restrict PR workflow execution to specific target branches or refs.

## Proposed change
Add a new `branchFilter` property to `org.remus.giteabot.admin.Bot` and expose it on the bot edit page at `bots/[id]/edit`.

This filter must be evaluated when incoming webhook events are processed and before a PR workflow is started. If the branch/ref does not match the configured allowlist, the workflow must not execute.

### Backend
- Add new persisted property `branchFilter` to `org.remus.giteabot.admin.Bot`
- Create database migration for the new column/property
- Default value: empty string
- Evaluate `branchFilter` as part of PR-workflow trigger decision logic for webhook-driven events
- Apply the filter in the webhook-to-PR-workflow path, including the logic reached from:
  - `org.remus.giteabot.webhook.UnifiedWebhookController`
  - provider-specific webhook handlers such as `org.remus.giteabot.gitea.GiteaWebhookHandler#handleBotWebhookEvent`

### UI
- Add a text input for `branchFilter` on `bots/[id]/edit`
- Place the field below the `Run workflow when PR head is updated` option
- Add descriptive help text for the field

## Filter behavior
- `branchFilter` is an allowlist for PR workflow execution
- Matching syntax should follow the existing branch/ref whitelist semantics used by GitHub.com or Gitea, using glob-style patterns
- If `branchFilter` is empty or `*`, events for all branches/tags are allowed
- Full ref names may be matched using prefixes such as `refs/heads/` and `refs/tags/`
- If the incoming PR branch/ref does not match the filter:
  - do not start the PR workflow
  - do not post a comment
  - create only a log entry indicating the event was ignored due to branch filter mismatch

## Suggested field description
`Branch (and ref name) allowlist for PR-Workflow execution, specified as glob pattern. If empty or *, events for all branches and tags are reported. Use refs/heads/ or refs/tags/ prefix to match full ref names. See https://pkg.go.dev/github.com/gobwas/glob#Compile documentation for syntax.`

## Acceptance criteria
- [ ] `Bot` has a new persisted `branchFilter` property
- [ ] A migration exists that adds `branchFilter` with default empty value
- [ ] The bot edit page (`bots/[id]/edit`) displays an editable `branchFilter` text input below `Run workflow when PR head is updated`
- [ ] The field includes help text equivalent to the description above
- [ ] When a webhook event would trigger a PR workflow, branch/ref matching is checked before the workflow is started
- [ ] If `branchFilter` is empty, PR workflows behave exactly as before (no filtering)
- [ ] If `branchFilter` is `*`, PR workflows behave exactly as before (no filtering)
- [ ] If the branch/ref matches the filter, the PR workflow starts normally
- [ ] If the branch/ref does not match the filter, the PR workflow does not start
- [ ] On non-match, no PR comment is posted and an ignore message is written to logs

## Test scenarios
1. **Empty filter**
   - Configure `branchFilter` as empty
   - Trigger a PR workflow via webhook
   - Verify workflow executes

2. **Wildcard filter**
   - Configure `branchFilter` as `*`
   - Trigger a PR workflow via webhook
   - Verify workflow executes

3. **Matching branch**
   - Configure `branchFilter` to match the incoming PR branch/ref
   - Trigger a PR workflow via webhook
   - Verify workflow executes

4. **Non-matching branch**
   - Configure `branchFilter` to a pattern that does not match the incoming PR branch/ref
   - Trigger a PR workflow via webhook
   - Verify workflow does not execute
   - Verify no PR comment is posted
   - Verify an ignore log entry is written

5. **Full ref matching**
   - Configure `branchFilter` using a full ref pattern such as `refs/heads/develop`
   - Trigger a matching webhook event
   - Verify workflow executes only for matching full refs

## Notes / implementation constraints
- Preserve current behavior for bots with no configured filter
- The branch filter must affect only the decision to start PR workflows from webhook events
- Logging should make it clear that execution was skipped because the branch/ref did not match the configured allowlist

## Assumptions

- The filter applies to the PR branch/ref value available in the webhook event used to decide whether a PR workflow starts.
- An empty string is the persisted default for existing and new bots.
- The requested glob behavior is intended to use the same or equivalent semantics already familiar from GitHub/Gitea branch filters.
- This issue is limited to PR-workflow triggering and does not change non-PR workflow behavior.

## Open questions

- Should the filter be matched against the PR target/base branch, the source/head branch, or whichever ref the current webhook logic already uses for trigger evaluation? The original issue mentions git-flow and target branch `develop`, but the exact branch source should be confirmed in implementation.
- If multiple patterns are needed in one field, what separator should be supported? The current issue text describes a single text input and a glob pattern, but does not explicitly define multi-pattern input.

---

tmseidel:
# What has been done?

Added a per-bot Branch Filter — an allowlist of branch/ref names that restricts which pull requests can trigger a bot's PR workflows.

- New util/BranchFilter.java: a stateless, dependency-free (pure JDK) helper that evaluates a comma-separated list of gobwas/glob-style patterns (*, ?, [..], {a,b}, **) against a ref. Empty / blank / * allow everything (back-compat). Full ref names (refs/heads/develop, refs/tags/v*) are supported and also match the short branch form.
- Bot.branchFilter column (VARCHAR(1000), NOT NULL DEFAULT '') added via migration V49 (h2 + postgresql).
- Enforcement in BotWebhookService#reviewPullRequest, the single point all provider handlers (Gitea/GitHub/Bitbucket/GitLab) pass through before starting a PR workflow. The filter matches against the PR's target (base) branch, falling back to the head (source) branch when no base ref is present. A non-matching PR is ignored — no workflow runs, no PR comment is posted — and the skip is logged at INFO.
- UI + i18n: new Branch Filter field on the bot edit form, with label + help text in all 8 locales (en, de, es, fr, ja, pt, zh_CN, plus base).
- Docs: trigger-conditions table and a detailed paragraph added to PR_WORKFLOWS.md; a Branch Filter bullet added to USER_GUIDE.md.
- 17 files changed, 602 insertions.

# Why?

In branch-based workflows (e.g. git-flow), a bot should only act on PRs targeting certain branches — e.g. only run reviews on PRs into releases/* and stay silent on PRs into main. Previously there was no way to scope a bot's PR-workflow triggering by branch; a bot either ran on every PR or not at all. The filter gives operators a per-bot allowlist with the historical default (blank/* = all refs) preserved, so existing bots are unaffected. It gates only whether a PR workflow starts — it does not change what the workflow does once running, and it does not affect issue/agent workflows.

# What has been tested?

- New Unit-Tests covering the new feature
- Manual tests

# How AI was used?

An AI coding assistant was used to draft the implementation from the feature spec: the BranchFilter utility (glob→regex conversion and full-ref matching), the reviewPullRequest gate and branch-resolution logic, the test suites, the migration, the form field, the 8 locale strings, and the doc updates. The user defined the requirement, reviewed the generated code and tests, and directed the fixes (e.g. the glob pattern subset and the base-over-head precedence). AI was not used for manual testing.


max-california (approved):
## 🤖 Agentic Code Review

## Summary

The branch gate is integrated at the appropriate shared PR-workflow entry point, preserves the blank-filter behavior, and has good coverage across matching and workflow short-circuiting.

## Findings

### MEDIUM

- **Malformed character classes can crash webhook processing** — `src/main/java/org/remus/giteabot/util/BranchFilter.java:167`

  Character-class contents are copied almost directly into the generated regex. Inputs such as `[]]`, `[a&&]`, or a trailing escaped character can therefore produce an invalid Java regular expression, causing `Pattern.compile()` to throw `PatternSyntaxException`. Because branch filters are administrator-provided and compilation happens before the `try` block in `reviewPullRequest`, the async webhook invocation exits exceptionally rather than recording or clearly logging the bot error.

  Validate patterns when saving the bot and show a form error, or catch invalid-pattern exceptions at the gate and fail closed with an actionable log/error. Adding malformed-pattern tests would prevent regressions.

### LOW

- **Migration comments identify the wrong version** — both `src/main/resources/db/migration/h2/V49__bots_branch_filter.sql:1` and `src/main/resources/db/migration/postgresql/V49__bots_branch_filter.sql:1` say `V47`, although the files are V49. Correcting this will avoid misleading migration history.

---

peledies:
LETS GO! Thanks for adding this so fast. now i can turn off my proxy that did this.

PR #392, opened 2026-09-14
fix(git): fence integration lifecycle across concurrent requests

CaeruleusAqua:
## What has been done?

- Added optimistic version checks to integration edits and SSH confirmation forms, and shared integration-row locking for lifecycle operations and bot assignment.
- Persist deletion_pending before remote cleanup, prevent reactivation/assignment while fenced, and make deletion retryable.
- Carry committed versions through managed-key setup, cleanup, and compensation so stale stages cannot overwrite newer tracking.
- Retain durable cleanup evidence for retry after remote deletion succeeds but local completion fails; safely cancel markers for requests that were never dispatched.
- Add a version-checked deletion-retry form accepting an optional same-owner replacement token without storing it or changing the endpoint/fence.
- Update webhook metadata with targeted queries so detached bot snapshots cannot restore stale integration assignments.
- Add idempotent H2/PostgreSQL V51 with lock_version, deletion_pending, and ssh_cleanup_verified. Existing V48-V50 remain unchanged.
- Document lifecycle/recovery behavior and translate the retry hint into all seven locales.

## Why?

This is the final part of the #368 split, built directly on merged #382. Setup, edits, key cleanup, deletion, and bot assignment must not overwrite each other across concurrent requests or application instances.

Remote mutations deliberately hold the integration row lock through local completion because Gitea does not accept a fencing token. Recovery markers commit beforehand; scans and key generation remain outside that lock. This can block competing admin operations while a remote call is running.

Genuinely ambiguous timed-out registrations with no observed key keep their recovery markers. If they never resolve, verified operator reconciliation is required rather than treating an empty lookup as proof of cancellation.

## What has been tested?

- Focused Maven Java 21 lifecycle suite: 135 tests passed, zero failures/errors/skips.
- Full Maven Java 21 clean verify with temporary /workspace/data: 1,899 tests passed, zero failures/errors/skips.
- After completing locale text: GitIntegrationControllerTest and I18nConfigTest, 33 tests passed.
- node src/test/js/git-integration-form.cjs and git diff --check passed.
- Real Spring/H2 transaction tests cover stale edit/confirmation forms, both bot-assignment/deletion orderings, in-flight registration, pre-POST deletion, stale cleanup/compensation, commit-failure retry, and same-owner/wrong-owner/stale-token deletion recovery.
- Native PostgreSQL 16.15 / Flyway 11.14.1: actual V50-to-V51 upgrades for all eight existing-column combinations, defaults/nullability, preservation of credentials/tracking and populated fence/evidence values, 16 SQL replays, unchanged prior history, and successful validation/no-op reruns.
- Separate JDBC/PostgreSQL checks confirmed stale-version conditional updates and row-lock blocking; these complement, rather than replace, the Spring transaction tests.
- Independent review findings were fixed and re-reviewed with no remaining blockers. No live Gitea or browser end-to-end run.

## How AI was used?

AI reconstructed the final concurrency slice from the archived implementation while preserving all merged review fixes. Separate agents reviewed transaction/recovery behavior and verified the migrations on native PostgreSQL. Findings led to regression tests and fixes for undispatched registration cleanup, durable cleanup evidence, and replacement-token recovery during fenced deletion.


max-california (approved):
## 🤖 Agentic Code Review

## Summary

The lifecycle fencing is thoughtfully implemented and backed by substantial concurrency and recovery coverage. Version checks, pessimistic row locking, durable cleanup evidence, targeted bot updates, and deletion retry behavior are internally consistent. I found no concrete correctness, security, or maintainability issues that should block this change.

## Findings

No findings.

---

max-california (approved):
## 🤖 Agentic Code Review

Summary: this looks like a thoughtful concurrency hardening change with strong test coverage. I did not find a clear correctness or security defect in the lifecycle fencing itself after tracing the controller/service flow, row locking, and retry paths. The main residual risk I see is operational: one repository update path now silently does nothing when the target bot row no longer exists, which can hide data loss / stale-caller problems.

## MEDIUM

### Silent no-op on webhook metadata updates after switching from entity save to bulk update
`src/main/java/org/remus/giteabot/admin/BotRepository.java`  
`src/main/java/org/remus/giteabot/admin/BotService.java`

`incrementWebhookCallCount(...)` and `recordError(...)` were changed from `botRepository.save(bot)` to JPQL `update` statements. That avoids accidentally re-merging a stale `Bot` snapshot, which is good, but the new repository methods return `void`, so the service cannot tell whether the row actually existed.

If a webhook is processed against a stale/deleted bot object, both methods now succeed silently while updating zero rows. Previously, a managed-entity save would at least surface optimistic/deletion issues more naturally. Given these methods are recording audit/operational state, silently dropping the update makes debugging much harder and can mask real races.

Suggested fix:
- change both repository methods to return `int` updated row count
- in `BotService.incrementWebhookCallCount` / `recordError`, verify that exactly one row was updated
- throw/log explicitly when the count is `0`

This keeps the stale-assignment protection while preserving failure visibility.

## LOW

### Consider surfacing optimistic-lock conflicts distinctly from generic “save failed”
`src/main/java/org/remus/giteabot/admin/GitIntegrationController.java`

The new version checks are good, but `OptimisticLockException` is currently handled by the same broad `catch (Exception e)` path as all other failures. Since stale-form submission is now an expected admin race condition rather than an exceptional server fault, it may be worth mapping it to a more specific flash message (“This integration was changed by another request; reload and retry”) instead of the generic save/setup/delete failure text. That would make the new fencing behavior easier for operators to understand.

---

CaeruleusAqua:
Addressed both findings in commit 37c0756 (pushed to the PR branch).

**MEDIUM — silent no-op on webhook metadata updates**
- `BotRepository.incrementWebhookCallCount(...)` and `recordError(...)` now return `int` (updated row count).
- `BotService` verifies that exactly one row was updated and logs an error when the count is zero, so a stale/deleted bot snapshot is no longer dropped silently.

**LOW — optimistic-lock conflicts surfaced distinctly**
- `GitIntegrationController` now catches `OptimisticLockException` / `OptimisticLockingFailureException` separately in save, SSH setup, delete and delete-retry, and shows a dedicated "changed by another request; reload and try again" flash message instead of the generic save/delete failure text.
- New key `flash.staleIntegration` added in all seven locales.

**Tests**
- `BotServiceTest`: added zero-row coverage for both methods (output capture).
- `GitIntegrationControllerTest`: added stale-version coverage for SSH setup and save.
- Ran `BotServiceTest`, `GitIntegrationControllerTest`, `BotWebhookServiceTest`, `UnifiedWebhookControllerTest`, `ManagedSshPersistenceTest`, `GitIntegrationServiceTest`, `GitIntegrationSshMigrationTest`, `IssueWorkflowOrchestratorTest`: all green (226 tests, zero failures/errors/skips).

tmseidel (changes requested):
Since this is the last PR to make the SSH feature working on gitea, I'd like to have this feature tested end-2-end.
There is already a docker-compose that creates a local gitea instance, probably you can reuse it to bootstrap a local environment for a end-to-end testing. Thx 

PS: The code itself LGTM

CaeruleusAqua:
End-to-end result for the managed SSH feature, run live against local Gitea 1.25.5 with the bot on this PR branch (native JDK 21, H2, `APP_ENCRYPTION_KEY` set, auto-login admin).

Environment notes: Gitea ran via a `systemtest/docker-compose-local-gitea.yml` variant on host ports 3001/222 (port 3000 is taken on this machine) with `SSH_DOMAIN=127.0.0.1`, seeded with a `root` admin, an API token, and `root/demo` plus a private `root/ssh-demo` repo. (Had to pull `gitea/gitea:1.25.5` from Docker Hub — `docker.gitea.com` timed out.)

What was exercised, all through the real admin UI and real Gitea API/git:
1. Created Git Integration `local-gitea` (Gitea, `http://localhost:3001`, API token) — saved successfully.
2. SSH preview initially failed ("SSH host-key preview failed") because `getAnySshCloneUrl()` only searches **private** repos and the account had only a public one at that point. After adding a private repo, the preview showed endpoint `127.0.0.1:222`, clone URL `ssh://git@127.0.0.1:222/root/demo.git`, and three host-key fingerprints.
3. Confirmed the fingerprints → "SSH was configured automatically". Gitea now lists the managed key (`AI Git Bot: integration-1-<uuid>`, id 1); the integration row has `transport=SSH`, `ssh_remote_key_id=1`, owner id 1, and the recorded title.
4. Real `git ls-remote ssh://git@127.0.0.1:222/root/ssh-demo.git` using the managed private key (decrypted from the bot DB for the test) and the recorded `known_hosts` returned `HEAD` and `refs/heads/main` — SSH authentication and transport work with the bot-generated key.
5. Switched the integration back to HTTP and saved → the Gitea key was deleted (0 keys remain) and the row shows `transport=HTTP` with all `ssh_remote_key_*` tracking cleared. Managed-key cleanup works end to end.

Two environment-level observations (not blockers, the official Docker image is unaffected):
- The stock Windows OpenSSH 9.5 client cannot complete a host-key scan against the container's OpenSSH 10.0 server (KEX mismatch), so the bot on this machine needed the newer OpenSSH from Git for Windows on `PATH`. On Linux this does not occur.
- Windows resolves `localhost` to `::1` while the Docker port forward is IPv4-only; the Gitea `SSH_DOMAIN` had to be `127.0.0.1` for the scan to reach the server.

One product observation: an account with only public repos gets a generic preview failure, because only private repos are considered for the SSH clone URL. If that restriction is intentional, a more specific hint would help operators.

CaeruleusAqua:
Follow-up with review hardening (`3d18ff8`) and CI end-to-end coverage (`1c2b5c3`).

**Review findings addressed (`3d18ff8`)**
- `known_hosts` pin can no longer be disabled silently: newly submitted values must resolve to exactly one SSH endpoint, otherwise validation rejects them. Previously stored values are untouched.
- Managed-key cleanup deletes by the stable Gitea key ID even when the title was renamed, instead of failing with a title-mismatch error.
- Bot webhook audit updates only touch the in-memory snapshot when the row update persisted (no phantom state), and still log explicitly on zero rows.
- Clearer message when prior-key cleanup fails during setup (the old key may still be live).

**End-to-end in CI (`1c2b5c3`)**
- New `GiteaSshEndToEndTest`: creates a repo on live Gitea, runs integration save → host-key preview → key registration → a real SSH clone through the bot workspace → managed-key cleanup with API verification at each step. Runs only when `GITEA_E2E_URL`/`GITEA_E2E_TOKEN` are set, otherwise skipped.
- New `e2e-gitea-ssh.yml`: boots `systemtest/docker-compose-local-gitea.yml`, seeds an admin token, and runs the test. Triggered for PRs touching the SSH paths.
- The compose is now headless-first-boot ready (SQLite, install lock, no open registration) and pins the SSH host/port so SSH clone URLs are correct.
- Verified locally against Gitea 1.25.5: full flow green in ~12s; without the env vars the test skips cleanly.

tmseidel (approved):
Great work!

PR #395, opened 2026-09-18
fix(gitea): write PR workflow changes to the source repository

myluoluo:
## What has been done?

- Resolve Gitea's authoritative PR head repository and branch before preparing a writable PR workspace. Fail if resolution or source checkout fails, without falling back to the target repository.
- Use that workspace for commit-to-pr in README Sync, i18n coverage, unit-test generation and E2E promotion.
- Reject offer-as-pr for Gitea fork PRs before pushing. Preserve same-repository offers, merge-time promotion and existing checkout behavior for other providers.
- Propagate failed writes, pushes and follow-up PR creation to workflow results, including partial E2E file-write failures.

## Why?

When a fork PR uses a branch name that also exists in the target repository, such as main, the old checkout can clone the target branch successfully. Its origin remains the target repository, so generated documentation or tests can be pushed there instead of to the PR's source branch.

Fetching a PR ref after a failed branch clone does not address this case. Writable workspaces need the source repository coordinates, and failed writes must not be reported as successful workflows.

## What has been tested?

- Rebased onto develop at 7aa8825 and preserved its managed-SSH additions. With Java 21 and Maven 3.9: clean test compilation, all 214 test classes in four batches (1,890 tests, zero failures/errors/skips), and the remaining verify/package phase with tests skipped because they had already run. Each Maven invocation had a 60-second limit.
- The JavaScript Git integration form regression and git diff --check passed.
- Regression tests cover two real local Git repositories with same-named main branches, source-only writes, authoritative-resolution failures, legacy provider fallback, lifecycle routing and failed-write reporting.
- Live Gitea 1.27.3 acceptance used temporary repositories and real PR comments, webhook delivery, model execution and Git pushes. Both a fork PR and a same-repository PR received a README-only bot commit on the source branch. PR head SHAs matched those commits, and the target main SHA stayed unchanged. Temporary repositories and test credentials were removed.
- The live test exercised README Sync commit-to-pr on the release-based deployment. Other workflows and failure cases were covered locally; the develop-based contribution is not claimed to have been redeployed for live testing.

## How AI was used?

AI assisted with diagnosis, implementation, regression tests, upstream compatibility checks and live acceptance. Separate AI reviewers assessed the plan and implementation; a review finding about partial E2E file writes was reproduced with failing tests, fixed and re-reviewed. The contributor directed the scope and reviewed the proposed approach.


max-california (approved):
## 🤖 Agentic Code Review

## Summary

The authoritative Gitea head resolution and fail-closed writable workspace address the wrong-repository push risk. The affected workflows now consistently surface checkout, write, push, and follow-up PR failures, and the regression coverage is strong.

No blocking, medium, or low-severity findings.

---

max-california (approved):
## 🤖 Agentic Code Review

Summary: the core fix looks sound. The new authoritative PR-head resolution closes the “same branch name in target repo” Gitea bug, and I like that write/push/follow-up-PR failures now propagate to workflow status instead of being silently treated as success. I only found one maintainability/correctness gap around duplicated head-resolution error handling.

## MEDIUM

### `isAuthoritativePullRequestFromFork` can now throw uncaught provider errors before workflow error reporting
`src/main/java/org/remus/giteabot/agent/validation/WorkspaceService.java:220`

`prepareWritablePullRequestWorkspace(...)` defensively wraps `getPullRequestHead(...)` and converts resolution failures into a `WorkspaceResult.failure(...)`, but `isAuthoritativePullRequestFromFork(...)` calls the same provider method directly and dereferences the result without any validation/catch.

That means the newly added early fork checks in:
- `src/main/java/org/remus/giteabot/prworkflow/readmesync/ReadmeSyncService.java:115`
- `src/main/java/org/remus/giteabot/prworkflow/i18n/I18nCoverageService.java:107`
- `src/main/java/org/remus/giteabot/prworkflow/e2e/promotion/SuitePromotionService.java:109`

will fail via an exception path if Gitea returns malformed PR data or the API call itself fails. In `ReadmeSyncService`/`I18nCoverageService` that becomes a generic “unexpected error”; in `SuitePromotionService` it relies on the outer broad catch. Functionally it won’t write to the wrong repo, but it’s inconsistent with the fail-closed/report-cleanly behavior you implemented for writable workspace preparation.

Suggested fix: either make `isAuthoritativePullRequestFromFork(...)` reuse the same guarded resolution logic/validation as `prepareWritablePullRequestWorkspace(...)`, or add a shared helper that returns a validated `PullRequestHead` / failure result so both call sites handle provider failures consistently and produce the intended workflow-facing message.

---

tmseidel (approved):
@myluoluo Haha, very good catch, I actually didn't test this branch e2e cause I don't need it. Thanks for fixing that. Code LGTM - Thx for your contribution

myluoluo:
> @myluoluo Haha, very good catch, I actually didn't test this branch e2e cause I don't need it. Thanks for fixing that. Code LGTM - Thx for your contribution

Thanks for reviewing! I ran into this in my Gitea setup, so I’m glad the fix helps. :)

PR #396, opened 2026-09-19
fix(agentic-review): reject incomplete review outcomes

### LINKED ISSUE #393: Agentic reviews sometimes finish without a usable review
Hi everyone,

I have noticed a problem with agentic reviews in my setup. Sometimes I do not get a usable review at all. The model returns an empty answer, or the bot posts an intermediate message like “Let me check the remaining files...” instead of an actual review. The run can still be marked as successful.

I looked into why this happens. With reasoning models, the completion budget can be used up without producing any visible answer. The review loop can then treat the response as finished and fall back to text from an earlier step. I also reproduced the empty-output case in a few small tests with reasoning models through OpenRouter.

I think we should improve how the review is brought to a proper conclusion, rather than just increase the token or round limits. My idea would be to:

- Separate exploring the code from writing the final review, and reserve enough budget for the final step.
- Keep a small structured record of findings, so results are not lost when the conversation context is shortened.
- Only publish a complete, validated review. If that is not possible, mark the run as failed instead of publishing an empty answer or an earlier intermediate message.
- Make the effective model limits and review budgets clear and configurable in the existing UI.
- Handle reasoning, cancellation and retries consistently, without adding special cases for individual models or accidentally publishing duplicate reviews.

OpenRouter’s model metadata could also help with configuring the limits, but I would keep that separate from the core reliability fix.

This would touch several parts of the review workflow, so I would like to discuss the direction before opening a large PR. I would start with a small fix to prevent incomplete reviews from being published, then split the remaining work into smaller PRs.

What do you think about this approach? Are there any existing plans or preferences that I should take into account?

CaeruleusAqua:
## What has been done?

- Require a completed, non-empty native text turn before accepting a review. Incomplete turns cannot execute tools.
- Remove the fallback that used earlier assistant text when a response was empty or a round/context budget ran out.
- Check the loop outcome before publishing reviews or clarification answers.
- Distinguish posted results, an empty diff, and failures so failed runs are not reported as skipped/successful.
- Add regression coverage through the real review strategy, tool router and agent loop.

## Why?

This is the first small reliability fix discussed in #393. A reasoning model can use up its completion budget without returning visible text. Previously, the bot could publish an earlier "Let me check..." message as a successful review.

The regression test reproduces that two-turn sequence without a live model. This PR addresses completion and publication guards. Native/legacy dispatch, finalization, structured findings and configurable budgets are separate follow-ups. It intentionally does not close the overall issue.

## What has been tested?

- Red/green reproduction of the stale-text publication, round exhaustion, legacy context exhaustion and failed-versus-skipped cases.
- Empty/completed/truncated turns, invalid turns containing tools, failed clarifications, legitimate no-diff results and normal final reviews.
- `mvn -B clean test`: **1,891 tests passed**, no failures, errors or skipped tests, in Linux using `maven:3.9-eclipse-temurin-21`.
- The full run used a fresh temporary `/workspace/data` directory, a test-only encryption key and temporary prompts. This isolates tests that do not declare the test profile from an existing development database. No database or test configuration changes are included.

## How AI was used?

AI assisted with tracing the failure, writing the fix and regression tests, running the test suite, and reviewing the diff.


max-california (approved):
## 🤖 Agentic Code Review

Summary: This looks like a solid reliability fix. The new guards line up with the bug described in the PR: incomplete/native-truncated turns and exhausted budgets no longer get published as if they were successful reviews, and the workflow now distinguishes true no-op (`NO_DIFF`) from actual failures. The added regression coverage is well targeted.

No correctness or security issues stood out in the changed paths.

### LOW
- No findings.

---

tmseidel (approved):
LGTM

PR #397, opened 2026-09-19
fix(agent-loop): dispatch legacy responses to the text strategy

### LINKED ISSUE #393: Agentic reviews sometimes finish without a usable review

CaeruleusAqua:
## What has been done?

- Dispatch strategy responses according to the resolved tool mode: native responses use the `ChatTurn` overload, legacy responses use the `String` overload.
- Strengthen the existing transport tests so they also check which strategy handler receives the response.
- Add a regression test with the real review strategy: a legacy file request is fulfilled and its content reaches the next model turn before a final review is returned.

## Why?

Related to #393 and complementary to #396. The loop already selected the legacy transport correctly, but then wrapped its response in `ChatTurn` and invoked the native handler. For strategies overriding both methods, this bypassed the legacy parser. An agentic review could therefore stop on its first context request.

This PR is independently based on `develop`. It contains only the dispatch fix and its tests; completion guards and the larger review improvements remain separate.

## What has been tested?

- Red/green tests for a legacy-only client, a native-capable client without tool descriptors, and a two-turn legacy review.
- Existing native tool-result forwarding and agent-loop tests.
- `mvn -B clean test`: **1,876 tests passed**, no failures, errors or skipped tests, in Linux using `maven:3.9-eclipse-temurin-21`, with fresh temporary application data and prompts.

## How AI was used?

AI assisted with tracing the dispatch path, writing the regression tests and fix, running the tests, and reviewing the diff.


max-california (approved):
## 🤖 Agentic Code Review

Summary: the fix looks correct and well-targeted. The main behavioral bug described in the PR is addressed in `AgentLoop`, and the added tests cover both the transport selection and the legacy review regression in a meaningful way. I did not find any blocking correctness or security issues in the changed code.

## Findings

No findings.

---

tmseidel (approved):
LGTM

PR #398, opened 2026-09-19
fix(ai): preserve native completion failures

CaeruleusAqua:
## What has been done?

- Preserve token-limit and unknown stop reasons in the native OpenAI-compatible, Anthropic, Google and Ollama clients, even when the response contains parsable tool calls.
- Return an incomplete native turn for an empty HTTP/JSON response instead of inventing a successful turn with an error message as its text.
- Keep the reported stop reason and token counters when a limit response has no visible content.
- Require Ollama's terminal `done` marker before treating a native streamed turn as complete.
- Add shared HTTP contract tests for all four clients and a truncated-stream regression test for Ollama.

## Why?

Follow-up to #393 and #396. The review completion guard needs the real stop reason, but the provider adapters could overwrite `MAX_TOKENS` or `OTHER` with `TOOL_USE` whenever a tool call was parsable. An empty response could also become a synthetic `END_TURN`.

This small step corrects native response metadata. It leaves the legacy text API and broader usage/budget work for separate follow-ups. Valid completed tool turns retain their existing behavior.

The branch was managed with git-spice above #396, then synchronized/restacked onto `develop` after #396 and #397 were merged.

## What has been tested?

- Red/green regression tests first reproduced the stop-reason overwrite in all four providers, then the empty-response and missing-terminal-marker cases.
- 28 shared native HTTP contract cases plus an additional Ollama streaming regression.
- `mvn -B clean test` after restacking: **1,921 tests passed**, no failures, errors or skipped tests. Linux, Java 21, Maven 3.9, fresh temporary application data/prompts.
- No live model/API inference was needed for these tests.

## How AI was used?

AI assisted with investigating the response handling, writing the fix and regression tests, running the tests, and reviewing the diff.


max-california (approved):
## 🤖 Agentic Code Review

Summary: the change looks correct and well-covered. The provider adapters now preserve non-success stop reasons instead of rewriting them to `TOOL_USE`, and the new contract tests exercise the previously broken cases across all four native clients. I didn’t find any blocking correctness or security issues in the diff.

## Findings

### LOW
- **Test coverage gap: Ollama truncated native streams don’t assert tool-call retention explicitly**
  
  The new `nativeStreamWithoutDoneChunkIsNotACompletedToolTurn` test correctly verifies that a stream without a terminal `done:true` chunk is no longer treated as a completed tool turn. However, it only asserts `stopReason == OTHER` and the partial text. Since `interpret()` still parses tool calls from the last non-final chunk, it would be useful to also assert `turn.toolCalls()` contains the streamed call. That would better lock in the intended behavior described in the PR: preserve native metadata/tool payloads while avoiding a false “complete” turn.

---

tmseidel (approved):
Looks good

PR #399, opened 2026-09-20
fix(ai): retain metadata for text-only turns

CaeruleusAqua:
## What has been done?

- Keep the provider's stop reason and token counts in `chatWithTools` when the tool list is empty/null or native tools are disabled.
- Reuse the response interpreters in the OpenAI-compatible, Anthropic, Google and Ollama clients. Remove the obsolete String-to-`ChatTurn.text` delegate.
- Preserve the existing text-request history conversion, default prompt, model/token overrides, Anthropic thinking/caching settings and Ollama JSON mode.
- Cover the response metadata and HTTP request contracts with 60 parameterized regression cases.

## Why?

Follow-up to #393 and #398. The text fallback was discarding completion metadata and inventing `END_TURN` with zero usage, including when the model hit its token limit. An empty response could also turn into a successful turn containing an error message.

This provides reliable typed text responses for the subsequent legacy-loop migration, using the existing API. The branch was created with git-spice above #398 and retargeted to `develop` after #398 was merged.

## What has been tested?

- Red/green: the 12 token-limit cases failed on the old code with `END_TURN` instead of `MAX_TOKENS`.
- 154 targeted provider, request-shape, auditing and factory tests passed.
- Full isolated `mvn -B clean test`: **1,981 tests passed**, no failures, errors or skipped tests; Java 21 / Maven 3.9 on Linux.
- Independent standards and spec reviews found no issues.

## How AI was used?

AI assisted with tracing the fallback paths, implementing the fix, writing and running the regression tests, and reviewing the diff.


tmseidel (approved):
LGTM

